PR #2, opened 2020-04-02
ptn0 folder structure and basic startup script

gufmar:
this PR proposes to create a ptn0 folder in /files with subfolders as we plan to use it on the nodes. Therein we can place scripts, templates and config files in an intuitive way.

it moves the existing genesis/topology files and adds a node startup script


gufmar:
Note: if this gets approved the folder initialisation script at https://github.com/cardano-community/guild-operators/wiki/Getting-started-with-Haskell-Node#folder-structure needs to be updated

PR #10, opened 2020-04-20
enhance topology json with operator and node names

gufmar:
tested: adding new fields seems not creating startup/runtime issues


mmahut:
Looks good.

PR #34, opened 2020-04-28
Change Proposal 2

redoracle:
Few changes, some cosmetic.


redoracle:
I guess sh linter is not interpretating the sudo option.. But the script works.. Do you know how to fix it? Maybe exporting SUDO variable?

PR #40, opened 2020-04-29
Update cabal-build-all.sh

hex-pool:
Added full path to cabal;
Alternatevely, ~/.ghcup/bin can be added to the PATH variable, however this could interfere with other cabal installations


rdlrt:
If you've followed the instructions (essentially [prereqs.sh#L50](https://github.com/cardano-community/guild-operators/blob/master/files/ptn0/scripts/prereqs.sh#L50)), it would already set up ~/.ghcup/bin in your $PATH environment variable. This means you shouldnt have to execute this line or change the reference.

Let me know if you that is not the case for you.

PR #49, opened 2020-05-04
Updated prereqs.sh

redoracle:
- removed unused directory
- added ~/.ghcup/bin to PATH env.

### DIFF
--- a/files/ptn0/scripts/prereqs.sh
+++ b/files/ptn0/scripts/prereqs.sh
@@ -58,9 +58,9 @@ echo "Installing Cabal 3.0.0 .."
 wget https://downloads.haskell.org/cabal/cabal-install-3.0.0.0/cabal-install-3.0.0.0-x86_64-unknown-linux.tar.xz
 tar xf cabal-install-3.0.0.0-x86_64-unknown-linux.tar.xz
 chmod 755 cabal
-mkdir -p ~/.cabal/bin
 mv cabal ~/.ghcup/bin
 rm -f cabal-install-3.0.0.0-x86_64-unknown-linux.tar.xz cabal.sig
+export PATH=$PATH:~/.ghcup/bin
 
 echo "Creating Folder Structure .."
 


files/ptn0/scripts/prereqs.sh
```diff
@@ -58,9 +58,9 @@ echo "Installing Cabal 3.0.0 .."
 wget https://downloads.haskell.org/cabal/cabal-install-3.0.0.0/cabal-install-3.0.0.0-x86_64-unknown-linux.tar.xz
 tar xf cabal-install-3.0.0.0-x86_64-unknown-linux.tar.xz
 chmod 755 cabal
-mkdir -p ~/.cabal/bin
```
rdlrt, line 61: ~/.cabal/bin is used as destination for output of cabal build outputs (part of cabal-built-all.sh), which resembles standard behaviour for cabal install mode

files/ptn0/scripts/prereqs.sh
```diff
@@ -58,9 +58,9 @@ echo "Installing Cabal 3.0.0 .."
...
 mv cabal ~/.ghcup/bin
 rm -f cabal-install-3.0.0.0-x86_64-unknown-linux.tar.xz cabal.sig
+export PATH=$PATH:~/.ghcup/bin
```
rdlrt, line 63: ~/.ghcup/bin should be added to $PATH via adding of ~/.ghcup/env in source (part of ghcup install), tho I havnt tested it since adding the silent mode installation (the interactive mode does prompt for confirmation to add ghcup/env to profile).

However, I agree we might want to reference the env file within script (to ensure the existing session retains the PATH reference too)

files/ptn0/scripts/prereqs.sh
redoracle: Than we also need it to be included in the PATH as well:
Warning: Installation path /root/.cabal/bin not found on the PATH environment variable.

files/ptn0/scripts/prereqs.sh
redoracle: I did test it from a debian-minimal image and PATH as is (not interactively) was not updating it. we should add ~/.cabal/bin as well as stated earlier.

files/ptn0/scripts/prereqs.sh
rdlrt: Yep which is done by ghcup install too (atleast in interactive mode)

files/ptn0/scripts/prereqs.sh
rdlrt: Try installing ghcup without the environment variable added recently.

redoracle: I think that:
```
. ~/.ghcup/env L50
```
has to replace the export after the cabal install and that will do the trick, since both cabal and ghup path are included there.. we may want also to do something like:
```
echo "source ~/.ghcup/env" >> ~/.bashrc
```
??
-- Note: I prefer "source" vs "." because I like it more :)

rdlrt:
Resolved via #51

PR #65, opened 2020-05-10
Revert "Update pHTN to Shelley"

psychomb:
Reverts cardano-community/guild-operators#64


rdlrt:
Will work with master

PR #92, opened 2020-05-15
Updated prereg.sh

ilap:
Added some automation to create different environment for different cardano-node instances.


rdlrt:
@ilap - One of the reasons we had to swap `read` values with hard coded ones was use of docker images, wonder if its better option to detect and update or to give response files for input for the choices instead?

PR #100, opened 2020-05-17
Updated scripts

Scitz0:
While getting familiar with the F&F exercises I created some scripts to handle transactions. 

I based my script on the existing sendADA.sh and then rewrote most of it. 
It now also handles multiple UTx0 inputs, ie doesn't matter how fragmented the address is. 
Defrag of address is now also a feature due to the above change if you set both destination and source to the same address and specify 'all' as the amount. 
There are other tweaks here and there as well.

env file updated with new protocol parameters and waitNewBlockCreated function, largely inspired by jormu-helper-scripts

balance.sh script added to get the balance of address, sourced by sendADA.sh.


Scitz0:
An explanation for requiring both source address and source sign key as arguments.

The reason is that for a staking address you have multiple key pairs, payment, staking etc and therefor multiple addresses. When sending ADA from a staking address it should be signed with the payment skey, not the staking skey. Its easier to let the executor decide what address and what sign key to use.

Scitz0:
I just saw that there is a lot of cli changes in the pipeline that will break the script.
Script tested and verified against pioneer-2 tag.

bergr01:
If I'm not mistaken, the new -n $output checks that are added to sendADA.sh check for whether or not command execution returned non-empty text string while executing, which is not what I was trying to check with my original $? exit status checks. Also since you're already invoking balance-related scripts, is it worth throwing an error or waiting-for-another-block if there are signs the tx either failed or is still waiting to be included in a block?

Scitz0:
Thanks for the response. Here are my comments for your observations. 

1) Failed check on build, sign and submit.
My original plan was to check the exit code as you suggested but after testing it seems that the cardano-cli command exit with code 0 even on failure. See attached screenshot where I intentionally put in wrong fee making submit command to fail. This is why I used the not so pretty solution to check if the response from the command contains an error message. On successful execution command returns nothing.  
image

2) TIMEOUT_NO_OF_BLOCKS
Good observation, the entire function taken from previous jorm scripts env file with some minor tweaks. Should say TIMEOUT_NO_OF_SLOTS as this is what it's based on. On testnet slot duration is one second but this might change.

PR #101, opened 2020-05-18
Update shlint.yml

rdlrt:
Scan all .sh files


rdlrt:
To be merged after #100

PR #117, opened 2020-05-22
Key Vault & Create Pool script

Scitz0:
Key vault api/script to be used to store pool/wallet keys encrypted by GPG

createPool.sh added that creates keys needed to run a core node
key.skey, vrf.skey and opcert stored in unencrypted pool folder
coldcounter, cold.skey, cold.vkey, kes.vkey & vrf.vkey stored in the encrypted vault.


Scitz0:
I didn't update docs as I wanted some feedback on implementation first.
Also createPool.sh is not complete, the register on chain part is not included yet.
Only operational node certificate part.

oldcryptogeek:
Was in the process of testing.  Why was this closed?

rdlrt:
@oldcryptogeek a newer PR which encompasses these changes will be added :)
But you can use these scripts if you'd like

PR #121, opened 2020-05-26
create pool, create delegation address

papacarp:
you can now update an address to be used for staking, and you can create and register pools.


papacarp:
To create a staking address:
1.  Create a new payment wallet ( wallet -> new -> payment)
2.  Upgrade the payment wallet to a staking wallet (wallet -> new -> staking) - seems to be a bug in the validation of the transaction balance amounts.  I can try to track it down tomorrow.

To create a pool and register it
1.  Make sure you have a staking wallet setup with enough funds to cover your intended pledge
2.  Create the pool (pool -> new)
3.  Restart your cardano-node to use the pool config files (as given to you in the output of the script)
4.  Register your pool (pool -> register)
- This will walk you through the following steps:
  - Choose which pool you will be registering
  - Set your pool parameters for pledge, cost, margin
  - Choose which wallet you will pay your fees with
  - Choose which wallet you will pledge with.  By default rewards address is specified to be same as pledge address.  We could make this optional later

PR #133, opened 2020-05-27
delegation and minor cleanup of register pools

papacarp:
Added delegation and cleaned up the pool register section to only require one wallet.  I think the pay and stake wallet need to be the same.


papacarp:
whoops.  looks like I need to go remember how to rebase

PR #136, opened 2020-05-28
Fix delegation - WIP

papacarp:


scripts/cnode-helper-scripts/cntools.sh
```diff
@@ -1,14 +1,14 @@
...
-# Creators: gufmar, Scitz0
+# Creators: gufmar, Scitz0, Papacarp
 # 2020-05-19 cntools initial release (concept)
 # 2020-05-24 helper functions moved cnlibrary & configuration to env file
 
 ########## Global tasks ###########################################
 
 # get common env variables
-. "$(dirname $0)"/env
+. "$(dirname $0)"/env_guild
```
rdlrt, line 11: Wouldnt this cause confusion between users? Not everyone using these script uses guild repo, and we'd hopefully rely on CNODE_HOME to distinguish between different networks

scripts/cnode-helper-scripts/cntools.sh
```diff
@@ -281,8 +284,28 @@ case $OPERATION in
...
+          unset password
+          echo "" && say "${RED}ERROR${NC}: failure during payment verification key decryption" "log"
+          echo "" && read -r -n 1 -s -p "press any key to return to home menu" && continue
+        fi
+      fi
+
       ${CCLI} shelley stake-address key-gen --verification-key-file "${staking_vk_file}" --signing-key-file "${staking_sk_file}"
       ${CCLI} shelley stake-address build --staking-verification-key-file "${staking_vk_file}" > "${staking_addr_file}"
+      # upgrade the payment address to an address that delegates to the new stake address
+      ${CCLI} shelley address build --payment-verification-key-file "${payment_vk_file}" --staking-verification-key-file "${staking_vk_file}" > "${stakepayment_addr_file}"
```
rdlrt, line 307: This might get a conflict as per #137

scripts/cnode-helper-scripts/cntools.sh
Scitz0: I also think we should keep existing env name, can be used with other networks as well, not exclusive for guild network in my book.

papacarp: yep, that is an artifact of my testing.

gufmar: would it make sense to have one env file with [common] and [testnet: pHTN ... ] sections?

There could be one initial variable ACTIVE_TESTNET="pHTN" and then each testnet in a case section.
Then the prereqs curl command is always the same, and switching is quick and easy

PR #137, opened 2020-05-28
Update as per breaking CLI changes to master

rdlrt:


Scitz0:
I think all changes are valid but I haven't compiled and tested against latest master.

I think we can merge this one.

ilap:
All goods, thx.

PR #141, opened 2020-05-28
Misc minor updates

Scitz0:
* Added prerequisites and code change to exit on error

* Clear trap on loop

* Remove all unwanted characters from name when creating wallet and pool


Scitz0:
Not sure how to best remedy SC2001

gufmar:
What about replacing the RegEx with a simpler substitution 
https://stackoverflow.com/questions/42740325/bash-variable-search-and-replace-instead-of-sed (also see 2nd/3th comment)

PR #144, opened 2020-05-28
fix delegation rebased, squashed, improved

papacarp:
ready to merge
1.  fixed a few more cli update changes
2.  incorporated all other changes from the last PR, and will close that down now.
3.  create staking address works, as does create pool, and register pool.
4.  waiting for epoch transition to confirm stake got assigned to pool properly - WORKS!


papacarp:
dropping these comments in here for posterity

Wouldnt this cause confusion between users? Not everyone using these script uses guild repo, and we'd hopefully rely on CNODE_HOME to distinguish between different networks

 @Scitz0
Scitz0 13 hours ago Collaborator
I also think we should keep existing env name, can be used with other networks as well, not exclusive for guild network in my book.

 @papacarp
papacarp 7 hours ago Author Collaborator
yep, that is an artifact of my testing.

 @gufmar
gufmar 4 hours ago Collaborator
would it make sense to have one env file with [common] and [testnet: pHTN ... ] sections?

There could be one initial variable ACTIVE_TESTNET="pHTN" and then each testnet in a case section.
Then the prereqs curl command is always the same, and switching is quick and easy

papacarp:
Ok, verified that it creates pools and pledges to them at the same time.  I think this is good to merge

PR #152, opened 2020-05-29
rotate kes

papacarp:
I had to rotate my keys on FF.  so added it to cntools.


papacarp:
Ok, I see we failed again.  any jq experts want to weigh in on the best way to do this:

`slotsPerKESPeriod=$(cat $GENESIS_JSON | jq -r '.slotsPerKESPeriod')`

without falling into the [UUOC](https://en.wikipedia.org/wiki/Cat_%28Unix%29#Useless_use_of_cat) linting issues?  Or should we just disable this lint item?

papacarp:
So my only question remaining at this point is do we regenerate the hot keys every time as well, or not.  sounds like we don't need to regenerate them.

rdlrt:
@papacarp Not everytime, but only when keys are about to expire.
Duncann commented on one of the PRs that we should be able to grab kes expiry status from metrics, havnt checked it yet.
But yeah - for now a manual option to renew is good to have

PR #165, opened 2020-05-31
Update prereqs.sh

matthijs-aeon:
Removed 'p7zip' prerequisite for CentOS.


Scitz0:
7zip should be removed from Debian apt-get command as well.

rdlrt:
The PR is stale now, and changes have already been incorporated, closing

PR #167, opened 2020-05-31
Update prereqs.sh

redoracle:
- added gnupg


redoracle:
tested on both Centos and Debian

PR #171, opened 2020-06-01
File protection, empty name check and misc display updates

### LINKED ISSUE #168: [CNTOOLS] all pools get deleted if you enter a blank pool name
I  just had all my pool folders get deleted when I went to create a new pool and then abandoned.  I was unable to exit the command.  I think we should detect a blank pool name and just exit rather than continuing to create the pool and asking for a password.

I suspect this area is deleting all pool entries:

https://github.com/cardano-community/guild-operators/blob/5ed849e2e8fb703ae430407fd9518e719bbe5b67/scripts/cnode-helper-scripts/cntools.sh#L1148

https://github.com/cardano-community/guild-operators/blob/5ed849e2e8fb703ae430407fd9518e719bbe5b67/scripts/cnode-helper-scripts/cntools.sh#L1153

Wallets might have a similar issue

Scitz0:
* Write protection of pool and wallet files
* Current config option 'PROTECT_KEYS' now refer to fs write protection
* A new config option added called 'ENCRYPT_KEYS' with same function as previously, ie key encryption
* Sanity check added for empty wallet or pool name, fixes #168
* Sanity check added for missing or empty wallet or pool folder
* Internal variable and config renaming to harmonize naming convention 'staking' -> 'stake'
* Display / menu updates to make end user experience better


papacarp:
Looks good.   The new locking gracefully skips that step if it doesn't have root access.  great work.

I'm going to open a separate bug for the UTXO funds issue for creating a pool.  We will either need to address it or specify that the largest UTXO must have enough funds to entirely handle the transaction or it will fail.  I'd vote for adding complexity to create the transaction with multiple (or maybe All?) utxo's eventually.

but this PR should be merged.  its good.

Scitz0:
Found a bug in the code, hold merge for now

Scitz0:
Should be ok to merge now I think

Scitz0:
Still some issues to sort out

PR #176, opened 2020-06-04
Fix dependencies with node-gyp on Ubuntu

rdlrt:
libssl version needs to be specific as it causes issues for nodejs-dev:

`nodejs-dev : Depends: libssl1.0-dev (>= 1.0.2) but it is not going to be installed`


rdlrt:
Going ahead with merge as it would cause an error for people using master

PR #177, opened 2020-06-04
Updated genesis

rdlrt:
Update Genesis to match IOHK QA for some bits:
- Reduce max Block Size (since most blocks go empty, makes little sense to have huge blocks)
-  Increase Transaction Size (allows more input/outputs, and potentially more certs)
- Reduce nOpt (to see the effects of going beyond more pools than optimal desired)
- Add rho and tau constants to match QA (it wasnt set before)
- Match protocolMagicId and networkMagic (not clear why they're still present), update them so that ones on older genesis do not connect to new one

### DIFF
--- a/files/ptn0/files/genesis.json
+++ b/files/ptn0/files/genesis.json
@@ -9,26 +9,26 @@
     },
     "minUTxOValue": 0,
     "decentralisationParam": 0.5,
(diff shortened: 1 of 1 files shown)


psychomb:
Fine with me. (not reconnected to pHTN yet :P )

PR #180, opened 2020-06-04
Update Genesis: reduce securityParam, troubleshooting rewards

rdlrt:
Logic being ( epochLength * activeSlotCoEff * (potential lower limit of d) ).
Thus, 900 * 0.4 *0.1 = 36

This is part of troubleshooting to see if that's what prevents rewards from being distributed


gufmar:
looks good

PR #183, opened 2020-06-04
on-the-fly protection removed

Scitz0:
Major cleanup of encryption/decryption and fs write protection. Handling it on-the-fly added too much complexity with little added value.

Manual decryption/encryption inkl write protection is now the method moving forward. Config parameters removed as they doesn't provide any value now.

Added extra sanity checks here and there.

It's now possible to run cntools without a running node to connect to. This adds the ability to access offline functions as decryption/encryption as well as payment wallet creation and pool creation.


gufmar:
LGTM

Issue #172, opened 2020-06-01
[CNTOOLS] Pool registration fails if largest UTXO does not have enough funds


papacarp:
The transaction implementation currently only uses the largest UTXO to fund a pool registration transaction.  This was done out of simplicity, but eventually we should probably clean this up.  Example failure:

```
-- Sending transaction to chain --
 -- Register Pool and Pledge on Chain --

--- Balance Check Payment Address -------------------------------------------------------
                           TxHash                                 TxIx        Lovelace
----------------------------------------------------------------------------------------
494bc78ad74afbe926344a17ce22147d923bd58b918b3d9b54b840920953d811     0         100000000
0eac44153aaee2859cc46c2c62aaa7343755ca58f1a5ca288a7f06c79ced1991     0          99980900

Total balance in 2 UTxO is 199980900 Lovelaces or 199.9809 ADA

Using UTxO with highest balance:
494bc78ad74afbe926344a17ce22147d923bd58b918b3d9b54b840920953d811     0         100000000

 -- Protocol Parameters --
TN Magic is 4001
Pool Deposit is 100000000
Current slot is 146028, setting ttl to 147028

 -- Calculate fee, new amount and remaining balance --
fee is 32700
ERROR: Not enough Lovelace in payment address (100000000 < 32700 + 100000000)
ERROR: failure during pool registration, removing newly created pledge and registration files

```

Scitz0:
Yes, we should handle it like send ADA function.
It shouldn't be that hard to replicate for the other functions that only use utxo with the highest balance. 
But there is also work being done to mitigate this issue by IOHK if I remember correctly so that all funds in an address can be treated as one. I don't have a reference to provide for this though to track progress.

rdlrt:
@Scitz0 that would be at cardano-wallet layer I assume?

Scitz0:
> @Scitz0 that would be at cardano-wallet layer I assume?

I think its this one.
https://github.com/input-output-hk/cardano-node/issues/1039
Status in progress

Scitz0:
Fixed in 6d27722

PR #194, opened 2020-06-05
Pool Modify

Scitz0:
Added a new pool modification option to change pool parameters.

Pool registration and wallet delegation now use all available utxo in base address


Scitz0:
In addition to modify function PR also fix #172

PR #198, opened 2020-06-07
Pool registration/modification input validation/change + misc

Scitz0:
- Library -
* Three new helper functions added:
  ADAtoLovelace, lovelacetoADA, pctToFraction
* getBalance & sendADA modified to use these new functions
* sendADA: lovelace conversion moved to cntools.sh and using said new functions
* sendADA: removed unnecessary balance check at start
* sendADA: all conversion moved to cntools.sh

 - cntools -
* Wallet >> Remove - now check both payment and base address for funds
* Funds >> Send - removed lovelaces in summary, ADA enough to show
* Funds >> Delegate - summary updated to include more relevant information
* Pool >> Register & Modify:
    Margin now entered in % instead of fraction number
    Validation added to all inputs that numbers are sound
    Added pool parameters to summary
    A pledge vs balance check added to summary, if balance is below set pledge a warning is given
* Pool >> Show - saved pool parameters added to output(not queried from chain)

### DIFF
--- a/scripts/cnode-helper-scripts/cntools.library
+++ b/scripts/cnode-helper-scripts/cntools.library
@@ -218,7 +218,7 @@ waitNewBlockCreated() {
 #   >> validateDecimalNbr 100.5
 validateDecimalNbr() {
   re_decimal_nbr='^[0-9]+([.][0-9]+)?$'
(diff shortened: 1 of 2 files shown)


Scitz0:
Noticed some issues, do not merge until I have some time to look it over

PR #203, opened 2020-06-08
topology updater v1

gufmar:
the first version, to propose the (centralised) concept and start collecting peer registrations, before a larger number of new pool operators join the topologies


oldcryptogeek:
Great Work @gufmar :tada:

PR #206, opened 2020-06-08
CNTOOLS - Total redesign of all menus

Scitz0:
Main navigation menu including all wallet and pool selection menus updated to use a new keyboard navigation menu instead of typing a number or letter to access the function you want. This should reduce the number of wrong selections.
Home/Cancel option added to all select menus.

### DIFF
--- a/scripts/cnode-helper-scripts/cntools.library
+++ b/scripts/cnode-helper-scripts/cntools.library
@@ -90,6 +90,133 @@ protectionPreRequisites() {
 }
 
 
+# Command    : selectOption [opt1] [opt2] ...
+#
+# Description: Create a bash menu to select one of the provided options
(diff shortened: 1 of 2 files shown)


Scitz0:
Docs need an update after this change if merged

papacarp:
The new navigation is great, but I'm one of those guys that liked the keyboard navigation thing without pressing up/down so much.  I didn't realize there were issues with people pressing the wrong keys.  

Would it be difficult to also select your option by letter?  for example the main menu U,W,F,P,Q.  Wallet menu is N,L,S,R,D,E,H.   Funds menu is S,D,H.  Pool menu is N,R,M,L,S,R,D,E,H.  

If there is a reason you did not do this then let me know.  Otherwise, I may take a stab at it when I get some time.

PR #218, opened 2020-06-12
Added FUNDS -> WITHDRAW REWARDS functionalty

hodlonaut:
Also corrected config file name in env

### DIFF
--- a/scripts/cnode-helper-scripts/cntools.library
+++ b/scripts/cnode-helper-scripts/cntools.library
@@ -485,13 +485,14 @@ getBalanceAllAddr() {
   fi
   if [ -f "${base_addr_file}" ]; then
     base_addr=$(cat "${base_addr_file}")
+    stake_addr=$(cat "${stake_addr_file}")
     getBalance ${base_addr} >/dev/null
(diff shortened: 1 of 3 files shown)


scripts/cnode-helper-scripts/cntools.library
```diff
@@ -485,13 +485,14 @@ getBalanceAllAddr() {
   fi
   if [ -f "${base_addr_file}" ]; then
     base_addr=$(cat "${base_addr_file}")
+    stake_addr=$(cat "${stake_addr_file}")
```
rdlrt, line 488: should we check if stake_addr_file exists (I understand that for 90% of cases it would, but there are few stuntmen around in workgroup who fudge the working by manually copying existing files and miss copying stake one)

I guess we can merge it for now and see if a case is required in future

hodlonaut: Duly noted, we should add stake_addr checks for a few different operations as a part of selectWallet - i.e. wallets that haven't been registered for staking shouldn't be shown on a few screens such as 'FUNDS -> DELEGATE -> SOURCE WALLET'

PR #245, opened 2020-06-20
[CNTOOLS] Wallet/Pool selection filter & quick navigation

Scitz0:
* Wallet & pool selection helper functions removed and getDirs() added. This to be able to do custom wallet and pool selection menu filtering in cntools.

* Relevant wallet balance and delegation status added directly to menu for each wallet entry.

* Only wallets and pools that should be selectable are now displayed.

* Quick navigation added, first shortcut key match is checked. If no shortcut key match is found it will see if there is an entry where the first letter match pressed key and if so jumps to that location.

The extra information and filtering does put some extra load before menu can be displayed. Have to check if this extra time is acceptable or not when there are lots of wallet and pools to display.

### DIFF
--- a/scripts/cnode-helper-scripts/cntools.config
+++ b/scripts/cnode-helper-scripts/cntools.config
@@ -31,6 +31,9 @@ POOL_PLEDGECERT_FILENAME="pledge.cert"
 POOL_CURRENT_KES_START="kes.start"
 POOL_DEREGCERT_FILENAME="pool.dereg"
 
+# limit for extended wallet selection menu filtering (balance check and delegation status)
+# if more wallets exist than limit set these checks will be disabled to improve performance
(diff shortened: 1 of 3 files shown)


Scitz0:
Please try before merging as it includes possible bug introductions with changes in all cntools wallet & pool selection menus.

As mentioned in PR note,
> The extra information and filtering does put some extra load before menu can be displayed. Have to check if this extra time is acceptable or not when there are lots of wallets and pools to display.

This could be valuable to check before merging.

Scitz0:
I have a commit pending that does a couple of optimizations and adds a configuration parameter that disables balance/delegation checks in wallet list if there are more than 10 wallets (default value).
It still does some key verification checks to only list valid wallets.

Just have to do some final verifications.

PR #246, opened 2020-06-22
Fixed 'typos' that could trip up noobs (or possibly even non-noobs) -…

kiwipool:
… missing \, upper case letters that should be lower & change opcert to op.cert

🥝


rdlrt:
LGTM! Thanks @kiwipool

PR #256, opened 2020-06-23
Added libsodium-dev to "sudo apt-get install" for "Debian/Ubuntu"

kiwipool:
Added libsodium-dev to "sudo apt-get install" for "Debian/Ubuntu" section in readiness for compilation of cardano-node 1.14.0.

I'm not sure EXACTLY what the equivalent package name(s) are for CentOS/RHEL/Fedora so haven't added these. However, I think they might be called:

Fedora - libsodium-devel
CentOS / RHEL - epel-release libsodium-devel

Not sure why the "Fixed 'typos'" typos pull is also showing here as it has been merged already?

matticus 🥝

### DIFF
--- a/docs/Setup-Node.md
+++ b/docs/Setup-Node.md
@@ -33,10 +33,10 @@ cardano-node run \
           --host-addr 0.0.0.0 \
           --port 5001 \
           --socket-path $CNODE_HOME/sockets/node0.socket \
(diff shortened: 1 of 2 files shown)


cardano-community:
Thanks, but it was already present in master - added as part of #254 :)

rdlrt:
It was already added on master, part of #254 😄

PR #276, opened 2020-06-30
topologyUpdater.sh: Setting PATH trips some users, some other cleanup

rdlrt:


rdlrt:
Would ${HOSTNAME} be a better alternative?
Given that current default gives an error as default - for ones who dont read ;)

Issue #277, opened 2020-06-30
[CNTOOLS] - Should registerStakeWallet be a seperate menu?


rdlrt:
As the topic says, just to have a discussion about whether we should have registerStakeWallet calls as a seperate call.

**Preface assumptions:**
- Network was changed
- Keys remain compatible (valid) , address hashing remains same too

_For Pools keys:_
If a pool was previously registered on old network , and not on new network... pool will be registered again by recreating/registering new keys
So that part is good

_For Stake key used by Pool:_
Currently cntools checks if there's a staking wallet address, before you can register pool...But we do not (might be difficult) check if staking address is registered..

_Proposal:_
1. Wallet -> New creates both payment and stake keys, a new task to register wallet is added. This aligns with Pool creation, and as long as inputs are valid keys , Key/Pool Registration can be executed multiple times.
2. Wallet -> New creates both payment and stake keys, then the registration task can be called in corresponding target (eg: Register Pool)
3. When doing Wallet -> New -> Stake, we just create keys, not register them. When Registering pool/delegating, we register the existing stake keys
4. Counter argument being we should encourage people to create new wallets every time network changes.

gufmar:
I run into this multiple times, solving it by completely swiping Priv folder with every new genesis.

Imo it would make sense as an additional menu entry. Leave the existing process for new base wallets as is, but have a new option to reuse(register) old wallets

Scitz0:
My take on it is this:

Moving registration of wallet to pool registration seems a bit weird to me. I do however recognize the issue presented.

As suggested in proposal 2 I think joining `New » Payment` and `New » Stake` into a single action is a good solution.
When doing this the wallet registration part should be split into its own action. This aligns better with how pool creation and registration work. The creation part just creates the keys needed and the registration part is its own thing and can be reused on network restart given that the keys are compatible. 
I'm not sure how to handle this last part though

papacarp:
I would vote for new wallets every time network changes.  I like the idea of keeping CNTOOLS as simple as possible to avoid bugs and support issues.  We have enough to do to just stay abreast of the cli changes and feature additions to cardano-node.  Just my 0.02ADA...

Scitz0:
I see your point papacarp and I agree, best practice to avoid as many issues as possible is to wipe both wallets and pools every time network resets to avoid any issues related to key changes etc. 

[...]

PR #286, opened 2020-07-03
CNTools wallet redesign [breaking change]

Scitz0:
Support for payment address(aka enterprise) removed and only base address support left. Everything that can be done with a payment address can also be done with the base address, ie doesn't need to be registered on-chain unless its delegated or used as a pledge wallet. This is done to present fewer options to end-user and remove confusion regarding the different address types. Because of this, the commit is set as a breaking change and major version bump to 1.0.

Previous payment wallets(ie not upgraded to stake wallet) won't be listed in CNTools as they are missing the stake keys. Either send all funds to a stake wallet before upgrading or use an external script to move funds afterward, ie for example with sendADA script.

Wallet registration is removed as a manual action and is now done automatically when delegating the wallet or using it as a pledge wallet when creating/modifying a pool. This also goes for genesis address, use an external script to send funds to a CNTools created wallet.

The payment and base address types were rooted deep in the code and as such, removing payment address support resulted in major code changes with possible breaking functionality. Please test thoroughly before merging into master.
CNTools docs also require an update because of this and should be committed to PR branch before merging with master.
Label do-not-merge set because of this and can be removed after docs update and tests have been performed. 

To address issue #277

### DIFF
--- a/docs/Scripts/cntools-common.md
+++ b/docs/Scripts/cntools-common.md
@@ -4,15 +4,15 @@ CNTools contains more functionality not described here.
 
 Step by Step guide to create a pool with CNTools
 
-* [Stake Wallet](#stake-wallet)  
-a stake wallet is needed for delegation/pledge
+* [Create Wallet](#create-wallet)  
+a wallet is needed for pledge and to pay for pool registration fee
 * [Create Pool](#create-pool)  
 create the necessary pool keys 
 * [Register Pool](#create-pool)  
 register the pool on-chain
 
 
-#### Stake Wallet
+#### Create Wallet
 
 **1.** `Choose Wallet [w]` and you will be presented with the following menu:
 ```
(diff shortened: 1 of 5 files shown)


Scitz0:
Docs and minor output update in the last commit. Should be ready for merge if the majority agree on this wallet redesign.

gufmar:
Pre-announcing and documenting is "key" in this situation, I would say.
Wonder if it's worth adding an "Warning: after upgrading .... ... I have read and understood" warning message in the integrated update menu

rdlrt:
Seeing a bit of drift between removal of payment, perhaps we can reduce the visibility but not completely discard it - having an option to be able to receive funds without having to affect the stake is good - will be lost in absence of this commit.
Also, given the lack of reading from a typical end user, it might be best to have that option for now anyways..

Scitz0:
> Seeing a bit of drift between removal of payment, perhaps we can reduce the visibility but not completely discard it - having an option to be able to receive funds without having to affect the stake is good - will be lost in absence of this commit.
> Also, given the lack of reading from a typical end user, it might be best to have that option for now anyways..

The drift is most likely due to the waitfornextblock step, not verifying that the balance match after new block.

The following modifications could help out:
* Add a balance check for last payment -> base address tx.
* Readd payment address balance in wallet list/show if it has funds, else omit. 

> Wonder if it's worth adding an "Warning: after upgrading .... ... I have read and understood" warning message in the integrated update menu

I have also been thinking about the update menu, how it could be improved to let user know what's new.
It would be nice to be able to add an update message, not sure where to store it though, inside library?. 
As we download library file anyway when checking for new version.
`CNTOOLS_VERSION_0.3_DESCRIPTION="Some text that describes what's new and if there are things to think about"`
If the user is at version 0.1 we print both descriptions for version 0.2 and 0.3. 

In addition to this, I think a message before showing start menu that a new version is available could be good, not everyone goes into the update menu just to see if there is a new version or not.

PR #299, opened 2020-07-07
cntools: improve delegators formatting

SmaugPool:
This improves significantly the legibility of delegators and stakes, and fix the formatting in logs too.


rdlrt:
Moved to #299

PR #306, opened 2020-07-08
cntools: identify owner stake and display in red if below pledge

SmaugPool:


scripts/cnode-helper-scripts/cntools.sh
```diff
@@ -2132,15 +2134,24 @@ case $OPERATION in
...
         printf "\r"
-        stake_address="581de0${key}"
         stake=$(jq ".esLState._utxoState._utxo | .[] | select(.address | contains(\"${key}\")) | .amount" "${TMP_FOLDER}"/ledger-state.json | awk 'BEGIN{total = 0} {total = total + $1} END{printf "%.0f", total}')
         reward=$(jq -r ".esLState._delegationState._dstate._rewards | .[] | select(.[0][\"credential\"][\"key hash\"] == \"${key}\") | .[1]" "${TMP_FOLDER}"/ledger-state.json)
         total_stake=$((total_stake + stake + reward))
-        say "$(printf "%-21s : %s" "Delegator ${delegator} hex key" "${key}")" "log"
-        say "$(printf "%-21s : ${CYAN}%s${NC} ADA (%s ADA)" " Stake (reward)" "$(numfmt --grouping "$(lovelacetoADA ${stake})")" "$(numfmt --grouping "$(lovelacetoADA ${reward})")")" "log"
-        delegator=$((delegator+1))
+        stake_color="${CYAN}"
+        if echo "${base_addr_hex}" | grep -q "${key}"; then
```
rdlrt, line 2145: This would report owner field only if pool delegated to has owner locally right?
Not an issue but just to confirm - because ledger-dump does provide owner hex for pools that we dont own too.

rdlrt: Nvm - its good addition, maybe we just add a comment for future with something like
# TODO : Check owner key using ledger

SmaugPool: Yeah, it needs `getRewardAddress` to work correctly. Maybe a check is missing to avoid issues.

SmaugPool: I now use the `rewardAccount.credential.key hash` from the ledger. 
Do you think it will work with multi-owner? I'm not sure :thinking:

SmaugPool: I now use the owners from ledger. I don't check the pledge for multi-owner yet though as this would require to delay the output.

Scitz0:
Please also add your change to the newly added docs cntools-changelog.md file.
Release 2.0.0. This way we don't have to collect all changes at merge time later.

SmaugPool:
@Scitz0 done, thanks!

PR #316, opened 2020-07-10
Onepointonefive

papacarp:
just a quick update to get cntools so it can generate a wallet address with 1.15.1.


cardano-community:
Covered by #317 which targets genesis branch

PR #317, opened 2020-07-11
Onepointonefive

cardano-community:
#316  - but points to genesis branch, so that its part of #298


rdlrt:
Changes are incomplete, and base is not up to date with genesis branch, will recreate as a new PR

PR #338, opened 2020-07-16
Allow metadata from existing url as well as new

papacarp:
As mentioned briefly in the guild chat, I'd like to be able to specify an external URL rather than have CNTOOLS create the metadata and require me to upload it to a server.  This PR shifts the order of questions slightly in both create pool and update pool flows to allow this as follows:

1.  we request the URL where metadata is hosted
2.  we test if there is a file at that location.  If no file, proceed to step 5
3.  If there is a file at that location, we ask if the user wants to use that file rather than enter in data through cntools.  If they say no, then proceed to step 5
4.  use the file at the URL specified to calculate the hash and skip to relay capture
5.  gather name, ticker, homepage, description as usual and then go to relay capture

### DIFF
--- a/scripts/cnode-helper-scripts/cntools.sh
+++ b/scripts/cnode-helper-scripts/cntools.sh
@@ -1277,7 +1277,7 @@ case $OPERATION in
       fi
 
 
-      say "\n${ORANGE}Please make sure you host your metadata JSON file (with contents as below) at ${meta_json_url} :${NC}\n"
+      say "\n${ORANGE}Please host file ${pool_meta_file} as-is at ${meta_json_url} :${NC}\n"
       say "{\n  \"name\": \"${meta_name}\",\n  \"ticker\": \"${meta_ticker}\",\n  \"description\": \"${meta_description}\",\n  \"homepage\": \"${meta_homepage}\"\n}" | tee "${pool_meta_file}"
     fi
 
@@ -1736,7 +1736,7 @@ case $OPERATION in
       fi
 
 
-      say "\n${ORANGE}Please make sure you host your metadata JSON file (with contents as below) at ${meta_json_url} :${NC}\n"
+      say "\n${ORANGE}Please host ${pool_meta_file} as-is at ${meta_json_url} :${NC}\n"
       say "{\n  \"name\": \"${meta_name}\",\n  \"ticker\": \"${meta_ticker}\",\n  \"description\": \"${meta_description}\",\n  \"homepage\": \"${meta_homepage}\"\n}" | tee "${pool_meta_file}"
     fi
 
--- a/scripts/cnode-helper-scripts/cntools.sh
+++ b/scripts/cnode-helper-scripts/cntools.sh
@@ -1207,7 +1207,6 @@ case $OPERATION in
     meta_homepage="https://foo.com" #default homepage
     meta_json_url="https://foo.bat/poolmeta.json" #default JSON
     pool_meta_file=${POOL_FOLDER}/${pool_name}/poolmeta.json
-    pool_meta_save_location=${POOL_FOLDER}/${pool_name}/
     if [[ -f "${pool_config}" ]]; then
       if [[ "$(jq -r .json_url ${pool_config})" ]]; then
         meta_json_url=$(jq -r .json_url "${pool_config}")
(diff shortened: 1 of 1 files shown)


scripts/cnode-helper-scripts/cntools.sh
```diff
@@ -1255,10 +1223,64 @@ case $OPERATION in
...
-    
-    say "\n${ORANGE}Please make sure you host your metadata JSON file (with contents as below) at ${meta_json_url} :${NC}\n"
-    say "{\n  \"name\": \"${meta_name}\",\n  \"ticker\": \"${meta_ticker}\",\n  \"description\": \"${meta_description}\",\n  \"homepage\": \"${meta_homepage}\"\n}" | tee "${pool_meta_file}"
-    
+
+    metadata_done=false
+    if wget -q  "$meta_json_url" -P "$pool_meta_save_location"; then
+      say "\nMetadata exists at URL.  Use existing data?\n"
+      case $(select_opt "[y] Yes" "[n] No") in
+        0) filename=$(basename "$meta_json_url")
```
rdlrt, line 1231: Can we leave the name saved within cntools to be poolmeta.json, helps with the consistency as well as troubleshooting support queries. In case of which ${pool_meta_file} is already set to the right value.

scripts/cnode-helper-scripts/cntools.sh
```diff
@@ -1255,10 +1223,64 @@ case $OPERATION in
...
+
+    fi
+    if [ ${metadata_done} = false ]; then
+
+      if [[ -f "${pool_meta_file}" ]]; then
+        meta_name=$(jq -r .name "${pool_meta_file}")
+        meta_ticker=$(jq -r .ticker "${pool_meta_file}")
+        meta_homepage=$(jq -r .homepage "${pool_meta_file}")
+        meta_description=$(jq -r .description "${pool_meta_file}")
+      fi
```
rdlrt, line 1246: During this validation, as per tests from @hodlonaut  - while testing large dummy JSON queries, we found that not all fields were extracted by cardano-wallet (thus, I assume SMASH too). Perhaps a #TODO comment to align with wallet in future - in case of discrepancy

scripts/cnode-helper-scripts/cntools.sh
```diff
@@ -1694,8 +1685,61 @@ case $OPERATION in
...
     fi
 
-    say "\n${ORANGE}Please host ${pool_meta_file} file as-is at ${meta_json_url}!${NC}\n"
-    say "{\n  \"name\": \"${meta_name}\",\n  \"ticker\": \"${meta_ticker}\",\n  \"description\": \"${meta_description}\",\n  \"homepage\": \"${meta_homepage}\"\n}" > "${pool_meta_file}"
+    metadata_done=false
+    if wget -q  "$meta_json_url" -P "$pool_meta_save_location"; then
+      say "\nMetadata exists at URL.  Use existing data?\n"
+      case $(select_opt "[y] Yes" "[n] No") in
+        0) filename=$(basename "$meta_json_url")
+           pool_meta_file=${pool_meta_save_location}${filename}
```
rdlrt, line 1693: Same as before

scripts/cnode-helper-scripts/cntools.sh
```diff
@@ -1694,8 +1685,61 @@ case $OPERATION in
...
+
+    fi
+    if [ ${metadata_done} = false ]; then
+
+      if [[ -f "${pool_meta_file}" ]]; then
+        meta_name=$(jq -r .name "${pool_meta_file}")
+        meta_ticker=$(jq -r .ticker "${pool_meta_file}")
+        meta_homepage=$(jq -r .homepage "${pool_meta_file}")
+        meta_description=$(jq -r .description "${pool_meta_file}")
+      fi
```
rdlrt, line 1707: Same as before (TODO comment to align with smash in future)

scripts/cnode-helper-scripts/cntools.sh
papacarp: thats fine.  It would also avoid any accidental overwrites of pool files if the user does something dumb like link to a file called `cold.skey` for example.  I can push up a change to do this today.

scripts/cnode-helper-scripts/cntools.sh
papacarp: I'm not clear what you want changed here.  just add a `#TODO` comment around this section of code?

papacarp: added a simple comment

PR #339, opened 2020-07-16
[CNTools] - Compatibility with Combinator

rdlrt:
Tasks:
- [x] Add HASH_IDENTIFIER where applicable to differentiate between network modes for commands used, required due to legacy Byron considerations
- [x] Manage config - add ptn0-praos.yaml, make prereqs default to combinator, and accept p argument to indicate praos mode
- [x] Add changes introduced as of cardano-node 1.16, once available
- [x] Add extended metadata field to cntools register/modify
- [x] Check 512b file size of metadata file
- [x] Add timestamp (nonce) to pool metadata file to make every creation unique.
- [x] Update doco to explicitly ask deleting private folders when upgrading node
- [x] Update changelog


rdlrt:
Wait for [this](https://github.com/input-output-hk/cardano-node/pull/1486) to be tested before merging - if it needs additional handling in cntools for combonator networks

PR #343, opened 2020-07-19
[Jormu] ITN Witness Script + Doco

rdlrt:
Doco for optional parameter to add ITN Witnesses to metadata JSON, to be consumed by cntools when registering metadata


rdlrt:
Best to merge for now - to prevent incomplete page for 5.0, can refine later as desired

PR #344, opened 2020-07-20
Bumped up default fixed pool cost; using ESC for Cancel now instead of c

hodlonaut:


rdlrt:
closing unfortunately due to target branch being master, will cause conflicts with existing PR, can re-raise to combonator branch

PR #350, opened 2020-07-22
minor tweaks

papacarp:


Scitz0:
It might not be clear for everyone what pooltool is(even if I doubt it). 
Add a link to your site for more information about it?

PR #389, opened 2020-07-29
CNTools [5.3.0]

Scitz0:
##### Added
- Ability to select a different pool owner and reward wallet
- Multi-owner support using stake vkey/skey files
- Added TIMEOUT_LEDGER_STATE in cntools.config to be used instead of static 60 seconds for querying shelley ledger-state.
- Option to delete private keys after successful backup
- More explicit error messages at startup

##### Changed
- POOL_PLEDGECERT_FILENAME removed from config, WALLET_DELEGCERT_FILENAME is used instead for delegation cert to pool, no need to keep a separate cert in pool folder for this, its the wallet that is delegated.
- Wallet vkeys no longer encrypted as skeys are the only ones in need of protection
- Pool >> Show view updated to show modified pool values if Pool >> Modify has been used to update pool parameters
- Update command change (change applied after this release is active):
  - Minor/Patch release: it will warn, backup and replace CNTools script files including cntools.config
  - Major release: No change, prompt user to backup and run prereqs.sh according to instructions.
- Troubleshooting improvements:
  - Split 'config in json format' and 'hasPrometheus' checks
  - Output node sync stats if Shelley transition epoch is to be calculated
  - Protocol parameters output check to give an improved error message
- Pool >> Show view updated to show modified pool values if Pool >> Modify has been used to update pool parameters
  - The section has also been updated to make it a little bit easier to read
- Pool >> Delegators view also use updated pledge value if a pool modification has been registered to check if pledge is met
- Use mainnet as default, since other testnets are either retired or not being maintained :(
- Backup original files when doing upgrades, so that users do not lose their changes.

##### Fixed
- Mainnet uses dedicated condition for slot checks
- Timeout moved to a variable in cntools.library
- KES Calculation for current KES period and KES expiration date
  **Please re-check expiration date using Pool >> Show**


rdlrt:
@Scitz0 - need to avoid adding changes that will break functionality for previously registered pools...We can work with additional checks or extra variables instead if needed

PR #413, opened 2020-08-10
Update cntools.sh

Straightpool:
Use srm instead of rm to securely delete private keys after backup (https://manpages.ubuntu.com/manpages/xenial/man1/srm.1.html)

### DIFF
--- a/docs/Build/node-cli.md
+++ b/docs/Build/node-cli.md
@@ -49,10 +49,3 @@ To start the node in passive mode, you can use the pre-built script below:
 cd $CNODE_HOME/scripts
 ./cnode.sh
 ```
-
-Once you have registered a pool, you might want to use the script below, edit the file to update the paths to the files/keys before execution. If you're using defaults, you only need to specify a POOLNAME to the start of the file:
-
-```bash
-cd $CNODE_HOME/scripts
-./cnode.sh
-```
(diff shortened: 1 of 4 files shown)


Straightpool:
Great idea on the more graceful integration, pull request updated per request.

rdlrt:
We would want to wait for merge to make a minor version bump, alongwith updated changelog

Scitz0:
The solution I propose is to create a delete file helper function that does this:
* Check if the file has +i flag from chattr just like decrypt function does and if so unlock it. 
* After that, check if srm is available on the system and if so use that to delete the file. If not it will use rm -f just like before. 
* Do this silently without warning messages to not freak end-user out with warning messages who do not have srm installed or want to install it.

Changelog message should clearly state that srm needs to be installed by user manually or by running prereq to have this added security function enabled. By doing it like this we remove additional built-in steps and questions for end-user.

Scitz0:
Another thing to keep in mind is that for ssd drives that most people use these days a program like srm has little to no use. 
https://www.howtogeek.com/234683/why-you-cant-securely-delete-a-file-and-what-to-do-instead/

This makes me more skeptical thinking about it if we should actually implement a feature like this, which might give a false sense of security.

mark-stopka:
> Another thing to keep in mind is that for ssd drives that most people use these days a program like srm has little to no use.
> https://www.howtogeek.com/234683/why-you-cant-securely-delete-a-file-and-what-to-do-instead/
> 
> This makes me more skeptical thinking about it if we should actually implement a feature like this, which might give a false sense of security.

Same goes for HDDs with copy-on-write or journal protected filesystem, you always need to use secure errasure for the entire drive / drives in a same RAID group, see NIST Special Publication 800-88 for details.

Scitz0:
It's correct that modern file systems also make deleting single files securely difficult. It shouldn't have any negative effects though even if the value to add it is limited. Please read the comment attached to commit added to move delete into a helper function for further info.

PR #419, opened 2020-08-12
added restart script

skokasik:


rdlrt:
Thanks for the PR, we'll revisit this shortly with systemd service integration, as it would make sense to keep things in sync

skokasik:
> Thanks for the PR, we'll revisit this shortly with systemd service integration, as it would make sense to keep things in sync

Appreciate the respond Priyank. I figure this is probably not the best way to ensure it is running. it is a bandage solution. systemd is the way to go.

rdlrt:
Operations for systemd added in #440

PR #440, opened 2020-08-23
CNTools - 5.4.0

### LINKED ISSUE #435: CNTools - remove delegators option?
The current implementation parsing ledger-state dump for delegators is very RAM intensive, slow, and "incorrect" as it doesn't take into account re-delegations correctly.

Keeping it as is might confuse if compared to the various online community/official explorers that show this information in a much better way. There is also a risk of crashes due to OOM issue dumping/parsing the ledger.

Should we remove this code from CNTools and if we feel there is merit, re-add it at a later point with improved implementation?

Scitz0:
##### Added
- Sanity check before launching a second instance of cnode.sh [eaa16f6]

##### Removed
- `Pool >> Delegators` removed. [closes #435]
  - If/when a better option than dumping and parsing ledger-state dump arise re-adding it will be considered.
  - Utilize the community explorers listed at https://cardano-community.github.io/support-faq/#/explorers

##### Fixed
- Block Collector script adapted for cardano-node 1.19.0.
  - Block hash is now truncated in the log, issue https://github.com/input-output-hk/cardano-node/issues/1738
- High CPU usage reported in a few cases when running Block Collector
  - Depending on log level, parsing, and byte64 enc each entry with jq could potentially put a high load on weaker systems. Replaced with grep to only parse entries containing specific traces.
- Docs for creating systemd block collector service file updated to include user env in run command

### DIFF
--- a/docs/Build/node-cli.md
+++ b/docs/Build/node-cli.md
@@ -50,3 +50,59 @@ cd $CNODE_HOME/scripts
 ./cnode.sh
 ```
 
+##### Run as systemd service
+
+The preferred way to run the node is through a service manager like systemd. This section explains how to setup a systemd service file.
+
+**1. Create systemd service file** 
+Replace `$USER` with the correct user for your system. Copy & paste all code below to create the service file.
+``` bash
+sudo bash -c "cat << 'EOF' > /etc/systemd/system/cnode.service
+[Unit]
+Description=C
+After=network.target
+
+[Service]
+Type=simple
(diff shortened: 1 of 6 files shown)


gufmar (approved):
LGTM

rdlrt:
Since we adding minor version change, would be good to add systemd instructions here.

Will find a time tonight to do so, unless someone gets there before

rdlrt (approved):
LGTM!

PR #444, opened 2020-08-24
[CNTools] More robust cnode.sh socket check

Straightpool:
When a Cardano node was not cleanly shut down, so the socket file still exists, cnode.sh will now delete the socket file before starting the node provided there is no cardano-node process running.


Scitz0 (approved):
LGTM

Scitz0:
Approved PR but some thoughts to consider. There could be more than one cardano-node process running on the system. Maybe better to grep for the specific socket file as it's included in the ps output.

Straightpool:
Good idea, updated grep.

Straightpool:
Alright using ps now instead of lsof, w´changed it though to look exactly for this one socket file or it will again not catch if there are multiple nodes running on one host

PR #451, opened 2020-08-31
Guild LiveView - RC1

Scitz0:
A replacement/compliment to the recently added sLiveView and the official built in cardano-node LiveView. The utility work independently of LiveView/SimpleView configuration and is as such a good way to get visual node metrics when cardano-node is deployed through systemd in SimpleView.

Inspiration is taken from a mix of community scripts(Adam@BUFFY Simple LiveView, CNTools, Martin@ATADA pingNodes) combined with the original LiveView. Metrics are taken from node EKG API combined with network tools to query peers. If CNTools Block Collector is enabled and used current epoch blocks are also displayed for core node. System metrics are intentionally left out for now as these are better monitored using pre-existing Linux tools like top, atop, htop, iotop etc.

Known issues:
- [x] Screen too small / resize can mess up output in some situations.
- [x] Offline tip ref & chain epoch calculation is coded for MainNet, add support for TestNets.
- [x] When node isnt in sync to shelley era, the prompt and errors remain hidden due to `tput`

image

image

Side - optimisations:
- [x] Move systemd install instructions to deploy-as-systemd.sh file
- [x] Make use of CNODE_NAME uniformly across the script to start using environment variables and folder names that begin with CNODE_NAME in respective cases.
- [x] Add`bc` to prereqs.sh
- [x] Move list of OS packages to pkg_list variable


Scitz0:
Please hold off on merge until properly tested and I have had time to fix shell check issues.

rdlrt (approved):
Good to go

PR #456, opened 2020-09-04
Resume logrotate functionality

### LINKED ISSUE #446: Logging cleanup
Guild config is set to log using json format and the internal log rotation built into cardano-node. As the internal log rotation is broken currently old log files are never removed. This quickly fills up the hard drive. The cnode.sh script on restart move old log files to an archive folder but never actually does any cleanup. 

I see a couple of options for how to solve this.

1) Replace json logging with stdout and let systemd service handle the logging through syslog. Breaks support for CNTools block collector though so not the best option.
2) Add a cronjob that clean old files from system, something like:
`1 0 * * * /usr/bin/find /opt/cardano/cnode/logs/ -mtime +7 -name "*.json" -print -exec /bin/rm {} \;`
Removes all files older than 7 days ending with .json in logs folder.
3) Do nothing, out of scope for CNTools/Guild to handle logging. Issue open against cardano-node logging hopefully fixed soon.

Any other options to consider for how to handle this?

rdlrt:
Closes #446 

Turns out node logging breaks for deleting old files because it uses hyphen as seperator to split filename with timestamp. Thus, remove the additional hyphen from config file templates, to allow log rotate functionality to work correctly.

### DIFF
--- a/files/ptn0-combinator.json
+++ b/files/ptn0-combinator.json
@@ -48,7 +48,7 @@
   "defaultScribes": [
     [
       "FileSK",
-      "/opt/cardano/cnode/logs/node-0.json"
+      "/opt/cardano/cnode/logs/node0.json"
     ]
   ],
   "hasEKG": 12788,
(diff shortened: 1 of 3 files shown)


gufmar (approved):
LGTM

alwaei:
Good

PR #457, opened 2020-09-04
gLiveView - Improved netstat parsing and peer analysis output

Scitz0:
* netstat parsing uniformity between peer analysis and live summary
* netstat parsing improved, would in some situation not parse incoming/outgoing correctly.
* peer analysis output timestamp made more defined and date skipped to make it clear that this is static information after analysis is done.
* peer analysis summary name changed from Peers Total to Unique Peers to make it clearer that multiple connections from same IPs are truncated.
* Skipped metric changed to be the difference between unique and total.


rdlrt:
The progress bar seems to be affected (going beyond the window) , will check further and report back

Scitz0:
Note for self, add filter for ekg/prom port on incoming connections.
Thanks to @ilap for pointing this one out.

rdlrt (approved):
LGTM

Issue #458, opened 2020-09-05
Guild LiveView - Future Requests


Scitz0:
Some gathered input from the community for how to improve Guild LiveView:

- Look into the output on mobile apps, does not seem to handle box-drawing characters properly in all cases. 
- Sanity check of the configuration values.
- Make the color scheme configurable.

rdlrt:
1. Font on mobile apps for UTF-8 characters often have different spacings. One of the ways would be replacing those indicators with standard ascii characters (eg: using # , = , _, - ) for the same

PR #463, opened 2020-09-06
gLiveView - Color themes, legacy mode and config check

Scitz0:
Addresses enhancements mentioned in #458

- New color theme config option has been added (dark|light). Colors have also been slightly modified to better suite different terminals. If user wants to override colors set in the two themes or add their own this should be easier as well.

- New legacy config option added. If enabled, unicode box-drawing characters will be replaced by standard ASCII characters. It was noticed that on mobile apps UTF-8 characters often have different spacings.

- Sanity checks added that configuration values are correctly set or identified.


rdlrt (approved):
Looks good, to merge after doco update
Awesome work @Scitz0

Issue #464, opened 2020-09-06
cardano-db-sync fails to build


TrevorBenson:
**Describe the bug**
cardano-db-sync fails to build as the instructions are written. It generates various errors when executing cabal-build-all.sh

```
undefined reference to 'crypto_vrf_proof_to_hash'
collect2: error: ld returned 1 exit status 
`gcc' failed in phase `Linker'. (Exit code: 1)

```
ending in:

```
cp: cannot stat '/root/git/cardano-db-sync/dist-newstyle/build/x86_64-linux/ghc-8.6.5/ouroboros-consensus-cardano-0.1.0.0/x/db-analyser/build/db-analyser/db-analyser': No such file or directory
Copying db-validator to /root/.cabal/bin/
cp: cannot stat '/root/git/cardano-db-sync/dist-newstyle/build/x86_64-linux/ghc-8.6.5/ouroboros-consensus-cardano-0.1.0.0/x/db-validator/build/db-validator/db-validator': No such file or directory
```

**To Reproduce**
1. Follow directions for https://cardano-community.github.io/guild-operators/#/Build/dbsync
2. Run $CNODE_HOME/scripts/cabal-build-all.sh
3. See error

**Expected behavior**
Running `cabal-build-all.sh` succeeds building cardano-db-sync and can be executed.

**Screenshots**
[https://pastebin.com/pAugFzVU](https://pastebin.com/pAugFzVU)
[https://pastebin.com/Pe7xZ2Un](https://pastebin.com/Pe7xZ2Un)

**Version:**
 - OS: Ubuntu 20.04
 - Product version: CNTools 5.4.0
 - Cardano Node version: 1.19.0
 - Network you're connecting to: 2017-09-23T21:44:51Z

**Additional context**
Using the following line from Node and CLI resolves the issue:
`echo -e "package cardano-crypto-praos\n  flags: -external-libsodium-vrf" > cabal.project.local`

PR #465, opened 2020-09-06
Fix cardano-db-sync build instructions

TrevorBenson:
This PR is intended to resolve issue #464


Scitz0 (approved):
LGTM with a comment.
For cardano-node build instructions there is also a special note for CentOS 7, should this be added for db-sync as well?
```
# On CentOS 7 (GCC 4.8.5) we should also do
# echo -e "package cryptonite\n  flags: -use_target_attributes" >> cabal.project.local
```

PR #470, opened 2020-09-10
[CNTOOLS 5.4.1] - Fix KES Expiry, closes #469

### LINKED ISSUE #469: CNTools - Improve KES expiration date calculation
Current formula to calculate KES expiration date in CNTools use this formula:
`expiration_time_sec=$(( current_time_sec - ( slot_length * slot_in_epoch ) + ( slot_length * slots_per_kes_period * remaining_kes_periods ) ))`
This is a bit flawed as it shouldn't subtract the full current epoch, at most one KES period as there are 3.33 KES periods per epoch. Shouldn't be based on epoch at all. Using the current formula and it will give a different date depending on when in the epoch you are. 
My suggestion for a new formula:
`expiration_time_sec=$(( current_time_sec - ( slot_length * slots_per_kes_period ) + ( slot_length * slots_per_kes_period * remaining_kes_periods ) ))`
Ie remove one full KES period instead to be on the conservative side.

Putting it in as an issue to not forget as I'm unable to create a PR currently. I also want to give it a second lookover.

Affects Guild LiveView date calculation as well.

rdlrt:


Scitz0 (approved):
LGTM

PR #473, opened 2020-09-11
Replace netstat with ss

### LINKED ISSUE #472: netstat is a deprecated utility that is not installed by default anymore, switch to using ss instead.
**Describe the bug**
The nestat utility is not installed by default anymore. Some distributions are even renaming the package net-tools to net-tools-deprecated, adding complication to the yum/apt installation logic. Instead of creating additional conditional logic to install netstat it makes more sense to simply update gLiveView.sh (and other scripts) to use ss instead.

**To Reproduce**
1. Run guild-operators build and run node and cli
2. Execute gLiveView.sh
3. Notice inbound connections are not listing properly.

**Expected behavior**
gLiveView.sh lists inbound/outbound connections.

**Version:**
 - OS: opensuse 15.2
 - Product version: CNTools 5.4.0
 - Cardano Node version: 1.19.0

**Additional context**
I will be submitting a PR from my fork with the changes from Telegram user 'Ja J'.

Scitz0:
Closes #472

### DIFF
--- a/scripts/cnode-helper-scripts/gLiveView.sh
+++ b/scripts/cnode-helper-scripts/gLiveView.sh
@@ -107,6 +107,12 @@ while getopts :l opt; do
 done
 shift $((OPTIND -1))
 
+if ! command -v "ss" &>/dev/null; then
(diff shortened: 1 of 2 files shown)


rdlrt (approved):
LGTM

Issue #475, opened 2020-09-12
prereqs.sh does not properly catch fedora os


TrevorBenson:
**Describe the bug**
Fedora attempts to use apt to install package dependencies. This is caused because fedora does not use an ID_LIKE parameter. 

**To Reproduce**
1. Setup Fedora 32
2. Download prereqs.sh
3. Run prereqs.sh
4. See error where it attempts to use apt
5. grep -i ^id /etc/os-release
6. See there is only ID, no ID_LIKE

**Expected behavior**
Script to properly identify the OS and install the required dependencies with the correct package manager.

**Version:**
 - OS: Fedora 32
 - Product version: CNTools 5.4.0
 - Cardano Node version: 1.19.0

**Additional context**
Below is a patch that allows identifying the system as a centos/rhel like system when OS_ID parameter is empty.

```
--- ./prereqs.sh.orig	2020-09-12 08:13:57.579262488 -0700
+++ ./prereqs.sh	2020-09-12 08:58:16.097430171 -0700
@@ -110,7 +110,11 @@
 
   # Determine OS platform
   OS_ID=$(grep -i ^id_like= /etc/os-release | cut -d= -f 2)
+  OS_REALID=$(grep -i ^id= /etc/os-release | cut -d= -f 2)
   DISTRO=$(grep -i ^NAME= /etc/os-release | cut -d= -f 2)
+  if [ -z ${OS_ID} ] && [ ${OS_REALID} == 'fedora' ]; then
+    OS_ID='rhel'
+  fi
 
   if [ -z "${OS_ID##*debian*}" ]; then
     #Debian/Ubuntu
@@ -140,11 +144,15 @@
     $sudo rpm --import https://dl.yarnpkg.com/rpm/pubkey.gpg > /dev/null
     $sudo yum -y update > /dev/null
     echo "  Installing missing prerequisite packages, if any.."
-    pkg_list="python3 coreutils pkgconfig libffi-devel gmp-devel openssl-devel ncurses-libs ncurses-compat-libs systemd systemd-devel libsodium-devel zlib-devel npm yarn make gcc-c++ tmux git wget epel-release jq gnupg libtool autoconf srm iproute bc tcptraceroute"
+    if [ ${OS_REALID} == 'fedora' ]; then
[...]

PR #476, opened 2020-09-14
Prereqs update - closes #474 and #475

### LINKED ISSUE #474: Error: Package: yarn-1.22.5-1.noarch (yarn) Requires: nodejs
**Describe the bug**
./prereqs.sh failed
Following the directions here:  https://cardano-community.github.io/guild-operators/#/basics?id=pre-requisites

To Reproduce:
**On fresh EC2 Linux2 instance:**
mkdir "$HOME/tmp";cd "$HOME/tmp"
curl -sS -o prereqs.sh https://raw.githubusercontent.com/cardano-community/guild-operators/master/scripts/cnode-helper-scripts/prereqs.sh
chmod 755 prereqs.sh
./prereqs.sh

**Expected behavior**
I was previous able to install a working node following the documentation here:  https://docs.cardano.org/projects/cardano-node/en/latest/getting-started/building-the-node-using-nix.html     It didn't require anywhere near so many dependencies.  Is this project so scattered that node.js is required?  Libsodium AND openssl?

**Screenshots**

**Version:**
 - OS: Standard Linux2 EC2 instance spun up fresh just for this

rdlrt:
- Add special case for Fedora in pre-requisites
- Eliminate GraphQL dependencies from common pre-requisites , as it's an "odd-man-out" (does not follow rest of the components , besides feels forced usage of node for a very minimalistic outcome).
- Take backup before overwriting `topology.json`, `topologyUpdater.sh` and handle `cnode.sh` in a way that it can preserve port as well as information to start node as core (path to keys) v/s relay.

### DIFF
--- a/docs/Build/graphql.md
+++ b/docs/Build/graphql.md
@@ -1,4 +1,4 @@
-!> - An average pool operator may not require cardano-wallet at all. Please verify if it is required for your use as mentioned [here](build.md#components)
+!> We have temporarily disabled updating build documentation for Cardano-GraphQL. The specific component does not follow the process/technology/language (requires npm, yarn) used by other components (cabal/stack), and the value provided by `cardano-graphql` over the (haskell-based) hasura instance has been negligible. Also, an average pool operator may not require cardano-graphql at all, please verify if it is required for your use as mentioned [here](build.md#components). The instructions below are `out of date`.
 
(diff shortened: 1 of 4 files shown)


Scitz0 (approved):
The only comment would be that its generally best to omit quotes when using bash pattern matching (=~).
For a simple string match like the one here using quotes work fine, but I would change to for example:
`if [[ "${OS_ID}" =~ ebian ]]; then`

Other than this 👍

PR #481, opened 2020-09-14
Version control and trap improvement

### LINKED ISSUE #479: [gLiveView] - Versioning and auto updates
Wonder if it makes sense to extend version and - optionally, handle auto updates (very simplified version) to `gLiveView` as well

Scitz0:
Closes #479


rdlrt (approved):
LGTM!
just updated Minor version to 0-rc1 (in case we receive reports of any unforeseen issues while upgrading), and added backup before update (to ensure any customisations are not lost for good)

PR #490, opened 2020-09-19
Guild LiveView 1.2 - Peer list navigation

Scitz0:
- Added support to cycle through all peers using left/right arrow keys. 'o'(outgoing)/'i'(incoming) keys used to select active list to navigate.
- endline function replaced with printf fillers. Fixes flickering in certain conditions as line was first cleared and then redrawn.
- Improved handling of 0 peer lists
- Add uptime and title to Peer Analysis view
- Various minor improvements


rdlrt (approved):
LGTM, arrow key may not be as obvious to some, perhaps we can add a comment somewhere - but rest is good

Scitz0:
> LGTM, arrow key may not be as obvious to some, perhaps we can add a comment somewhere - but rest is good

Good idea, added

PR #493, opened 2020-09-21
gLiveView 1.3 - Update and slot interval

Scitz0:
Internal update now replace CNODE_HOME with correct VNAME if a custom is used.

Improved slot interval calculation


rdlrt (approved):
LGTM!

PR #494, opened 2020-09-22
[gLiveView 1.4] - Reverse order for CNODE_HOME and SOCKET

rdlrt:
When setting user vars manually the SOCKET parameter should be specified after CNODE_HOME , unless mentioning the complete path


Scitz0 (approved):
👍

PR #498, opened 2020-09-24
gLiveView 1.6

Scitz0:
- Allow three failed connections in a row before closing app in case of a temporary glitch/restart.

- Clear screen when moving between core/relay mode, else display can get messed up.

- calculate KES expiration date in main loop as it can change in case of key rotation and node restart

- Change order in exit function to properly reset colors in some situations.


rdlrt (approved):
LGTM

PR #500, opened 2020-09-26
Revert change to cnode.sh

rdlrt:
We possibly missed why the order was changed in the first place.

If deploying as systemd, the env vars wont be resolved, and so the env file wouldnt be called either.

If customised is needed, it would CNODE_HOME variable's value changed when installing via `prereqs` or by hand.


Scitz0 (approved):
LGTM

PR #503, opened 2020-09-29
Closes #502 , Add Haskell Language Server flag to prereqs

### LINKED ISSUE #502: Prereq.sh fails to install cabal
Hi,

having trouble with Prereqs.sh

running script installs GHCUP and GHC but when installing cabal I get a waterfall of Please type yes or no and press enter

rdlrt:
ghcup installer has changed to include HLS server as a flag, defaulting to no in prereqs.sh


Scitz0 (approved):
LGTM

PR #506, opened 2020-09-30
Closes #504; Add cardano-address and bech32 to build process

### LINKED ISSUE #504: Update cabal-build-all if git repo is cardano-node, to add custom dependencies
1. Update cabal-build-all.sh to add cabal.project.local to the folder if git repo is cardano-node with correct targets
2. Handle auto updates as part of cabal-build-all.sh to the local file
3. Add argument to skip overwriting local file

rdlrt:
Update cabal-build-all.sh script to include latest cabal.project.local from guild repo (it is to ensure that we can maintain updated tags for cardano-address and bech32 binaries without impacting user steps), and add an overwrite flag to only overwrite cabal.project.local if specified.
The change only occurs if current folder is named `cardano-node` . Thus, it should match git repo being node one when following guild repo instructions.

### DIFF
--- a/docs/Build/node-cli.md
+++ b/docs/Build/node-cli.md
@@ -22,11 +22,8 @@ git fetch --tags --all
 git pull
 git checkout 1.19.0
 
-echo -e "package cardano-crypto-praos\n  flags: -external-libsodium-vrf" > cabal.project.local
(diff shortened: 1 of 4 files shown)


hodlonaut (approved):
Looks good to me

Scitz0 (approved):
👍

PR #509, opened 2020-10-02
Env patch

rdlrt:
- env: If using defaults, allow automatic detection of vars. Simplify customisations for users when changing env file
- prereqs: Add dialog to prereqs.sh, required by cntools 6.0.0
- prereqs: Backup env file before overwriting
- gLiveView 1.7

PR #513, opened 2020-10-07
Added a new user parameter 'RETRIES' to specify how many times gLiveV…

cloudstake:
…iew.sh should try to re-connect to the running cnode before it exits. This gives the option to increase tries to account for manual stop and start of cnode.service without the tool timining out


Scitz0 (changes requested):
LGTM but I would request that you move the RETRIES variable above THEME as it currently split the THEME comment/info about dark/light in the middle.

Scitz0 (approved):
👍  will be included in gLiveView 1.7 (#509) soon to be merged.

rdlrt:
Closing this as added to 509 already, thanks @cloudstake

PR #526, opened 2020-10-12
Make TopoUpdater updates and variable extraction uniform with repo

rdlrt:


Scitz0 (approved):
👍

gufmar (approved):
LGTM - great integration

gufmar (approved):
LGTM

gufmar:
There are two Updater-Service side changes that might be mentioned in the release notes.
The more important one: the fetch API endpoint now contains the IOG hydra default topology relay node for all invalid requests. (eg pending subscription, or temporary wrong blockNo) This makes the onboarding much easier and fail-safe.
As a second minor extension the returned topology list in addition to the distance on earth surface now also contains continent, country and region for every returned peer. This gives the SPO a better understanding of where its node is connected.

PR #527, opened 2020-10-12
cntools update and backup

Scitz0:
- removed dialog howto help text as it was mostly distracting
- backup removed from internal update as only cntools.sh and cntools.library are updated. Only cntools.config need to be saved and this is only updated through prereqs.sh
- cntools backup now include the entire files and scripts folders


rdlrt (approved):
LGTM!

PR #532, opened 2020-10-14
CNODE_HOME sed fix

Scitz0:
- add code block to topologyUpdater.sh to rename CNODE_HOME in case of custom VNAME
- update sed replacement in prereqs.sh and gLiveView.sh


rdlrt (approved):
LGTM!

PR #566, opened 2020-10-25
Add CCLI dirname to PATH even if CCLI found

rdlrt:
To ensure cardano-node binary is available in $PATH when specifying custom CCLI


Scitz0 (approved):
LGTM

PR #572, opened 2020-11-01
add cncli installation and run script

Scitz0:
CNCLI is developed by Andrew and available at https://github.com/AndrewWestberg/cncli
It's a RUST implementation for low-level interaction with cardano-node.
Current functionality include cardano-node ping and chain sync into sqlite db for different block queries.

use cncli-install.sh to install RUST, CNCLI and deploy it as a systemd service. Script can also be used to upgrade CNCLI. Updated prereqs.sh needed to install libsqlite3 and env updated with variable for path to CNCLI executable.

### DIFF
--- a/scripts/cnode-helper-scripts/gLiveView.sh
+++ b/scripts/cnode-helper-scripts/gLiveView.sh
@@ -57,18 +57,20 @@ PARENT="$(dirname $0)"
 
 usage() {
   cat <<EOF
-Usage: $(basename "$0") [-l] [-b <branch name>]
+Usage: $(basename "$0") [-l] [-p] [-b <branch name>]
 Guild LiveView - An alternative cardano-node LiveView
 
 -l    Activate legacy mode - standard ASCII characters instead of box-drawing characters
--b    Use alternate branch to check for updates - only for testing/development (Default: Master)
+-p    Disable default CNCLI ping and revert to legacy tcptraceroute if available, else use regular ICMP ping.
+-b    Use alternate branch to check for updates - only for testing/development (Default: Master)  
 EOF
   exit 1
 }
 
-while getopts :lb: opt; do
+while getopts :lpb: opt; do
   case ${opt} in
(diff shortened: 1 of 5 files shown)


scripts/cnode-helper-scripts/gLiveView.sh
```diff
@@ -390,21 +390,34 @@ checkPeers() {
     peerPORT=$(echo "${peer}" | cut -d: -f2)
     [[ -z ${peerIP} || -z ${peerPORT} ]] && continue
     
-    if [[ "${peerIP}" = "${lastpeerIP}" ]]; then
+    if [[ ${direction} = "out" ]]; then
+      if [[ -n ${CNCLI} && -f ${CNCLI} ]]; then
```
rdlrt, line 394: Just thinking, should we instead have a flag for CNCLI usage?
So that if Flag is set to Y and CNCLI is present in $PATH? Could be helpful for comparison at times or while waiting for a fix for example

Scitz0: The downside with the old method(tcptraceroute) is that from the remote cardano-node this is seen as a bad handshake connection and is logged with an error if these traces are active. So using cncli is a much cleaner way to ping as it does a proper handshake. This is why I set it as the primary method. I have done extensive testing using bad input(unresolvable dns, bad IP etc), seems to handle all of these cases gracefully.

rdlrt (approved):

scripts/cnode-helper-scripts/gLiveView.sh
rdlrt: All good to have it at default, but would 
be good to have a flag regardless - I simply dont trust IOG to not wake up one fine day and realise they want to change the handshake as part of future HFC event

Scitz0: true, adding option to disable cncli ping by flag

PR #576, opened 2020-11-04
CNTools 6.2.0, gLiveView 1.11, logMonitor

Scitz0:
CNTools 6.2.0
- Ability to post metadata on-chain, e.g. (but not limited to) Adams https://vote.crypto2099.io/
- Block view updated to match CNCLI / logMonitor update.
- chattr for file locking now optional to use, cntools.config setting added
- Minor tweak (non-breaking), unnecessary bech32 conversion in wallet import

logMonitor
- cntoolsBlockCollector.sh replaced with logMonitor.sh as a general purpose log monitoring script.
- code rewritten to only depend on env file instead of CNTools as before
- code utilizing the same pattern as the other scripts for retaining user config

gLiveView 1.11
- Updated block view to match CNCLI / logMonitor change.
- sync percentage added if shelley transition epoch is known

CNCLI
- cncli.sh script added with four subcommands available
  - **sync** : Start CNCLI chainsync process that connects to cardano-node to sync blocks stored in SQLite DB
  - **leaderlog** - Loops through all slots in current epoch to calculate leader schedule
  - **validate** - Confirms that the block made actually was accepted and adopted by chain
  - **ptsendtip** - Send node tip to PoolTool for network analysis and to show that your node is alive and well with a green badge
  - **migrate** - command to migrate old blocklog(cntoolsBlockCollector) to new format (post cncli)
- Meant to be run in conjunction with logMonitor.sh script as currently parsing logs is the only way to get the hash of a block supposed to have been created. In the future SQLite DB from CNCLI could be used to get blocks created by verifying against VRF. Still won't catch invalid blocks so there will always be a place for the logMonitor script.

Prereqs
- CNCLI installation added with `-c` flag that also install RUST to be able to compile.

Systemd deployment
- deploy-as-systemd.sh now handle deployment of all systemd services
  - cnode.service
  - cnode-logmonitor.service
  - cnode-cncli-sync.service
  - cnode-cncli-leaderlog.service
  - cnode-cncli-validate.service
  - cnode-cncli-ptsendtip.service
- All services depend on cnode.service so all actions on this main service is replicate to all child/sub services.
- A delay of 30s added to all subservices to let cardano-node properly wake up

Docs updated to match changes. (70% done... need rescan as code has been re-written several times)


Scitz0:
Should be ready and fully working... just need to look over and update documentation

rdlrt (approved):
Have not had a chance look at in detail, but I trust Ola's judgement on this - as cncli is mainly his work with Andrew's support. So will let him decide when to merge, will hopefully be able to do some testing before merged to master

PR #578, opened 2020-11-09
Replace leftover HASH_IDENTIFIER

rdlrt:


Scitz0 (approved):
Until we can solve it in a better way this is fine.

Issue #588, opened 2020-11-16
Sample Service/Timer Units: topologyupdater.


TrevorBenson:
cnode-topoologyupdater.service
```
[Unit]
Description=CNODE topology updater
After=network-online.target

[Service]
Type=oneshot
User=cardano
RemainAfterExit=true
WorkingDirectory=/opt/cardano/cnode/scripts
ExecStart=/bin/bash -l -c "exec /opt/cardano/cnode/scripts/topologyUpdater.sh"
KillSignal=SIGINT
StandardOutput=syslog
StandardError=syslog
SyslogIdentifier=cnode-topologyupdater
KillMode=mixed

[Install]
WantedBy=multi-user.target

```

cnode-topologyupdater.timer
```
[Unit]
Description=Wake up cnode-topologyupdater.service once an hour hour
Requires=cnode-topologyupdater.service

[Timer]
OnUnitActiveSec=1h 30sec
AccuractySec=1us

[Install]
WantedBy=cnode-topologyupdater.service timers.target
```

Scitz0:
This is exactly how I imagined implementing the timer service for topoloyUpdater. 
deploy-as-systemd.sh script handle all the systemd deployments so would be best added here.

We also need to separate the fetch and push in topologyUpdater.sh so that we can integrate the fetch in cnode.sh and also modify the cnode.service file with RuntimeMaxSec to restart the node at a given interval, asking the user when running deploy-as-systemd.sh if it's a relay and how long to wait between restarts. User can easily just rerun the deployment script to change the timer or disable it.

TrevorBenson:
I must have typod while editing the content. I changed it to say "cnode-topologyupdater.service" everywhere instead of my original service unit name so it was more in line with your service unit naming conventions. Sorry about that.

Onto the restart of cnode.service. I find hardcoding a `RuntimeMaxSec` and Restart=always into the cnode.service a little less elegant of a solution than I would hope for. You mention the user can use the script to change or disable the timer. What methods does this follow, does it disable the service make the change and then enable it again? If not, how thoroughly have you tested changing or disabling the RuntimeMaxSec without actually restarting the service unit? Below I discuss some issues I've observed via systemd.

That said I haven't had time to rebuild one of the testnet servers to try out the alpha branch yet. So please take the following as general observations, not an attack on your implementation :+1:.

**Scenario**
[...]

PR #591, opened 2020-11-20
Topology Updater systemd deployment based on input from #588

Scitz0:
Changes in PR is done with full backwards compatibility for old crontab job style execution. By default, topologyUpdater.sh still does a combined push/fetch call when run. Two new flags added, `-f` & `-p` to either disable fetch or push. Added to be able to separate the calls.

Adds option in `deploy-as-systemd.sh` script to deploy topologyUpdater.sh as systemd services. This will deploy both push & fetch oneshot service files as well as timers for a scheduled 60 min node alive message and cnode restart at the user set interval when running the deploy script.
```
- cnode.timer            : handles the scheduled restart of cardano-node(cnode.sh), default every 24h
- cnode-tu-fetch.service : fetches a fresh topology file before cnode.service file is started/restarted
- cnode-tu-push.service  : pushes a node alive message to Topology Updater API
- cnode-tu-push.timer    : schedules the push service to execute once every hour
```
env now takes an 'offline' arg to replace the previous static check when sourced. This change was required for some scripts to be able to source the entire env file even when there is no node alive.

Topology Updater docs updated to reflect changes.

### DIFF
--- a/docs/Scripts/topologyupdater.md
+++ b/docs/Scripts/topologyupdater.md
@@ -8,7 +8,6 @@ If you have run [prereqs.sh](basics.md#pre-requisites), this should already be a
 
 Before the updater can make a valid request to the central topology service, he must query the current tip/blockNo from the well synced local node. It connects to your node through the configuration in the script as well as the common env configuration file. Customize these files for your needs.  
 
-
 To download topologyupdater.sh manually you can execute the commands below and test executing topology Updater once (it's OK if first execution gives back an error):
 ``` bash
 cd $CNODE_HOME/scripts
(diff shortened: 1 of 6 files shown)


gufmar (commented):
Lgtm

Awesome work again.

rdlrt (approved):

TrevorBenson:
LGTM as well. I like the timer units and how fetch is tied to the `${vname}.service`. I believe not having the After allow the node startup to be successful even when the fetch fails.

That said I'm pretty tired and could be misremembering, have any unit tests been run which cause fetch to fail/crash and confirms it does not prevent the `${vname}.service` from starting? If so that's the only thing that comes to my tired mind ATM. :wink:

Scitz0:
> LGTM as well. I like the timer units and how fetch is tied to the `${vname}.service`. I believe not having the After allow the node startup to be successful even when the fetch fails.
> 
> That said I'm pretty tired and could be misremembering, have any unit tests been run which cause fetch to fail/crash and confirms it does not prevent the `${vname}.service` from starting? If so that's the only thing that comes to my tired mind ATM. 😉

During testing, I had many failed topologyUpdater fetch calls due to buggy implementations and not once did I notice this preventing the main cnode.service from starting. This is due to the fact that the cnode-tu-fetch.service BindsTo directive is one-sided, meaning it requires that cnode.service is active(running) but cnode.service doesn't case if cnode-tu-fetch.service is running or not. The `Before=${vname}.service` setting only handles order of startup, not status of execution.

PR #595, opened 2020-11-23
WIP - gLiveView updates for 1.23.0

rdlrt:


Scitz0 (changes requested):
Looks good but we could replace `cut -d. -f 2` with `tr -d '.'` and compare against the full version number, ie then comparison becomes `-lt 1230`. If for some reason IOG should decide to bump the mayor version anytime soon.

Scitz0 (approved):
👍

PR #611, opened 2020-12-01
cncli.sh inherit from environment

TrevorBenson:
Allows inheriting from environment for cncli.sh variables. 

1. When POOL_DIR is defined automatically configures the ID and keys for the pool, if undefined it retains the existing logic.
2. When POOLTOOL API_KEY or TICKER are defined these are directly inherited, if undefined it retains the existing logic.


Scitz0 (changes requested):
1) We dont put code in the user variables section, this is strictly for user input. Move code to cncliInit() function where the other user variables get default values if not set.
2) Keep variables in User Variables section but commented by default.
3) To keep it consistent, please use ${} format for all variables even though not necessary.
4) POOL_DIR, I can't see this variable set anywhere? Is that a docker variable? Better to add to cncli as well in that case to keep it consistent.

TrevorBenson:
POOL_DIR was added to cnode.sh. When it exists cnode.sh knows where to find the keys to start up as a block producer instead of a relay.

TrevorBenson:
Hope that is more to your liking.

Scitz0:
> Hope that is more to your liking.

Getting there 😄. This has nothing to do with your PR but I see flaws in our previous implementation regarding wallet and pool filename convention. Ie in cntools.config we have the option to customize the filenames but then we statically set it in cnode.sh and here in cncli.sh. I propose we move all wallet and pool naming convention from cntools.config to env to have it common between all scripts and replace hardcoded filenames in the other scripts to use the variable name. But I can do this in a separate PR as I have an idea for how to do this and I'm currently working on CNTools 6.3.0. Will most likely move POOL_DIR to env as well, better to have it set in one place instead of having the same variable in multiple places. 

Approving PR as it is for now.

Scitz0 (approved):
👍  see my other comment regarding filename convention.

TrevorBenson:
>  see my other comment regarding filename convention.

I think that is a good idea to have shared variables as much as possible.

It made me wonder if we had any other cntools.config variables which might be useful to share. After a quick review the only other thing that stands out is `POOL_FOLDER`. I think cnode.sh and cncli.sh could potentially leverage this.

For cnode.sh a `POOL_DIR` implementation was put in place which uses a full path, but still using the hardcoded keys as you mentioned.  By sharing `POOL_FOLDER` into the env we could pass `POOL_NAME=Company_TCKR1` instead of `POOL_DIR=/opt/cardano/cnode/priv/pool/Company_TCKR1`. 

In which case cnode.sh would change:

```
[user@cardano-testnet scripts]# diff -u cnode.sh cnode.sh.w_pool_name 
--- cnode.sh    2020-12-05 09:33:18.000000000 +0000
+++ cnode.sh.w_pool_name        2020-12-05 21:34:50.935478276 +0000
@@ -11,7 +11,7 @@
 # Common variables set in env file   #
 
-#POOL_DIR="${CNODE_HOME}/priv/pool/TEST"                 # set pool dir to run node as a core node
+#POOL_DIR="${POOL_FOLDER}/TEST"                          # set pool dir to run node as a core node
 #TOPOLOGY="${CNODE_HOME}/files/topology.json"            # override default topology.json path
 
@@ -24,6 +24,12 @@
 [[ -z "${TOPOLOGY}" ]] && TOPOLOGY="${CNODE_HOME}/files/topology.json"
 [[ -z "${POOL_DIR}" ]] && POOL_DIR="${CNODE_HOME}/priv/pool/TEST"
 
+if [[ -z "${POOL_NAME}" ]] ; then
+  POOL_DIR="${POOL_FOLDER}/TEST"
+else
+  POOL_DIR="${POOL_FOLDER}/${POOL_NAME}"
+fi
+
 if [[ -S "${SOCKET}" ]]; then
   if pgrep -f "[c]ardano-node.*.${SOCKET}"; then
      echo "ERROR: A Cardano node is already running, please terminate this node before starting a new one with this script."
```

For cncli.sh not only would it know what the default key names are supposed to be if the user renames them, but with `POOL_FOLDER` + `POOL_NAME` it could autodetect which folder holds the keys, even if multiple folders exist under `priv/pool/`.

Scitz0:
I agree, this could be cleaned up a bit. Need to have a better name for it. 
Though your above suggestion is not based on the latest alpha where all variables from cnode.sh are moved to env.

Scitz0:
I'm adding a couple of improvements regarding this in the node-1.24 PR I will soon push.

PR #612, opened 2020-12-02
CNTools 6.3.0

Scitz0:
Changed
- printTable function replaced with bash printf due to compatibility issues
- Improved workflow in pool registration/modification for relays and multi-owner.

Fixed
- Error output for prerequisite checks


Scitz0:
Marking as do-not-merge for now as I have additional changes to implement. Ie comment in #611 regarding filename convention.

rdlrt (approved):

PR #618, opened 2020-12-04
Mac os prereqs improvements

hodlonaut:


Scitz0:
A couple of questions:

Does all recent OSX versions have the uname Darwin?

CNTools
As check for systemd-ask-password and chattr is ignored for Darwin, won't this create issues in encrypt/decrypt/backup?
chattr can be disabled in cntools.config. Better to force config setting for this to off for Darwin instead of ignoring check.
Ie `[[ -z ${ENABLE_CHATTR} ]] && ENABLE_CHATTR=true` in library. 
The row before, adding: `[[ $(uname) == Darwin ]] && ENABLE_CHATTR=false`
For systemd-ask-password I don't have an answer right now but if you have a suggestion that is both OSX and Linux compatible I can look into this.

gLiveView
`netstat` was used in early versions of gLiveView, could dig through the history to find the command used and update it and use it for OSX instead of `ss`

Scitz0:
All above-mentioned comments should now be handled including additional improvements in other parts of the code.

Scitz0 (approved):
👌

rdlrt (approved):

PR #623, opened 2020-12-06
Update prereqs.sh - CNCLI integration (latest tag instead of master, automake for libsodium)

Straightpool:
fixed cncli integration:
- automake needed for libsodium build
- only the latest tag should be build not the latest master (otherwise alpha builds of cncli are build)


rdlrt (approved):

rdlrt (changes requested):
Sorry, Need to merge against alpha branch to avoid conflicts, Thanks!!

Straightpool:
Will redo on alpha branch

PR #624, opened 2020-12-06
Update cncli.sh - Added TimeZone

Straightpool:
added timezone for leaderlog


Scitz0 (commented):
BLOCKLOG_TZ is set in env file. Ie we save all dates to DB in UTC and convert to TZ set at runtime when displaying the value instead. So you can set whatever timezone you want in env file and CNTools will convert it for you. Makes it easy to change TZ at a later point and still get the correct value.

Straightpool:
Excellent, even better solution, missed that parameter in the ENV file.

PR #625, opened 2020-12-06
reduce-runtime-pool-show

TrevorBenson:
new ledger_pstate variable


Scitz0 (approved):
👍 LGTM
If you want you can use bash herestrings instead of piping. Reduces the processes used and performance impact, though most likely very small. 
ie, instead of `echo ${something} | process input` you can use `process input <<< ${something}`

TrevorBenson:
I knew the pipe caused an extra shell which is inneficient, but I always forget the name of "herestrings" to verify usage. Thanks!

TrevorBenson:
Performance tests: https://gist.github.com/TrevorBenson/47b6c76160c3421132d4b4f8457ed664

PR #626, opened 2020-12-06
Update prereqs.sh - CNCLI integration (latest tag instead of master)

Straightpool:
+ Automake for libsodium and sqlite3 for blockdb

fixed cncli integration:

automake needed for libsodium build
sqlite3 needed for blockdb
only the latest tag should be build not the latest master (otherwise alpha builds of cncli are build)


Scitz0 (approved):
👌

PR #641, opened 2020-12-10
gLiveView: prevent null for grep for peers IN

rdlrt:


Scitz0 (approved):
Testing both empty result, one cncli connection and multiple cncli connections all seem to work.

PR #642, opened 2020-12-10
Update cncli.sh - Sync should connect to remote BP

Straightpool:
sync should connect to the remote bp if the bp is remote

### DIFF
--- a/scripts/cnode-helper-scripts/cncli.sh
+++ b/scripts/cnode-helper-scripts/cncli.sh
@@ -276,7 +276,7 @@ cncliInit() {
 #################################
 
 cncliSync() {
-  ${CNCLI} sync --host 127.0.0.1 --network-magic "${NWMAGIC}" --port "${CNODE_PORT}" --db "${CNCLI_DB}"
+  ${CNCLI} sync --host "${PT_HOST}" --network-magic "${NWMAGIC}" --port "${PT_PORT}" --db "${CNCLI_DB}"
 }
 
 #################################


rdlrt (approved):

Scitz0:
I'm not sure I agree with the change actually. The reason that remote connection was added for pooltool sendtip is to get as accurate data as possible when sending block tip to pooltool so that you would be sure pooltool can pick up all your bp's blocks, even if orphaned and discarded by your own relays. This is an extra connection and separate from cncli DB created by cncli sync. 

There is really no need to run cncli sync remotely. The cncli DB is used for pooltool sendslots, but this is just a leaderlog calculation and runs just as well on relay as on bp. All the leaderlog processing and keys needed for this as well as the cncli db itself reside locally on the relay anyway. The only thing cncli sync against the remote bp does is to add an extra delay to the sync. 

Or have I missed something in my reasoning above, could very well have :) ?

Straightpool:
Why is this important? 

See https://github.com/AndrewWestberg/cncli

**Sendtip section:**
The sendtip command is used to communicate with pooltool.io so you can have a green badge on their website with your current tip height. pooltool tip image

**It is important to point this command at your core nodes.** This will help pooltool capture any orphan blocks. There is no guarantee that an orphan block you make will be seen by pooltool. Pointing to your core nodes should help with that.

You cannot point sendtip to the core node directly, you need to point sync to the core node.

I confirmed this on the JorManagr Telegram group with Adam Dean.

Scitz0:
Yes and I agree with that I also mentioned in my reply

> The reason that remote connection was added for pooltool sendtip is to get as accurate data as possible when sending block tip to pooltool so that you would be sure pooltool can pick up all your bp's blocks, even if orphaned and discarded by your own relays.

But cncli sendtip has nothing to do the with sync process. ptsendtip and sync are two separate processes both syncing against the node. Sync saves all blocks into the cncli.db sqlite db. sendtip doesn't care about the sync or the cncli.db, it has it's own active connection to the node to fetch the latest block.

cncli sync, leaderlog, validate, sendslots all work together and query the common cncli.db that sync keeps up-to-date. 
cncli sendtip however does its own thing separate from those. It's only sendtip that should connect to bp, the rest can run wherever you want, on bp itself or on a relay syncing against the relay node.

Straightpool:
> But cncli sendtip has nothing to do the with sync process. ptsendtip and sync are two separate processes both syncing against the node. Sync saves all blocks into the cncli.db sqlite db. sendtip doesn't care about the sync or the cncli.db, it has it's own active connection to the node to fetch the latest block.
> 
> cncli sync, leaderlog, validate, sendslots all work together and query the common cncli.db that sync keeps up-to-date.
> cncli sendtip however does its own thing separate from those. It's only sendtip that should connect to bp, the rest can run wherever you want, on bp itself or on a relay syncing against the relay node.

Excellent, I didn't see that. I just saw that ptsendtip does use the host defined here already. In this case feel free to undo my merge and ignore.

PR #648, opened 2020-12-11
Revert "Update cncli.sh - Sync should connect to remote BP"

Straightpool:
Reverts cardano-community/guild-operators#642

Not needed as discussed for sync as ptsendstip only needs this remote connection and already uses the variables PT_HOST and PT_PORT


Scitz0:
Uneeded as master was reset.

PR #653, opened 2020-12-17
Add default CNODE_PORT

rdlrt:
We didnt have a default for this variable if user does not have it in user variables section (upgrading from old version)


hodlonaut (approved):
Looks good

PR #656, opened 2020-12-19
docker build

redoracle:
Hey boyz... 

here some major updates across all the files...

- more ENV variables can be used
- health-check implemented so Docker will know if Cardano node is having basic behaviors (Flatlines and cardano-ping as conditions)
- workflows updated with Priyank suggestions


files/docker/node/addons/entrypoint.sh
```diff
@@ -1,37 +1,35 @@
...
 
-if [[ $dbsize < $tnsizedb ]] && [[ $NETWORK == "testnet" ]] ; then
-cp -rf $CNODE_HOME/priv/testnet-db ${CNODE_HOME}/db \ 2>/dev/null
-fi
-
-if [[ "${POOL_NAME}" ]] ; then 
-export POOL_DIR="$CNODE_HOME/priv/pool/$POOL_NAME"
-echo "POOL_DIR set to: $POOL_DIR" ;
+if [[ "$dbsize" -lt "$bksizedb" ]]; then
+cp -rf $CNODE_HOME/priv/$NETWORK-db/* ${CNODE_HOME}/db 2>/dev/null
```
TrevorBenson, line 26: We found that data inside the `${CNODE_HOME}/db` can cause cardano-node to fail to start. @Scitz0 I believe relocated block db/logs to `${CNODE_HOME}/guild-db/blocklog`.

TrevorBenson: I think I misunderstood this part initially. Looking it over it appears to be a snapshot deployment, not the blocklogs. So everything LGTM.

TrevorBenson (approved):
LGTM

PR #658, opened 2020-12-20
fix wallet import failing because of missing missing argument to cardano-address key public

igelineau:
When I use the Wallet Import option with the latest version of cardano-address (3.0.0), cardano-wallet (2020.12.8), I get an error:

```
Usage: cardano-address key public (--without-chain-code | --with-chain-code)
  Get the public counterpart of a private key

Available options:
  -h,--help                Show this help text

The private key is read from stdin.To get extended public key pass '--with-chain-code'. To get public key pass '--without-chain-code'.
user error (Bech32 error: string is too short)
user error (Bech32 error: string is too short)
Error: Wrong input size of 0
bech32: user error (StringToDecodeTooShort)
Shelley command failed: key verification-key  Error: /opt/cardano/cnode/priv/wallet/adwdada/payment.skey: Invalid key.
Shelley command failed: key non-extended-key  Error: /tmp/cntools/payment.evkey: /tmp/cntools/payment.evkey: openBinaryFile: does not exist (No such file or directory)
```

Adding the `--with-chain-code` option solves the issue and imports the wallet correctly.

### DIFF
--- a/scripts/cnode-helper-scripts/cntools.sh
+++ b/scripts/cnode-helper-scripts/cntools.sh
@@ -410,8 +410,8 @@ case $OPERATION in
     payment_xprv=$(cardano-address key child 1852H/1815H/0H/0/0 <<< ${root_prv})
     stake_xprv=$(cardano-address key child 1852H/1815H/0H/2/0 <<< ${root_prv})
     
-    payment_xpub=$(cardano-address key public <<< ${payment_xprv})
-    stake_xpub=$(cardano-address key public <<< ${stake_xprv})
+    payment_xpub=$(cardano-address key public --with-chain-code <<< ${payment_xprv})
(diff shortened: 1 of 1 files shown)


rdlrt:
Thanks a lot for the PR.

Guild doco uses cabal-build-all which currently downloads cardano-addresses 2.1.0 (not 3.0). This is because there are some features that have been disabled in later releases, causing bit of a trouble to drill down to a specific derivation, as the plan is to see if we can import HD Wallet operations too. Once we have an alternative, we can proceed to support higher cardano-address binaries, but until then - we might have to delay the integration

igelineau:
oh I see, thanks for the explanations.
I'll keep it patched on my side you guys figure it out then !

rdlrt:
Added via #922 while maintaining compatibility with older version

PR #660, opened 2020-12-20
docker build update

redoracle:
- removed 1 workflow 
- removed duplicate lines in docker stage3
- added manual dispatch to node-version workflow and local change to trigger the docker build


redoracle:
Trying to solve conflicts

PR #670, opened 2020-12-22
add a variable for cncliLeaderlog.

btbf:
add a variable for cncliLeaderlog.
Custom json-path when CNTOOL & TIMEZONE

I would like to add a variable so that it can be used when it is introduced outside of CNTOOL. Also, I want to set the time zone in cncliLeaderlog.


Scitz0 (commented):
cncli source env file for common variables, genesis files are taken from here.
If you use a custom setup you can just set CONFIG in env to your config file and it should pick up the genesis files automatically from config file, no need to set them separately in cncli.sh.

The timezone in database should always be UTC to keep it consistent over time even if the user decide to show date in a different timezone. It's better to adapt timezone at display than in blocklog db. env file has a setting for this as well called BLOCKLOG_TZ.
In CNTools I use this like this to modify the date:
`at=$(TZ="${BLOCKLOG_TZ}" date '+%F %T %Z' --date="${at}")`
at is the field in blocklog db for the specific block.

Please let me know if you need any further assistance to set it up for your specific needs, most should be possible as it is without further code changes.

Scitz0:
Closing PR due to the previous comment.

PR #674, opened 2020-12-24
Cntools 6.4.0

Scitz0:
As alpha was just merged to master I'm now ready to merge this non-production ready CNTools update to alpha. 
Should be in a pretty stable state. This is a major update to the core code of CNTools. Might be worth a major version bump but let's postpone that for later.

- Logging completely re-designed from the ground. Previous selective logging wasn't very useful. All output(almost) now outputted to both the screen and to timestamped log file. One log file is created per CNTools session. Old log file archived in 'archive' subfolder and last 10 log files kept, rest is pruned on CNTools startup.

When println function is used the following logging options are available:
  ```
OFF      : logging disabled, output only to tty
LOG      : logged as DEBUG but not printed to tty
DEBUG    : verbose output, logged and printed to tty
INFO     : normal output printed to tty and logged
ACTION   : e.g cardano-cli executions etc, logged but not printed to tty
ERROR    : stderr output and error messages
```
When println function is not used, ie when using file descriptors
```
0 : stdin  - unchanged
1 : stdout - prints to tty and log as INFO
2 : stderr - prints to tty and log as ERROR
3 : custom - prints to tty and log as DEBUG
6 : custom - normal stdout, no logging
7 : custom - normal stderr, no logging
8 : custom - same as (1) but useful if you want to restore normal stdout for a block of code and then revert to custom stdout. Check out waitForInput() function for example usage.
9 : custom - same as (2), see info for (8).
```

- Vacuumlabs cardano-hw-cli added to prereqs.sh, option '-w' incl Ledger udev rules. Initial implementation not tested. Depend on software update(1.2.0) from Vacuumlabs not yet released to be fully functional. 
- Because of HW wallet support transaction signing has been re-designed to make the workflow consistent weather using hw wallet or not. Raw tx first witnessed by all signing keys separately and then assembled and signed instead of signing directly with all signing keys. 
- Option added to disable Dialog for file/dir input

### CODE REVIEW

File: scripts/cnode-helper-scripts/cntools.library [hunk 1]
```diff
@@ -530,52 +531,54 @@ function selectPool {
...
-  [[ ${enc_req_files} -gt 0 ]] && say "${FG_YELLOW}encrypted pools found but NOT listed, please decrypt to show${NC}"
+  [[ ${enc_req_files} -gt 0 ]] && println "DEBUG" "${FG_YELLOW}encrypted pools found but NOT listed, please decrypt to show${NC}"
   if ! selectDir "pool" "${pool_dirs[@]}"; then return 1; fi # ${dir_name} populated by selectDir function
   pool_name="${dir_name}"
 }
 
 
-# Command    : getPassword [optional:confirm]
-# Description: Get password from user on STDIN
-# Parameters : confirm   >   Optional parameter, force user provide password twise to confirm
```
TrevorBenson, line 545: `# Parameters : confirm   >   Optional parameter, force user provide password twice to confirm`

scripts/cnode-helper-scripts/cntools.library
```diff
@@ -530,52 +531,54 @@ function selectPool {
...
-# Return     : populates $password variable, make sure to unset variable when done
-# Examples of Usage:
-#   >> local userPassword=$(getPassword confirm)
+# Command     : getPassword [optional: confirm]
+# Description : Get password from user on STDIN
+# Parameters  : confirm  >  [optional] force user provide password twise to confirm
```
TrevorBenson, line 546: `# Parameters  : confirm  >  [optional] force user provide password twice to confirm`

TrevorBenson (approved):
Didn't scrutinize every line, but put eyes on each. Only found a single typo and nothing else jumped out at me that can't be tested inside the alpha branch.

LGTM

gufmar (commented):
LGTM

PR #679, opened 2020-12-28
telegram alerts

gufmar:
first version of telegramSend function, defined in env. Guide to setup the own bot and find the chat id

### DIFF
--- a/docs/Scripts/logmonitor.md
+++ b/docs/Scripts/logmonitor.md
@@ -15,30 +15,3 @@ Log output is handled by syslog and end up in the systems standard syslog file, 
 
 ##### View Blocklog
 Best viewed in CNTools or gLiveView. See [CNCLI](Scripts/cncli.md) for example output.
-
-#### Telegram alerts
-
-If known but unwanted errors occur on your node, or if characteristic values indicate an unusual status , cntools can send you Telegram alert messages. 
-
-To do this, you first have to activate your own bot and link it to your own Telegram user. Here is an explanation of how this works:
-
-1. Open Telegram and search for "*botfather*"
-
-2. write him your wish: `/newbot`
-
-3. define a name for your bot, such as `cntools_[POOLNAME]_alerts`
-
-4. Botfather will confirm the creation of your bot by giving you the unique **bot access token**. Keep it save and private.
-
-5. Now send at least one direct message to your new bot.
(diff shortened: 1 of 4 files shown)


docs/Scripts/logmonitor.md
```diff
@@ -14,4 +14,31 @@ Use the `deploy-as-systemd.sh` script to create a systemd unit file (deployed to
...
+
+To do this, you first have to activate your own bot and link it to your own Telegram user. Here is an explanation of how this works:
+
+1. Open Telegram and search for "*botfather*"
+
+2. write him your wish: `/newbot`
+
+3. define a name for your bot, such as `cntools_[POOLNAME]_alerts`
+
+4. Botfather will confirm the creation of your bot by giving you the unique **bot access token**. Keep it save and private.
```
Scitz0, line 31: LGTM, just a small typo in "Keep it **save** and private."

I can see why you placed it in logmonitor md as this is one use case for the TG alerts. But it could just as well be used in other scripts so maybe it deserves its own section in the script section of the docs?

Scitz0 (approved):

jackkitley:
I feel this needs more detail. 
1. You can give it any name? Doesnt need to be in this format? cntools_[POOLNAME]_alerts
2. After you get your token and chat id you need to add this to the env file in two places. I dont feel this is clear.
3. "Keep it save and private." = "Keep it safe and private."

Scitz0:
We haven't really decided what to do with this implementation yet. It's not currently used in any of the scripts.

gufmar:
> 1. You can give it any name? Doesnt need to be in this format? cntools_[POOLNAME]_alerts

yes, feel free to give it any still free name. the "such as" example just aims to be self-explanatory for you and the future. 

> 2. After you get your token and chat id you need to add this to the env file in two places. I dont feel this is clear.

it basically means that you have to add the token and the chat ID (2 values) to the env files. This must be done in the same way as you set there your pool name or a custom ECG port for example. 

> 3. "Keep it save and private." = "Keep it safe and private."

thanks, fix will be merged soon.
As @Scitz0 mentioned this was a preparation only so far. 
The idea was to actively use it for certain (time) critical events/alerts. For example if blocklog detects invalid / missed / ghosted slots/blocks.

PR #686, opened 2021-01-05
CNTools 7.0.0

Scitz0:
Changelog below include previous commits from #674
Commit also include a new env doc page that explains the purpose of the common env file as well as a general update of all the different script docs.
External scripts that rely on cntools.library updated.

-----------------------
## [7.0.0] - 2021-01-XX
Though mostly unchanged in the user interface, this is a mayor update with most of the code re-written/touched in the back-end.
Only the most noticeable changes added to changelog.

##### Added
- HW Wallet support through Vacuumlabs cardano-hw-cli (Ledger Nano X/S & Trezor T)
  - Vacuumlabs cardano-hw-cli added as build option to prereqs.sh, option '-w' incl Ledger udev rules. Software from Vacuumlabs and Ledger app still early in development and may contain limitations that require workarounds.
  - Because of HW wallet support, transaction signing has been re-designed. For CLI and HW wallet pool reg, raw tx is first witnessed by all signing keys separately and then assembled and signed instead of signing directly with all signing keys. But for all other HW wallet transactions, signing is done directly without first witnessing.
  - Requires updated Cardano app in Ledger/Trezor set to be released in January 2021 to use in pool registration/modification.
- Option added to disable Dialog for file/dir input in cntools.config

##### Changed
- Logging completely re-designed from the ground. Previous selective logging wasn't very useful. All output(almost) now outputted to both the screen and to a timestamped log file. One log file is created per CNTools session. Old log file archived in logs/archive subfolder and last 10 log files kept, rest is pruned on CNTools startup.
  - DEBUG    : Verbose output, most output printed on screen is logged as debug messages except explicitly disabled, like menu printing.
  - INFO     : Informational and the most important output.
  - ACTION   : e.g cardano-cli executions etc
  - ERROR    : error messages and stderr output
- Verbosity setting in cntools.config removed.
- Offline workflow now use a single JSON transaction file holding all data needed. This allows us to bake in additional data in the JSON file in addition to the tx body to make it much more clear what the offline transaction do. Like signing key verification, transaction data like fee, source wallet, pool name etc. It also lets the user know on offline computer what signing keys is needed to sign the transaction.
  - Sign Tx moved to Transaction >> Sign
  - Submit Tx moved to Transaction >> Submit

### DIFF
--- a/docs/Scripts/cntools-changelog.md
+++ b/docs/Scripts/cntools-changelog.md
@@ -5,13 +5,13 @@ All notable changes to this tool will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [7.0.0] - 2021-01-XX
-Though mostly unchanged in the user interface, this is a mayor update with most of the code re-written/touched in the back-end.  
+## [7.0.0] - 2021-01-08
+Though mostly unchanged in the user interface, this is a major update with most of the code re-written/touched in the back-end.  
 Only the most noticeable changes added to changelog. 
 
 ##### Added
 - HW Wallet support through Vacuumlabs cardano-hw-cli (Ledger Nano X/S & Trezor T)
-  - Vacuumlabs cardano-hw-cli added as build option to prereqs.sh, option '-w' incl Ledger udev rules. Software from Vacuumlabs and Ledger app still early in development and may contain limitations that require workarounds.
+  - Vacuumlabs cardano-hw-cli added as build option to prereqs.sh, option '-w' incl Ledger udev rules. Software from Vacuumlabs and Ledger app still early in development and may contain limitations that require workarounds. Users are recommended to familiarise their usage using test wallets first.
   - Because of HW wallet support, transaction signing has been re-designed. For CLI and HW wallet pool reg, raw tx is first witnessed by all signing keys separately and then assembled and signed instead of signing directly with all signing keys. But for all other HW wallet transactions, signing is done directly without first witnessing.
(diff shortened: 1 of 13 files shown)


docs/Scripts/env.md
```diff
@@ -0,0 +1,71 @@
...
+# User Variables - Change as desired #
+# Leave as is if unsure              #
+######################################
+
+#CCLI="${HOME}/.cabal/bin/cardano-cli"                  # Override automatic detection of path to cardano-cli executable
+#CNCLI="${HOME}/.cargo/bin/cncli"                       # Override automatic detection of path to cncli executable (https://github.com/AndrewWestberg/cncli)
+#CNODE_HOME="/opt/cardano/cnode"                        # Override default CNODE_HOME path (defaults to /opt/cardano/cnode)
+CNODE_PORT=6000                                         # Set node port
+#CONFIG="${CNODE_HOME}/files/config.json"               # Override automatic detection of node config path
+#SOCKET="${CNODE_HOME}/sockets/node0.socket"            # Override automatic detection of path to socket
```
rdlrt, line 26: Should we perhaps add a `...` after this line to not worry about any mismatch?

docs/Scripts/cntools-changelog.md
```diff
@@ -5,6 +5,29 @@ All notable changes to this tool will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [7.0.0] - 2021-01-XX
+Though mostly unchanged in the user interface, this is a mayor update with most of the code re-written/touched in the back-end.
```
rdlrt, line 9: Typo: mayor -> major

rdlrt (approved):
LGTM

PR #706, opened 2021-01-13
package-mismatch-os-version

TrevorBenson:
Should fix the RHEL/CentOS & Fedora package name for libusb. Also set it to the default packages instead of the development libraries which is not required to compile cardano-hw-cli from my experience.


TrevorBenson:
Ignore the first commit, IDE kept the prior commit message, this has nothing to do w/ DBSYNC.

cardano-community (approved):
LGTM

PR #707, opened 2021-01-13
Update process check to gather params to use port instead of CNODE_HOME

rdlrt:
Since CNODE_HOME is not available to those not using CNODE_HOME, start using CNODE_PORT in `ps -ef` query


hodlonaut (approved):
Looks good

PR #712, opened 2021-01-13
Fix rustup not in path when called.

TrevorBenson:
Source the cargo env


Scitz0 (approved):
LGTM

scripts/cnode-helper-scripts/prereqs.sh
```diff
@@ -302,6 +302,7 @@ if [[ "${INSTALL_CNCLI}" = "Y" ]]; then
     if ! command -v "rustup" &>/dev/null; then
       echo "  installing RUST..."
       if ! output=$(curl --proto '=https' --tlsv1.2 https://sh.rustup.rs -sSf | sh -s -- -y 2>&1); then echo -e "${output}" && err_exit; fi
+      source $HOME/.cargo/env
```
rdlrt, line 305: Isnt this already set by rustup install command?

Scitz0: It is added to users profile file yes by the script, but not available inside prereqs directly after. So either sourcing .profile or $HOME/.cargo/env like in this PR is needed. Else the following two lines of code will fail.

PR #713, opened 2021-01-13
Replace local privilege wget with escalated privilege curl

TrevorBenson:
Wget was used to download one file, but was not installed in the earlier software prerequisites installation.


Scitz0 (approved):
Correct, curl should be used exclusively in our scripts. Copy&Paste from Ledger docs and missed.
Though moved sudo back to bash call for Ledger udev rules script as it's here its needed.

PR #723, opened 2021-01-16
typo in cncli.sh and topologyUpdater.sh update

Scitz0:
priority fix, typo that mess up cncli.sh and topologyUpdater.sh on update 🤦‍♂️


rdlrt (approved):

TrevorBenson (approved):
Thanks! I would have opened a ticket Monday but I haven't had enough time to confirm that it wasn't my own problem within multi-stage container builds causing an issue within the templating.

PR #724, opened 2021-01-17
Guild LiveView 1.18.0

Scitz0:
Performance enhanced version by reducing the number of external calls and processing in each loop. By default, it now tries to use Prometheus metrics if found in the config, else EKG. Incl other enhancements, Prometheus and bash native parsing is around 10x as fast as the previous version. The new EKG + jq parsing is around 5 times as fast as the previous version.

Because of this, the default refresh rate was reduced from 5s to 2s. It even runs perfectly fine w/o any major performance penalty with a refresh of 1s on my test systems.

### DIFF
--- a/scripts/cnode-helper-scripts/env
+++ b/scripts/cnode-helper-scripts/env
@@ -221,21 +221,6 @@ else
   NETWORK_IDENTIFIER="--testnet-magic ${NWMAGIC}"
   [[ "${NWMAGIC}" = "1097911063" ]] && NETWORK_NAME="Testnet" || NETWORK_NAME="Custom"
 fi
-getEraIdentifier() {
-    if ${CCLI} query protocol-parameters --byron-era ${PROTOCOL_IDENTIFIER} ${NETWORK_IDENTIFIER} &>/dev/null; then ERA_IDENTIFIER="--byron-era" && ERA_WITNESS="TxWitness ByronEra"
-  elif ${CCLI} query protocol-parameters --shelley-era ${PROTOCOL_IDENTIFIER} ${NETWORK_IDENTIFIER} &>/dev/null; then ERA_IDENTIFIER="--shelley-era" && ERA_WITNESS="TxWitness ShelleyEra"
-  elif ${CCLI} query protocol-parameters --allegra-era ${PROTOCOL_IDENTIFIER} ${NETWORK_IDENTIFIER} &>/dev/null; then ERA_IDENTIFIER="--allegra-era" && ERA_WITNESS="TxWitness AllegraEra"
-  elif ${CCLI} query protocol-parameters --mary-era ${PROTOCOL_IDENTIFIER} ${NETWORK_IDENTIFIER} &>/dev/null; then ERA_IDENTIFIER="--mary-era" && ERA_WITNESS="TxWitness MaryEra"
-  else 
-    ERA_IDENTIFIER=""; ERA_WITNESS=""
-    [[ ${OFFLINE_MODE} = "N" ]] && return 1
-  fi
-  return 0
-}
-if ! getEraIdentifier && [[ ${OFFLINE_MODE} = "N" ]]; then echo "Failed to query protocol-parameters from node, not yet fully started?" && return 1; fi
-
-PROT_PARAMS="$(${CCLI} query protocol-parameters ${ERA_IDENTIFIER} ${PROTOCOL_IDENTIFIER} ${NETWORK_IDENTIFIER} 2>&1)"
-if [[ -n "${PROT_PARAMS}" ]] && ! DECENTRALISATION=$(jq -er .decentralisationParam <<< ${PROT_PARAMS} 2>/dev/null); then DECENTRALISATION=0.5; fi
 
 SHELLEY_GENESIS_START_SEC=$(date --date="$(jq -r .systemStart "${GENESIS_JSON}")" +%s)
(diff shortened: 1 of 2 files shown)


TrevorBenson (approved):
LGTM

scripts/cnode-helper-scripts/env
```diff
@@ -297,4 +282,21 @@ STANDOUT='\e[7m'
...
+  elif ${CCLI} query protocol-parameters --allegra-era ${PROTOCOL_IDENTIFIER} ${NETWORK_IDENTIFIER} &>/dev/null; then ERA_IDENTIFIER="--allegra-era" && ERA_WITNESS="TxWitness AllegraEra"
+  elif ${CCLI} query protocol-parameters --mary-era ${PROTOCOL_IDENTIFIER} ${NETWORK_IDENTIFIER} &>/dev/null; then ERA_IDENTIFIER="--mary-era" && ERA_WITNESS="TxWitness MaryEra"
+  else 
+    ERA_IDENTIFIER=""; ERA_WITNESS=""
+    [[ ${OFFLINE_MODE} = "N" ]] && return 1
+  fi
+  return 0
+}
+# special exit code for scripts that want to source env and doesn't care about ERA_IDENTIFIER or ERA_WITNESS
+if ! getEraIdentifier && [[ ${OFFLINE_MODE} = "N" ]]; then echo "Failed to query protocol-parameters from node, not yet fully started?" && return 2; fi
```
rdlrt, line 297: Nothing big, but just noticed we have OFFLINE_MODE check already on line 292, and essentially getEraIdentifier will return 1 only if  ${OFFLINE_MODE} = "N".
Possibly below?
```suggestion
[[ ! getEraIdentifier ]] && echo "Failed to query protocol-parameters from node, not yet fully started?" && return 2
```

scripts/cnode-helper-scripts/gLiveView.sh
```diff
@@ -150,14 +149,13 @@ if [[ "${NO_INTERNET_MODE}" == "N" ]]; then
       fi
     else
       mv "${PARENT}"/env.tmp "${PARENT}"/env
-      echo -e "Common env file downloaded: ${PARENT}/env"
-      echo -e "This is a mandatory prerequisite, please set variables accordingly in User Variables section in the env file and restart Guild LiveView\n"
-      exit 0
+      myExit 0 "Common env file downloaded: ${PARENT}/env\nThis is a mandatory prerequisite, please set variables accordingly in User Variables section in the env file and restart Guild LiveView\n"
```
rdlrt, line 152: Minor wording since we already retain user variables
```suggestion
      myExit 0 "Common env file downloaded: ${PARENT}/env\nUser variables will be retained, but please perform a sanity check in User Variables section of env file and restart Guild LiveView\n"
```

scripts/cnode-helper-scripts/gLiveView.sh
```diff
@@ -521,21 +519,40 @@ checkPeers() {
...
+.cardano.node.ChainDB.metrics.slotNum.int.val //0,
+.cardano.node.Forge.metrics.remainingKESPeriods.int.val //0
+] | @tsv' <<< "${data}")
+  read -ra data_arr <<< ${data_tsv}
+  epochnum=${data_arr[0]}
+  slot_in_epoch=${data_arr[1]}
+  slotnum=${data_arr[2]}
+  remaining_kes_periods=${data_arr[3]}
+else
+  data=$(curl -s -m ${EKG_TIMEOUT} "http://${EKG_HOST}:${prom_port}/metrics" 2>/dev/null)
```
rdlrt, line 537: should we check for `.hasPrometheus[0]` instead of EKG_HOST? I bet some folks may have changed Prometheus server listen port for grafana but not added one for EKG

rdlrt (changes requested):
Looks good, multi value to tsv is nice! Just couple of minor suggestions

scripts/cnode-helper-scripts/env
Scitz0: Correct, as `[[ ${OFFLINE_MODE} = "N" ]]` is checked in `getEraIdentifier()` it will always return 0 in offline mode so no need to check this again. Though function call can not be used inside `[[ ]]` so I suggest it will be changed to this instead:
`if ! getEraIdentifier; then echo "Failed to query protocol-parameters from node, not yet fully started?" && return 2; fi`

scripts/cnode-helper-scripts/gLiveView.sh
Scitz0: This message is only output if this check fails: `if [[ -f "${PARENT}"/env ]]; then`
Meaning that no previous env file is found, ie gLiveView.sh script has been downloaded and executed without env being present in the same dir. On update with retained User Variables for env file, we output the message: `echo -e "\nUpdate to env file applied successfully!"`. Though actually pointless in the end if we don't add a waitForInput() type of code after as code will continue and clear screen without the user ever seeing the output.

scripts/cnode-helper-scripts/env
rdlrt: agreed :)

scripts/cnode-helper-scripts/gLiveView.sh
rdlrt: ah right, agree

rdlrt (approved):

PR #730, opened 2021-01-20
Changes for cardano-node and cardano-cli 1.25.0

rdlrt:
> Place to collect changes post upgrade, Not to be merged yet

## Changed
- gLV-1.19: Update metrics to exclude ChainDB and Forge
- CNTools-7.1.0: Update metrics to exclude ChainDB references
- Make corresponding changes in sLiveView.sh and cncli.sh
- Update config templates for metrics and tracers
  - Retire config parameters that are no longer used by node
  - Retire stale praos config from templates as well as prereqs script


Scitz0 (approved):
Looks good so far. I think we should add staging network to the list as well.

Another thing we need to look over and handle before Mary HF is how multi-assets affect all utxo queries in CNTools for balance check and wallet listing. I have had this on my ToDo for a while but other stuff has gotten the priority. This also includes sending tokens. Not sure it has to hinder this PR though. If we add staging network and merge this PR, then I can focus on the asset stuff in another PR.

PR #732, opened 2021-01-23
Add more checks for config

rdlrt:
- Tracer check for ChainDb
- Ensure old mapping backends have been deleted (prompting for upgrade)
- Add CONFIG to pgrep to grab PID
- If EKG or Prom are specified in config, but they're already in use by another node - we'd often get quite a confusing behaviour depending on network the other node is coming from. Add a check to ensure Prometheus port (if USE_EKG = 'N') is also being listened by CNODE_PID. Same for EKG_PORT if USE_EKG='Y'
- Move ss/lsof/jq check to env instead of gLiveView

### DIFF
--- a/scripts/cnode-helper-scripts/env
+++ b/scripts/cnode-helper-scripts/env
@@ -460,6 +460,12 @@ getCurrentKESperiod() {
   echo $(( tip_ref / SLOTS_PER_KES_PERIOD ))
 }
 
+# Description : Calculate expected interval between blocks
+slotInterval() {
+  if [[ -z ${DECENTRALISATION} || $(echo "${DECENTRALISATION} < 0.5" | bc) -eq 1 ]]; then d=0.5; else d=${DECENTRALISATION}; fi
+  echo "(${SLOT_LENGTH} / ${ACTIVE_SLOTS_COEFF} / ${d}) + 0.5" | bc -l | awk '{printf "%.0f\n", $1}'
+}
+
 # Description : Identify current era and set variables accordingly
 getEraIdentifier() {
     if ${CCLI} query protocol-parameters --byron-era ${NETWORK_IDENTIFIER} &>/dev/null; then ERA_IDENTIFIER="--byron-era" && ERA_WITNESS="TxWitness ByronEra"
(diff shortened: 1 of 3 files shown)


scripts/cnode-helper-scripts/env
```diff
@@ -144,6 +144,12 @@ fi
 if ! jq -r . "${CONFIG}" >/dev/null 2>&1; then
   echo "Could not parse ${CONFIG} file in JSON format, please double-check the syntax of your config, or simply download it from guild-operators repository!"
   return 1
+elif [[ "$(jq -r .TraceChainDb ${CONFIG})" != "true" ]]; then
+  echo "The ${CONFIG} file suggests that you have set TraceChainDb parameter to false. Please change it to true, as it is required to extract statistics from node."
+  return 1
+elif [[ "$(jq .options.mapBackends.cardano.node.BlockFetchDecision.peers /opt/cardano/cnode/files/config.json 2>&1)" != "null" ]]; then
```
Scitz0, line 150: This will return null for all configs. `cardano.node.BlockFetchDecision.peers` have to be quoted. The jq command can also utilize the -e flag available to check exit code for if a specific query return ok result or not. Config path is also hardcoded.

So something like this could replace the if check:
```
elif jq -e .options.mapBackends.\"cardano.node.BlockFetchDecision.peers\" "${CONFIG}" &>/dev/null; then
```

Scitz0 (changes requested):
LGTM when the comment has been addressed.

scripts/cnode-helper-scripts/env
```diff
@@ -234,7 +242,17 @@ BLOCKLOG_DB="${BLOCKLOG_DIR}/blocklog.db"
...
+
+if [[ -n "${CNODE_PID}" ]]; then
+  if [[ "${USE_EKG}" == "N" ]] && [[ "$(ss -lnpt | grep "${CNODE_PID}," | awk '{print $4}' | cut -d: -f2)" != **"${PROM_PORT}"** ]]; then
+    echo "ERROR: You specified ${PROM_PORT} as your Prometheus port, but it looks like the cardano-node (PID: ${CNODE_PORT} ) is not listening on this port. Please update the config or kill the conflicting process first."
+    exit 1
+  elif [[ "$(ss -lnpt | grep "${CNODE_PID}," | awk '{print $4}' | cut -d: -f2)" != **"${EKG_PORT}"** ]]; then
+    echo "ERROR: You specified ${EKG_PORT} as your EKG port, but it looks like the cardano-node (PID: ${CNODE_PID} ) is not listening on this port. Please update the config or kill the conflicting process first."
+    exit 1
+  fi
+fi
```
Scitz0, line 255: The only issue I see with this is the ss utilization. In gLiveView we have added a check for the existence of this command as well as lsof as fallback command for this.
E.g
```
if command -v "ss" &>/dev/null; then 
  use_lsof='N'
elif command -v "lsof" &>/dev/null; then 
  use_lsof='Y'
else
  myExit 1 "'ss' and fallback 'lsof' commands missing, please install using latest prereqs.sh script or with your packet manager of choice.\nhttps://command-not-found.com/ss can be used to check package name to install."
fi
```
And then in code we run it like this:
```
if [[ ${use_lsof} = 'Y' ]]; then
      peers=$(lsof -Pnl +M -i4 | grep ESTABLISHED | awk -v pid="${CNODE_PID}" -v port=":(${CNODE_PORT}|${EKG_PORT}|${PROM_PORT})->" '$2 == pid && $9 !~ port {print $9}' | awk -F "->" '{print $2}')
    else
      peers=$(ss -tnp state established 2>/dev/null | grep "${CNODE_PID}," | awk -v port=":(${CNODE_PORT}|${EKG_PORT}|${PROM_PORT})" '$3 !~ port {print $4}')
    fi
```
Is this something we should take into account in env as well?

scripts/cnode-helper-scripts/env
rdlrt: Agreed, thanks

scripts/cnode-helper-scripts/env
rdlrt: Hmm I think its prolly good to move those checks (alongwith check for `jq` to env as well - as it will be used by each tool script except cnode.sh)

Scitz0 (approved):

PR #770, opened 2021-02-09
CNTools 7.2.0 - Multi-Assets

Scitz0:
Adds multi asset token compatibility to CNTools.

- New *Advanced* section added
  - Hidden by default if not activated by runtime flag `-a` or in cntools.config.
  - Metadata transactions moved to this new section.
  - Asset management like policy creation, asset minting & burning added in this new section.
- All operations updated to support multi-assets
- `Wallet >> Show` completely redesigned.
- `Funds >> Send` now support sending both Ada and custom assets. It's possible to send multiple assets in a single transaction to destination wallet/address.

ToDo:
- [x] Fix sendADA.sh script
- [x] Add encryption operations to policies
- [x] Add asset folder to CNTools backup
- more?


rdlrt (approved):
Looks good, great work and completion!!

A minor comment about handling output in wallet > show made offline to consider if we can display it better, but it can also be handled in a seperate PR after merging this one

Issue #747, opened 2021-01-30
gLiveView peers unreachable on centos 8 due to tcptraceroute version


brainrecursion:
**Describe the bug**
The version of tcptraceroute which gets installed on centos 8 does not work with gLiveView.sh. 
The command has to be run as root due to permissions. Also the output does not include [open] or [closed] which gLiveView relies on.

**To Reproduce**
1. Install on centos 8 without cncli
2. Test gLiveView.sh 
3. All out peers are shown as unreachable.
4. Run tcptraceroute command outside gLiveView to verify

**Expected behavior**

**Version:**
 - OS: CentOS 8
 - Product version: gLiveView 1.19.2
 - Cardano Node version: 1.25.1
 - Network you're connecting to: testnet

**Additional context**

rdlrt:
I use gLiveView on CentOS 8 myself. tcptraceroute does *NOT* need to be run as root. If for some reason you have invalid permissions, it is likely an issue at your OS level - or the process of deploying tcptraceroute on your image.

brainrecursion:
This is running on standard centos 8 image or red hat image on Azure.

rdlrt:
@brainrecursion - We have now added `ss` as option before checking for tcptraceroute - as `ss` should be part of most linux distributions and is already used for fetching peers list

PR #772, opened 2021-02-10
Fix curl command usage across repo;Use ss for rtt before tcptraceroute; closes #754 and #747

### LINKED ISSUE #754: gLiveView.sh / tcptraceroute using bug
tcptraceroute can't using without root privileges, therefore in gLiveView script this util always return "socket: Operation not permitted" and all peers always marks as unreachable.

rdlrt:
- Remove compulsion for tcptraceroute, as precedence will be cncli -> ss -> tcptraceroute -> ping
- Add `ss` for RTT check, and update info/usage help texts to reflect the same
- Fix curl command usage across repo to fail with exit code for cases where:
  - output was sent to contents of file
  - curl command was in a condition


Scitz0 (approved):
LGTM

PR #774, opened 2021-02-11
update setup_mon.sh to show -i instead of -h

chadle-git:
small change for the Usage, it was showing a -h for hostname and below it shows a -i when running the script.


rdlrt (approved):
Thanks

PR #794, opened 2021-02-22
Minor Identation fixes

rdlrt:
- [x] gLV - Version not bumped as code is unaffected (just identation)
- [x] topologyUpdater - also, added a change to report IPv4 address when using curl
- [x] common env
- [x] cncli.sh
- [x] cntools.library
- [x] cntools.sh
- [x] Topology Updater to use custom peers outside of API (so that those IPs are not shared)
- [x] Update cncli to 1.4.0 and use API instead of ledger-dump


Scitz0 (approved):
LGTM so far

Scitz0 (approved):
Going on trusted source for this one 😄

gufmar (approved):
Lgtm

PR #811, opened 2021-02-25
Allow user to run gLiveView.sh from any directory

eddietejeda:
The current implementation requires that the user run gLiveView.sh from the `$CNODE_HOME/scripts` directory. Meaning that I can't call it directly from whatever directory I am in. For example, by just typing full path, like this: 

`~/> $CNODE_HOME/scripts/gLiveView.sh`

or by creating a symlink in my home directory called `~/scripts/` that points to the `$CNODE_HOME/scripts` that I can call like this:

`/> ~/scripts/gLiveView.sh`

This patch checks to see if `$CNODE_HOME` is defined and that `$CNODE_HOME/scripts` is a directory. If so, `PARENT` is set to that directory. Otherwise, it reverts to the default implementation.

### DIFF
--- a/scripts/cnode-helper-scripts/gLiveView.sh
+++ b/scripts/cnode-helper-scripts/gLiveView.sh
@@ -53,7 +53,7 @@ setTheme() {
 
 GLV_VERSION=v1.19.4
 
-PARENT="$(dirname $0)"
+PARENT=$( [ -d "$CNODE_HOME/scripts" ] && echo "$CNODE_HOME/scripts" || echo "dirname $0" )
 [[ -f "${PARENT}"/.env_branch ]] && BRANCH="$(cat ${PARENT}/.env_branch)" || BRANCH="master"
 
 # Set default for user variables added in recent versions (for those who may not necessarily have it due to upgrade)


rdlrt:
Not sure if I understand at all what you're trying to get at..
1. Its NOT true that you need to run gLiveView in $CNODE_HOME/scripts structure.
2. You're adding a change that already had `dirname $0` with condition to *ADD* a seperate case for `$CNODE_HOME/scripts` and if not come to same value...so effectively you're not really changing anything for case in question.

TrevorBenson:
@eddietejeda I suggest you just create yourself an alias:

`alias gLiveView.sh='/opt/cardano/cnode/scripts/gLiveView.sh'`

1. Add that line to your `~/.bashrc` for the user who owns all the guild ops files (as long as you use cnode as your directory).
2. Then either logout and back in login or `source ~/.bashrc` to activate the alias.
3. Now you can call gLiveView.sh from any directory and the original `$(dirname $0)` will work as intended to find the env.

You can also tab complete the alias similar to tab completing executable files in your path. I believe this should cover the various issues you have run into and described.

eddietejeda:
Thanks for the suggestions! We can close this issue.

PR #817, opened 2021-02-26
Increase cncli ptsendslots wait time after epoch boundary

Scitz0:


TrevorBenson:
@Scitz0 
What is driving the change from 5 minutes to 15? Not saying do not change it, just trying to grok the why.

Thanks

Scitz0:
> @Scitz0
> What is driving the change from 5 minutes to 15? Not saying do not change it, just trying to grok the why.
> 
> Thanks

Slots failed to send for a couple of people at the last epoch boundary. It seems the epoch boundary calculations take longer to process on pooltool end. 5 -> 15 was added to give pooltool backend more time to do all processing needed.

TrevorBenson:
> Slots failed to send for a couple of people at the last epoch boundary.

If after this PR is merged we see similar issues reported at the 15 minute mark I suggest we consider randomizing the submission by +/- 5 minutes so reduce the quantity of parallel submissions. Not saying thats the root cause as I haven't had any issues with sendslots myself, but with tightly coupled NTP time a fixed point in time for all SPO's to submit could also cause high resource utilization.

TrevorBenson (approved):
LGTM

Scitz0:
> > Slots failed to send for a couple of people at the last epoch boundary.
> 
> If after this PR is merged if we see similar issues reported at the 15 minute mark I suggest we consider randomizing the submission by +/- 5 minutes to reduce the quantity of parallel submissions. Not saying thats the root cause as I haven't had any issues with sendslots myself, but with tightly coupled NTP time a fixed point in time for all SPO's to submit could also cause high resource utilization.

Sure, randomizing it could be good to add. Though currently, it is randomized within a 60s window as this is the default sleep rate of cncli.sh.

PR #830, opened 2021-03-09
Prereqs script brew fix

MattHalloran:
On MacOS, scripts/prereqs.sh would not complete because the brew install command was treating the package list as one name. The exact error I was getting is:

> No available formula or cask with the name "coreutils gnupg jq libsodium tcptraceroute".

Removing the quotes around the package list seems to fix it.


rdlrt (approved):
Approved, thanks

PR #832, opened 2021-03-09
Update yum install error message to include all arguments

bruceharrison1984:
This just updates the error message to include all args when packages fail to install.

Not including the `--allowerasing` flag in the error message is confusing since the printed command run in RHEL7 just fine, but fails with the `--allowerasing` flag. Including the flag allows the user to try to command that was actually run.


Scitz0 (approved):
LGTM, makes sense to include in the output.

PR #833, opened 2021-03-09
Update docker build script

MattHalloran:
In an earlier commit, guild-topology.sh was created, and master-topology.sh was deleted. The alpha 3 docker build script was not updated to reflect this, and thus failing.

Note: I'm not actually sure how to run these tests before the pull request is created, so if anyone can enlighten me that'd be great:)


redoracle:
> In an earlier commit, guild-topology.sh was created, and master-topology.sh was deleted. The alpha 3 docker build script was not updated to reflect this, and thus failing.
> 
> Note: I'm not actually sure how to run these tests before the pull request is created, so if anyone can enlighten me that'd be great:)

Hi Matt, 

thanks for your feedbacks. The PR has been created.

You can still use the dockerfile_stage3 which is up to date, is not much different from the alpha version.

Anyhow in order to test it, you can change the reference manually until the PR will be approved.

rdlrt (approved):

PR #840, opened 2021-03-15
Re-arrange default for variables in env

rdlrt:
Version Check moved to after defaults are set. This prevents cnode.sh from starting as relay instead of Core node if someone for example is using master (1.25.0) versus tagged release (1.25.1)


hodlonaut (approved):
LGTM

Scitz0 (approved):
LGTM, can't foresee any issue with this.

PR #849, opened 2021-03-16
Remove launchpad & update epel-release in prereqs

rdlrt:


Crypto2099 (approved):
Looks good

Scitz0:
Looking good, can see that PR is merged but we should cleanup/delete `files/config-launchpad.json` as well.

PR #851, opened 2021-03-18
Replace wall_ms with nodeStartTime based calculation, 

rdlrt:
as rts is not populated in 1.26, but nodeStartTime is present across the versions


Scitz0 (approved):
LGTM

PR #853, opened 2021-03-20
Adapt for changes in cardano-node 1.26.x

Scitz0:
- [x] Update Protocol parameters as per new naming convention
- [x] Dont fail at startup due to lack of ERA Identifier check
- [x] Remove era identifiers where appropriate for cardano-cli execution
  - [x] For now, any calls to build-tx dont resolve ${ERA_IDENTIFIER} , this automatically bypasses adding era to the transaction build commands.
- [x] CNTools code adapted for changes in ledger-state dump used by 'Pool >> Show'

### DIFF
--- a/docs/Scripts/cntools-changelog.md
+++ b/docs/Scripts/cntools-changelog.md
@@ -5,7 +5,7 @@ All notable changes to this tool will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [8.1.0] - 2021-03-22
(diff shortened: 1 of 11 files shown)


Scitz0:
> * The complexity here is probably going to be relevant for offline mode transactions. One of the ways could be always expecting latest (mary) era, and using TTL that is based on current slot calculated from genesis.

Building an offline transaction in one era and submitting it in another is most likely a rare use case. I think it could be fine to consider this a limitation?

rdlrt:
> Building an offline transaction in one era and submitting it in another is most likely a rare use case. I think it could be fine to consider this a limitation?

Yip, happy with this approach too.

rdlrt (approved):
LGTM

PR #865, opened 2021-03-28
node 1.27.0, cncli 2.1.0, CNTools 8.3.0, DB-Sync/PGREST integration and more...

rdlrt:
(Created by @Scitz0 )

- new min node version: 1.27.0
- new min cncli version: 2.1.0
- Utilize the new pool-params/stake-snapshot cli query introduced in 1.27.0 instead of ledger-dump(CNTools & cncli.sh)
- New env variable called `PGREST_API`.
- New library function isPoolRegistered() for verifying if a pool is registered or not using either simple reg cert file detection (if REST API not set/reachable) or proper dbsync lookup using REST API. Used by Pool >> Show|List|Register|Modify
- Pool >> Show now use PostgREST API(if set), or the new pool-params cli query as fallback method.
- Menu has been re-designed with both back & home options. Instead of returning to home menu after the completed operation user is returned to the last menu.
- A new Advanced option called `Chain Queries` added.
  - Menu is dynamically built based on queries(JSON files) in DBSYNC_QUERY_FOLDER (env variable, default files/dbsync/queries) three levels deep.
  - A download option lets the user download the latest uploaded queries on Guild Operators GitHub site.
  - Query files
    - Contains menu path, description, variables, and queries(multiple)
    - Executes a predefined DBSync RPC/function through PostgREST API
    - Variables used in RPC call can either be user input, CNTools variables like EKG metrics, or an item in the result from a previous query(in the same query file)
    - Result from RPC call can either be printed or silent(only to be used for later query)
    - Output is printed as JSON
- CNTools changelog updated (date set in future, update before master merge!)
- A new script file is added called deploy-dbsync-functions.sh to deploy needed functions to dbsync PostgreSQL DB. To be run on the system that is hosting the db [**not downloaded by prereqs.sh currently**]
- Option to mint/burn assets in hybrid mode
- Transaction >> Sign now automatically tries to find the correct signing keys instead of having the user manually select the correct files
- 1.27.0 introduced a few changes in CLI commands for assets minting/burning
- Script update code unified between scripts


scripts/cnode-helper-scripts/env
```diff
@@ -30,6 +30,7 @@ CNODE_PORT=6000                                         # Set node port
 #USE_EKG="N"                                            # Use EKG metrics from the node instead of Promethus. Promethus metrics(default) should yield slightly better performance
 #TIMEOUT_LEDGER_STATE=300                               # Timeout in seconds for querying and dumping ledger-state
 #IP_VERSION=4                                           # The IP version to use for push and fetch, valid options: 4 | 6 | mix (Default: 4)
+#PGREST_API="http://127.0.0.1:8050"                     # Uncomment and set to enable PostgREST API queries against DB-Sync
```
rdlrt, line 33: Wonder if we should add a test via netcat or something to check and unset PGREST_API if the IP:port is not available?

scripts/cnode-helper-scripts/deploy-dbsync-functions.sh
```diff
@@ -0,0 +1,249 @@
+#!/bin/bash
+# shellcheck disable=SC1090,SC2086
+
+PARENT="$(dirname "$0")" 
+. "${PARENT}"/env offline
+
+if ! command -v psql &>/dev/null; then 
+  echo -e "${FG_RED}ERROR${NC}: psql command not found, make sure that you have Cardano DBSync setup correctly"
+  echo -e "\nhttps://cardano-community.github.io/guild-operators/#/Build/dbsync\n"
```
rdlrt, line 9: Should we point this to https://cardano-community.github.io/guild-operators/#/Appendix/postgres ?

scripts/cnode-helper-scripts/cntools.sh
```diff
@@ -2306,50 +2309,83 @@ function main {
               fi
               tput rc && tput ed
             fi
+            current_epoch=$(getEpoch)
             getPoolID ${pool_name}
             if [[ ${CNTOOLS_MODE} = "OFFLINE" ]]; then
-              [[ -f "${POOL_FOLDER}/${pool_name}/${POOL_REGCERT_FILENAME}" ]] && pool_registered="YES" || pool_registered="NO"
-              [[ -f "${POOL_FOLDER}/${pool_name}/${POOL_DEREGCERT_FILENAME}" ]] && ledger_retiring="?" || ledger_retiring=""
-            else
+              isPoolRegistered ${pool_name}
```
rdlrt, line 2315: I wonder if this should work in OFFLINE mode at all, as I assume offline mode should mean no internet connection at all

scripts/cnode-helper-scripts/env
Scitz0: sounds like a good idea

scripts/cnode-helper-scripts/deploy-dbsync-functions.sh
Scitz0: 👍

scripts/cnode-helper-scripts/cntools.sh
Scitz0: True, the offline way of checking if the pool is registered is to look for a pool reg cert, but I guess most only use the transaction -> sign option or maybe wallet list to get wallet address when in offline mode.

Scitz0 (approved):

PR #867, opened 2021-03-29
Fix ledger-state dump in cncli.sh

Scitz0:
A change was made in 1.26.x when using --out-file in ledger-state dump to save the file in binary format instead of plain json. Changed to redirect output to file instead of using built in --out-file.


gufmar (approved):
LGTM

PR #868, opened 2021-03-30
Fix Relay registration as per feedback from testers

rdlrt:


Scitz0 (approved):
👌 
Though a quick note regarding changelog. We usually bump patch version regularly in alpha/testing to notify users of pending update. But then consolidate all changes in a single changelog to be pushed to master? Ie in this case, should we not join 8.1.0 and 8.1.1 and just call it 8.1.1?

rdlrt:
> 👌
> Though a quick note regarding changelog. We usually bump patch version regularly in alpha/testing to notify users of pending update. But then consolidate all changes in a single changelog to be pushed to master? Ie in this case, should we not join 8.1.0 and 8.1.1 and just call it 8.1.1?

While I view it the same as you, I feel from audit pov going ahead, it could be useful trace for quick overview, especially given that few users were already using node-1.26.x branch

PR #878, opened 2021-04-07
Support for user installed bash

nikolaypavlov:
On systems like Amazon Linux with older versions of bash you may encounter errors like that:
```
./scripts/cntools.sh: line 795: mapfile: -d: invalid option
mapfile: usage: mapfile [-n count] [-O origin] [-s count] [-t] [-u fd] [-C callback] [-c quantum] [array]
```

This patch allows to pick the newest version installed by the user elsewhere. This is also [recommended shbang to improve  portability](https://stackoverflow.com/questions/10376206/what-is-the-preferred-bash-shebang).


rdlrt:
Thanks, added a commit to make it consistent across other scripts as well

rdlrt (approved):

Scitz0 (commented):
LGTM

PR #879, opened 2021-04-07
ipv4 v6 checks

gufmar:
split ipv4, ipv6 and dns hostname check
don't stop whole script if a single CUSTOM_PEER entry is invalid


Scitz0 (approved):
👍

gufmar (commented):
lgtm

Issue #887, opened 2021-04-09
cnode-helper-scripts/env: .hasEKG can be an array or scalar


nyetwurk:
https://github.com/cardano-community/guild-operators/blob/3f301c94348ba77bad6ff3ce05a7a3a47cdf4a29/scripts/cnode-helper-scripts/env#L187

Support EKG as either:

```bash
jq -r '.hasEKG | if .|type=="array" then .[1] else . end'
```

Scitz0:
Thanks, included in a future update. #888

Issue #884, opened 2021-04-08
-bash: export: `PI-POOL_HOME=/home/ada/pi-pool': not a valid identifier


wcatz:
**Describe the bug**
The environment I am attempting to create with interactive mode is creating incorrect variable.

**To Reproduce**
1. start prereqs.sh interactively
2. Please enter the project path (default: /opt/cardano): /home/ada
2. Please enter directory name (default: cnode): pi-pool
3. source $HOME/.bashrc

**Expected behavior**
I would like to see a folder 'pi-pool' in $HOME with a working export in .bashrc.
I see it is capitalized for for bash variable but it is not happy with the '-'.
option to create folder name separately or whatever fixes this behavior
ie export PI_POOL_HOME=/home/ada/pi-pool

**Screenshots**
-bash: export: `PI-POOL_HOME=/home/ada/pi-pool': not a valid identifier

**Version:**
 - OS: Ubuntu 20.10 aarch64
 - Product version: latest
 - Cardano Node version: 1.26.1
 - Network you're connecting to: mainnet
**Additional context**
Just a thank you

rdlrt:
Use of non-alpha numeric characters in the name of top level folder is not recommended - not because it is difficult to handle on specific flavour, but can cause trouble on different flavours for different variables.

wcatz:
Thank you, I realize it is more of a feature request.

PR #888, opened 2021-04-09
fixes for reported GitHub issues

Scitz0:
- Restrict to alnum characters in prereqs interactive mode [#884]
- Handle both array and scalar format for hasEKG in config file [#887]

### DIFF
--- a/scripts/cnode-helper-scripts/prereqs.sh
+++ b/scripts/cnode-helper-scripts/prereqs.sh
@@ -153,6 +153,7 @@ if [[ "${INTERACTIVE}" = 'Y' ]]; then
   clear
   CNODE_PATH=$(get_input "Please enter the project path" ${CNODE_PATH})
   CNODE_NAME=$(get_input "Please enter directory name" ${CNODE_NAME})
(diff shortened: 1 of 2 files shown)


scripts/cnode-helper-scripts/prereqs.sh
```diff
@@ -153,6 +153,7 @@ if [[ "${INTERACTIVE}" = 'Y' ]]; then
   clear
   CNODE_PATH=$(get_input "Please enter the project path" ${CNODE_PATH})
   CNODE_NAME=$(get_input "Please enter directory name" ${CNODE_NAME})
+  CNODE_NAME=${CNODE_NAME//[^[:alnum:]]/_}
```
rdlrt, line 156: I think we cannot have this here without updating the update mechanism in other scripts - as the tools doing update in place would likely place wrong vnames

rdlrt (approved):

scripts/cnode-helper-scripts/prereqs.sh
rdlrt: ny bad, we're not allowing different foldername from env var, all good

PR #913, opened 2021-05-11
Added logic to hoover down hw-cli archive on Mac OS

hodlonaut:
tar command to open the archive seems to be working on Mac just like it does on Linux so didn't bother with temporary rename of the output file.


rdlrt (approved):
LGTM

PR #932, opened 2021-05-21
IOG updates

rdlrt:
- Remove Genesis Hash from config templates, as there are different hydra builds with different genesis hashes (as prepwork for alonzo)
- Swap default switches `-l` for prereqs and cabal-build-all.sh to use custom fork of libsodium by default, and use system libsodium *ONLY* if `-l` argument is added to prereqs and/or cabal-build-all scripts
- env: checkUpdate - Add executable permission after downloading file


gufmar (approved):
LGTM

Scitz0 (approved):
👍  Added a commit to make EKG default option.

PR #946, opened 2021-05-31
bump node/cli versions to 1.27.0

DamjanOstrelic:


Scitz0:
We try not to merge directly to master to keep the flow consistent and branches in line, eg. PR -> alpha -> master
So, please create a new PR using alpha as the base branch instead. Other than that 👍

DamjanOstrelic:
> We try not to merge directly to master to keep the flow consistent and branches in line, eg. PR -> alpha -> master
> So, please create a new PR using alpha as the base branch instead. Other than that 👍

will do

PR #948, opened 2021-06-01
Docs typos and updates

DamjanOstrelic:
I have been going through the docs and fixing typos, and updating some commands if needed.

The styling is subjective, I did it as I think looks okay so let me know if you're okay with it or not.

This PR includes `Home`, `Basics`, `Build and Run`, and `Docker` sections. If it's good, I will make other PRs for the rest of the sections. 

For this one, some notes/questions:

- `cardano-wallet` -> switched the git checkout command to get the latest release, however, the latest release is not compatible with the latest `cardano-node` release as per official documentation. I have tested it with `1.27.0` and it seems to work though... So my question is do we checkout against `master` or latest release in the docs?
- `metadata-tool` - I have left it as checkout to `master` in the docs, not sure why now, but let me know if to switch to the latest release or keep it as is.

### DIFF
--- a/docs/Build/wallet.md
+++ b/docs/Build/wallet.md
@@ -20,9 +20,9 @@ cd cardano-wallet
 
 You can use the instructions below to build the latest release of [cardano-wallet](https://github.com/input-output-hk/cardano-wallet).
 
-!> - Note that the latest release of `cardano-wallet` may not work with the latest release of `cardano-node`. Please check the compatibility of each `cardano-wallet` release yourself in the official docs, e.g. https://github.com/input-output-hk/cardano-wallet/releases/tag/v2021-05-26.
+!> - Note that the latest release of `cardano-wallet` may not work with the latest release of `cardano-node`. Please check the compatibility of each `cardano-wallet` release yourself in the official docs, e.g. https://github.com/input-output-hk/cardano-wallet/releases/latest.
 
-> The cardano-wallet repo does not work yet with cabal, hence alternate for now is using stack to build
+> The cardano-wallet repo does not work yet with `cabal`, hence the alternative for now is building with `stack`
 
 ``` bash
 git fetch --tags --all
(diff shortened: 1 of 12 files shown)


docs/Build/wallet.md
```diff
@@ -18,67 +18,79 @@ cd cardano-wallet
 
 ##### Build Cardano Wallet
 
-You can use the instructions below to build the cardano-node, same steps can be executed in future to update the node (replacing appropriate tag) as well.
+You can use the instructions below to build the latest release of [cardano-wallet](https://github.com/input-output-hk/cardano-wallet).
+
+!> - Note that the latest release of `cardano-wallet` may not work with the latest release of `cardano-node`. Please check the compatibility of each `cardano-wallet` release yourself in the official docs, e.g. https://github.com/input-output-hk/cardano-wallet/releases/tag/v2021-05-26.
```
rdlrt, line 23: We could use https://github.com/input-output-hk/cardano-wallet/releases/latest instead of the `tag/v2021-05-26`

rdlrt (approved):
Thanks, LGTM, just 1 comment about wallet tag

rdlrt:
> * `cardano-wallet` -> switched the git checkout command to get the latest release, however, the latest release is not compatible with the latest `cardano-node` release as per official documentation. I have tested it with `1.27.0` and it seems to work though... So my question is do we checkout against `master` or latest release in the docs?

I think the way we have it now (as part of this PR) in the guide is good - the Note asks users to check compatibility in release notes, and then it's up to user whether they want to use specific tag.

> * `metadata-tool` - I have left it as checkout to `master` in the docs, not sure why now, but let me know if to switch to the latest release or keep it as is.

Until the component is bought to the same 'visibility' as other IOG components, I fear we'd have to stick with master - so that's good

docs/Build/wallet.md
DamjanOstrelic: > Until the component is bought to the same 'visibility' as other IOG components, I fear we'd have to stick with master - so that's good

Hehe, yeah that was my feeling as well.

PR #951, opened 2021-06-06
Doc updates

DamjanOstrelic:
Hopefully some improvements, questions:

1. In create a pool section in `cntools-common.md`, should `rotate kes keys` be after pool creation or after registration? 

2. do we need to add a note for wallet import HW (also in `cntools-common.md`) - does it need `cardano-hw-cli`? I didn't try it myself.


Scitz0:
> 1. In create a pool section in `cntools-common.md`, should `rotate kes keys` be after pool creation or after registration?

Looking into this in more detail I can see that our offline workflow diagram show to rotate after creating the pool while in the code its added to the pool registration section and instructions printed for hybrid registration(not modification) mode. I currently cant see any reason why we can not just add the op.cert generation at pool generation directly and get rid of it in pool registration. The only thing would be in testnet mode if you re-register the same pool after network reset to get a fresh start or in mainnet mode if you retire your pool and then wants to reregister it with the same cold key.

But to answer your question, it doesn't really matter if you do it after pool creation or after pool registration, as long as op.cert is generated and set to online node. 


> 2. do we need to add a note for wallet import HW (also in `cntools-common.md`) - does it need `cardano-hw-cli`? I didn't try it myself.

Could be good yes. Things to note regarding this is that the same rules apply as for mnemonic import, only the first account and first index/address of that account is derived/imported. Another limitation is that it can not me the main owner of a pool using CNTools due to the fact that we use the main owner wallet to pay for tx fee and this is not supported by hw wallets. All these limitations is outlined / printed after hw wallet import but the same information should be present in our docs as well.

Speaking of this, could be good to add a section regarding mnemonic import as well highlighting the current limitations as well as advantages it gives. 

And yes, it does require cardano-hw-cli yes and the code will check for this. It will also manually ask if the hardware device you own is one of these three models, Ledger S, Ledger X, Trezor Model T. As these are the only ones supported. 

Mnemonic info: https://github.com/cardano-community/guild-operators/blob/alpha/scripts/cnode-helper-scripts/cntools.sh#L490-L506
HW wallet import: https://github.com/cardano-community/guild-operators/blob/alpha/scripts/cnode-helper-scripts/cntools.sh#L570-L584


The CNTools docs hasn't really kept up with all the development. I admit that I don't find this as rewarding/fun to spend time on as actually coding so always end up as a TODO. Really great that you have taken the time to look into this. Feel free to add any stuff you see missing, like explanation for how to access advanced menu option by providing -a runtime argument or enable in settings(disabled by default). I'm sure there is other stuff as well.

PR #955, opened 2021-06-08
Make PGRest queries independent of index/ID to utilise Restful-ness of postgREST

rdlrt:
- [x] get_active_stake
- [x] get_delegator_count
- [x] get_pool_metadata
- [ ] get_pool_owners (need to handle multi-owner)
- [x] get_pool_relays
- [x] get_pool_retire
- [x] get_pool_update
- [x] get_pool_opcert

- [x] Update dependent tools
CNTools Pool >> Show updated to utilize above updated functions.
Pool KES rotation now using get_pool_opcert to get the latest counter number from chain.
General message metadata to funds >> send added (unrelated to PR but was easier to add in this PR)


DamjanOstrelic:
Is there an easy way to test the changes with `deploy-dbsync-rpc.sh`? I don't know how to configure it to get jsons from your branch

Scitz0:
> Is there an easy way to test the changes with `deploy-dbsync-rpc.sh`? I don't know how to configure it to get jsons from your branch

I haven't looked into the changes yet as I'm at work and a bit busy but when you run the script it prints an example curl command to execute the rpc locally.

DamjanOstrelic:
> > Is there an easy way to test the changes with `deploy-dbsync-rpc.sh`? I don't know how to configure it to get jsons from your branch
> 
> I haven't looked into the changes yet as I'm at work and a bit busy but when you run the script it prints an example curl command to execute the rpc locally.

Yes, that part is okay. I meant, how do I deploy the functions from this branch `pgrest-additions`, and not what's in the `alpha` branch because `deploy-dbsync-rpc.sh` fetches function jsons from the `alpha` branch I think.

rdlrt:
> how do I deploy the functions from this branch `pgrest-additions`, and not what's in the `alpha` branch because `deploy-dbsync-rpc.sh` fetches function jsons from the `alpha` branch I think.

The branch is dependent on .env_branch in $CNODE_HOME/scripts folder, so for example - if you start gLiveView using -b pgrest-dnd argument, your next instance will check updates from this branch - as that's what .env_branch would contain

DamjanOstrelic:
Seems to be an issue with passing `_pool_hash_id` argument to the subquery:

```
SELECT get_active_stake('pool1z2ry6kxywgvdxv26g06mdywynvs7jj3uemnxv273mr5esukljsr', 122);
ERROR:  column "_pool_hash_id" does not exist
LINE 6: ...s.pool_id =  (SELECT id from pool_hash where view=_pool_hash...
                                                                                                  ^
QUERY:  SELECT ( SELECT json_build_object(
'active_stake_sum', SUM (es.amount)
)
FROM epoch_stake AS es
WHERE es.epoch_no = _epoch_no
AND es.pool_id =  (SELECT id from pool_hash where view=_pool_hash_id)
)
CONTEXT:  PL/pgSQL function get_active_stake(text,numeric) line 14 at RETURN
```

rdlrt:
> Seems to be an issue with passing 

Wierd , seems I did not git add from right copy...it should be _pool_bech32 on that line, will edit it from mobile in a few

DamjanOstrelic:
> > Seems to be an issue with passing
> 
> Wierd , seems I did not git add from right copy...it should be _pool_bech32 - updated now

Works okay now, so did we decide to switch the queries to be ID independent rather than the pre-query Ola talked about?

Scitz0:
Looks good to me, haven't tested the change though. Even if it means adding an extra table lookup I think in the end it might be the better solution than relying on querying the same instance for every chained query.

Is there any case where there could be multiple entries in DB for the same bech32? Eg. if a pool is retired and then re-registered for example? Or it will just reuse the old entry in that case?

Scitz0:
Though not sure the other queries is as "easy" to replace as pool bech32 vs id.

DamjanOstrelic:
> Though not sure the other queries is as "easy" to replace as pool bech32 vs id.

Yes, exactly, but we can do it hehe, is the plan to have all queries listed above on this one PR then and not 1 PR per query? If so, do I make a PR to the `pgrest-additions` branch if I translate a query?

rdlrt:
> , is the plan to have all queries listed above on this one PR then and not 1 PR per query

Yes, this way we can continue to reuse same branch.



> Though not sure the other queries is as "easy" to replace as pool bech32 vs id.

Yip, but the idea would be similar, just size of query might prolly increase

DamjanOstrelic:
I tested the queries, they all seem to work but I have 2 questions:
1. get_pool_retire.json -> if it is not retired, there is no return, I guess that's how CNTools is set up to work?
2. get_pool_update.json -> I cannot seem to get any valid result from this one e.g. `curl -d _pool_bech32=pool1pvyt2d468tlzr77cymae90hgj73aret457zfktnvgev6kmx5nk3 -s http://localhost:8050/rpc/get_pool_update`
to NULL the `active_epoch_no` and `state` fields, I thought I would get all the updates (and there must be at least one when the pool is registered)? But instead I just get null:

```
curl -d _pool_bech32=pool1pvyt2d468tlzr77cymae90hgj73aret457zfktnvgev6kmx5nk3 -s http://localhost:8050/rpc/get_pool_update
null
```

rdlrt:
> 1. if it is not retired, there is no return

I had added a seperate ticket on trello for handling default cases...but currently, yes - tho there are breaking changes added, which will need to be updated on CNTools end (the last item in this PR)

> 2\. pool1pvyt2d468tlzr77cymae90hgj73aret457zfktnvgev6kmx5nk3

I do not see this pool on guild network

DamjanOstrelic:
> I do not see this pool on guild network

Ah you're right, the example query for that one wasn't updated, probably uses mainnet. Not a big issue, but I think it's the only outlier in the example commands that doesn't use the guild network.

files/dbsync/rpc/get_pool_owners.json
```diff
@@ -1,29 +1,31 @@
...
     "BEGIN",
       "RETURN ( SELECT json_agg(js) json_final FROM ( SELECT json_build_object(",
-          "'owner_hash', encode(po.hash::bytea, 'hex'),",
+          "'owner_addr', tout.address,",
           "'reward_addr', sa.view",
         ") js",
         "FROM pool_owner AS po",
-        "LEFT JOIN stake_address AS sa ON RIGHT(encode(sa.hash_raw::bytea, 'hex'), -2) = encode(po.hash::bytea, 'hex')",
-        "WHERE po.registered_tx_id = _pool_tx_id",
+        "LEFT JOIN stake_address AS sa ON sa.id = po.addr_id",
```
Scitz0, line 25: addr_id is not a column of pool_owner table. 
I think this has to be reverted to previous SQL command:
`LEFT JOIN stake_address AS sa ON RIGHT(encode(sa.hash_raw::bytea, 'hex'), -2) = encode(po.hash::bytea, 'hex')`

Scitz0 (commented):
Looks good, all rpc passed my test except owners. See comment for that.

Scitz0:
Hmm, ran into a problem looking into updating CNTools. I grab both latest and active pool update from DB, this to compare and display if something in the registration has been updated and is set to go live in upcoming epoch. With changes queries, it will always grab the latest.

files/dbsync/rpc/get_pool_owners.json
DamjanOstrelic: I have that column:

2021-06-09 23_01_07-pgAdmin 4

rdlrt:
> Looks good, all rpc passed my test except owners. See comment for that.

Just checked, it looks like you might be using dbsync executable from 9.0, the `addr_id` got added [right after](https://github.com/input-output-hk/cardano-db-sync/commit/6f9c0346ad88ee9e1031a4269f14e952bf233619#diff-df0ad6e43880f09c90ebf95f19110178aba6890df0010ebda7485029e2b543b4) the release was marked - but before release was shipped.

Also, there was a major schema update on Apr-07 on dbsync end, alongwith another one yesterday on master branch for alonzo.   As checked with Erik, it looks like the next release is almost ready to be tagged - hence, would be nice if we use `release/10.0.x` for this branch.

files/dbsync/rpc/get_pool_owners.json
Scitz0: Ok, updating dbsync on my end 👍

Scitz0:
Updated CNTools to use the new rpc. Though haven't fully tested it. Mainly used for Pool >> Show. 
Need to create a pool modification and verify updated data. Also metadata update.

Pool retire rpc need an update. I previously handled this in CNTools but we need to move this functionality to rpc instead.
This line in cntools.library:
`[[ $(jq '.announced_tx_id' <<< ${pretire}) -lt $(jq '.[0].registered_tx_id' <<< ${pupd_latest}) ]] && retiring_epoch=0 # invalidated by newer update`
Eg. its possible to send a pool retire tx and then later invalidate it by sending a pool update/modify tx. So need to return retire row only if that is later than latest pool update.

Scitz0:
pool owner rpc needs to be fixed to support multi owner pool. Looked briefly at it but didn't have the time to fix it.

Other than this, CNTools is compatible and tested for the changes.

Scitz0 (approved):
Approving PR with a note to update pool owner rpc to handle multi-owner setups. A separate issue will be created for this to keep track.

PR #962, opened 2021-06-19
KES calculation update and prereqs fix

Scitz0:
KES calculation moved from CNTools & gLiveView into a common function in env file. It will by default use node metrics with provided pool KES start period as fallback.

prereqs.sh script changed to properly handle non-existent branches and not try to replace content of script if file was never downloaded or if content is corrupt.

**PS!!**
Needs to be properly tested before merged as I haven't had time to do this.

### DIFF
--- a/scripts/cnode-helper-scripts/env
+++ b/scripts/cnode-helper-scripts/env
@@ -86,19 +86,6 @@ OFFLINE_MODE='N'
 [[ $(basename $0 2>/dev/null) = "topologyUpdater.sh" ]] && OFFLINE_MODE='Y' # for backwards compatibility
 
 [[ -z ${BRANCH} ]] && { [[ -f "${PARENT}"/.env_branch ]] && BRANCH=$(cat "${PARENT}"/.env_branch) || BRANCH=master; }
-REPO_RAW="https://raw.githubusercontent.com/cardano-community/guild-operators"
-
-if [[ ${OFFLINE_MODE} = N && ${BRANCH} != master && ${BRANCH} != alpha ]]; then
-  if ! curl -s -f -m ${CURL_TIMEOUT} -o config.json.tmp ${REPO_RAW}/files/config-mainnet.json 2>/dev/null ; then
-    echo -e "WARN!! The folder was configured against ${BRANCH} branch - which does not exist anymore, falling back to alpha branch"
-    BRANCH=alpha
-    echo "${BRANCH}" > "${CNODE_HOME}"/scripts/.env_branch
-  fi
-fi
-
-URL_RAW="${REPO_RAW}/${BRANCH}"
-URL="${URL_RAW}/scripts/cnode-helper-scripts"
-URL_DOCS="${URL_RAW}/docs/Scripts"
 
 export LC_ALL=C.UTF-8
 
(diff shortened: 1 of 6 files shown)


scripts/cnode-helper-scripts/env
```diff
@@ -389,7 +376,20 @@ getAnswerAny() {
 #             : 2 = update failed
 checkUpdate() {
   [[ "${UPDATE_CHECK}" != "Y" ]] && return 0
-  if curl -s -f -m 10 -o "${PARENT}/${1}".tmp "${URL}/${1}" 2>/dev/null; then
+
+  if [[ ${OFFLINE_MODE} = N && ${BRANCH} != master && ${BRANCH} != alpha ]]; then
+    if ! curl -s -f -m ${CURL_TIMEOUT} "https://api.github.com/repos/cardano-community/guild-operators/branches" | jq -e ".[] | select(.name | contains(\"${BRANCH}\"))" &>/dev/null ; then
```
rdlrt, line 381: Works for most cases, just an edge case here - `cntools-8.4.1` vs `cntools-8.4.11` might not get caught

scripts/cnode-helper-scripts/prereqs.sh
```diff
@@ -418,7 +418,7 @@ echo "Downloading files..."
 pushd "${CNODE_HOME}"/files >/dev/null || err_exit
 
 
-if ! curl -s -f -m ${CURL_TIMEOUT} -o config.json.tmp ${URL_RAW}/files/config-mainnet.json 2>/dev/null ; then
+if ! curl -s -f -m ${CURL_TIMEOUT} "https://api.github.com/repos/cardano-community/guild-operators/branches" | jq -e ".[] | select(.name | contains(\"${BRANCH}\"))" &>/dev/null ; then
```
rdlrt, line 421: same as above

scripts/cnode-helper-scripts/env
```diff
@@ -89,6 +85,8 @@ OFFLINE_MODE='N'
 [[ $(basename $0 2>/dev/null) = "cnode.sh" ]] && OFFLINE_MODE='Y' # for backwards compatibility
 [[ $(basename $0 2>/dev/null) = "topologyUpdater.sh" ]] && OFFLINE_MODE='Y' # for backwards compatibility
 
+[[ -z ${BRANCH} ]] && { [[ -f "${PARENT}"/.env_branch ]] && BRANCH=$(cat "${PARENT}"/.env_branch) || BRANCH=master; }
```
rdlrt, line 88: Should we inherit BRANCH from parent environment variable (just got caught out coz I sourced env file before in my session using alpha branch)?

rdlrt (approved):
Happy with the rest, just some minor comments for alignment, once discussed - we should be good to merge

scripts/cnode-helper-scripts/env
Scitz0: correct, better to do exact match, fixing.

scripts/cnode-helper-scripts/env
Scitz0: What about -b flag on runtime?
For let's say topologyUpdater.

PR #969, opened 2021-06-28
Pre-source env in online mode where applicable to avoid setting OFFLINE_MODE to 'Y' in parent scripts

rdlrt:
Using `./env offline` enforces variable OFFLINE_MODE to "Y" in scripts, rendering checkUpdate calls to exit without actual checks


Scitz0 (approved):
LGTM

PR #972, opened 2021-07-08
Update docs

rdlrt:


Scitz0 (approved):
👌

docs/Build/pgrest.md
```diff
@@ -36,15 +37,16 @@ ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT ON TABLES TO web_anon;
...
 Now that your user is created in database, you can create the postgrest config. A basic sample is below:
 
-!> Replace `${USER}` and `${PASSWORD}` to match the username and password you can authenticate with for `psql` (defined in `$CNODE_HOME/priv/.pgpass` file in our example).
+!!! info "Reminder !!"
+    Replace `${USER}` and `${PASSWORD}` to match the username and password you can authenticate with for `psql` (defined in `$CNODE_HOME/priv/.pgpass` file in our example).
 
 ``` bash
 cat << 'EOF' > $CNODE_HOME/priv/pgrest.conf
-db-uri = "postgres://${USER}:${PASSWORD}@localhost/cexplorer"
+db-uri = "postgres://${USER}@/cexplorer"
```
DamjanOstrelic, line 49: I think I am missing something here. Why can we leave just the `@`? If we really can, then probably the line above should change as well:

`Replace `${USER}` and `${PASSWORD}` ...`

Scitz0: Check information here:
https://postgrest.org/en/v7.0.0/configuration.html#db-uri

Scitz0: but yeah, maybe the text should change as you say to remove ${PASSWORD}

rdlrt: @DamjanOstrelic the sample deployment instructions for Postgres and PostgREST were updated to use socket instead of host and work using peer authentication, so that we dont have to tinker with `ident`/`md5`  + not have hardcoded passwords amywhere. The format for db-uri when using passwords is correct.
Updated comment right before, nice catch

PR #975, opened 2021-07-11
Add multiple pool owners support

DamjanOstrelic:


Scitz0:
I did not get the expected result with this query:
I expect it has to do with pool updates. Got 19 results for the pool I tested with but should get 3. 

And we should remove bech32 address, its not part of owner registration and has nothing to do with owner. 

What about this instead:

```
    "CREATE OR REPLACE FUNCTION get_pool_owners(_pool_bech32 text)",
    "RETURNS JSON STABLE LANGUAGE PLPGSQL AS $$",
    "BEGIN",
    "  RETURN ( SELECT json_agg(js) json_final FROM ( SELECT json_build_object(",
    "      'reward_addr', sa.view",
    "    ) js",
    "    FROM pool_owner po",
    "    LEFT JOIN (",
    "      SELECT ph.id AS ph_id, MAX(po.registered_tx_id) AS po_registered_tx_id",
    "      FROM pool_hash ph",
    "      LEFT JOIN pool_owner po ON po.pool_hash_id = ph.id",
    "      WHERE po.pool_hash_id = (SELECT id FROM pool_hash WHERE view=_pool_bech32)",
    "      GROUP BY ph.id",
    "      ORDER BY ph.id) AS lu ON lu.ph_id = po.pool_hash_id",
    "    LEFT JOIN stake_address AS sa ON sa.id = po.addr_id",
    "    WHERE po.registered_tx_id = lu.po_registered_tx_id",
    "    ORDER BY lu.ph_id",
    "  ) t );",
    "END; $$;"
```

DamjanOstrelic:
Which pool did you test with?

Removing the bech32 address would make it simpler, and I would do it this way in that case:

```
SELECT json_agg(js) json_final FROM ( SELECT json_build_object(
         'owner', sa.view
        ) js
        FROM pool_owner po
        LEFT JOIN stake_address AS sa ON sa.id = po.addr_id
        WHERE po.registered_tx_id = (
          SELECT MAX(pool_owner.registered_tx_id)
          FROM pool_owner
          INNER JOIN pool_hash ON pool_owner.pool_hash_id = pool_hash.id
            and pool_hash.view=_pool_bech32
          ) GROUP BY sa.view ) t
```

As a stake address could have multiple payment addresses, it makes sense to remove it I think (even though CLI ones have only one). But it still might be useful information... Not sure if to build a separate json object with owner payment addresses associated with the owner stake address, or just return stake addresses...

Scitz0:
> Which pool did you test with?

My mainnet pool that I know have three owners registered.
pool1qqq6qqa0hpzvumv5p87ynczfmdj557xuwlc3289ke42g72z7f74


> Removing the bech32 address would make it simpler, and I would do it this way in that case:

This updated query worked great 👍 


> One thing I did think of is what happens if the owner's stake address is deregistered? Do we care about checking that?

For this query I don't think its something we should care about, the query is about what owners are registered for the pool, not if they are valid or not. Can be a separate query to get the status of a stake address. 


> As a stake address could have multiple payment addresses, it makes sense to remove it I think (even though CLI ones have only one). But it still might be useful information... Not sure if to build a separate json object with owner payment addresses associated with the owner stake address, or just return stake addresses...

Yeah, a stake address can have any combination of payment addresses to form the base address so it doesn't make sense in this instance to include. A separate query that returns all **used** addresses associated with a stake address might not be bad but not really needed at the moment.

DamjanOstrelic:
> > Which pool did you test with?
> 
> My mainnet pool that I know have three owners registered.
> pool1qqq6qqa0hpzvumv5p87ynczfmdj557xuwlc3289ke42g72z7f74
> 
> > Removing the bech32 address would make it simpler, and I would do it this way in that case:
> 
> This updated query worked great 👍
> 
> > One thing I did think of is what happens if the owner's stake address is deregistered? Do we care about checking that?
> 
> For this query I don't think its something we should care about, the query is about what owners are registered for the pool, not if they are valid or not. Can be a separate query to get the status of a stake address.
> 
> > As a stake address could have multiple payment addresses, it makes sense to remove it I think (even though CLI ones have only one). But it still might be useful information... Not sure if to build a separate json object with owner payment addresses associated with the owner stake address, or just return stake addresses...
> 
> Yeah, a stake address can have any combination of payment addresses to form the base address so it doesn't make sense in this instance to include. A separate query that returns all **used** addresses associated with a stake address might not be bad but not really needed at the moment.

Thanks for the feedback, I agree 👍  Will change the query later today.

Scitz0 (approved):
Verified to be working after the small fix applied.

PR #989, opened 2021-07-22
Make dbsync.sh env check offline

rdlrt:


hodlonaut (approved):
What's our official definition of the 'offline' parameter by the way? Does it mean it's-okay-for-the-node-to-be-stopped essentially?

Scitz0 (approved):
👍

rdlrt:
> What's our official definition of the 'offline' parameter by the way? Does it mean it's-okay-for-the-node-to-be-stopped essentially?

Essentially it is not-to-check for things via a running node, and rely on config items, folder structure, etc as defaults (using user variables). This brings basic essentials from env that to initialize those parameters in absence of a running node.

PR #993, opened 2021-07-24
Stake Distribution table (grest schema)

DamjanOstrelic:
💥 This query creates and populates the `stake_distribution` table that we can use for:
 - live stake (total or individual pools)
 - stake address balances
 - stake address totals in more detail (rewards, withdrawals, available rewards...)
 
It automatically updates existing entries and inserts new entries on subsequent runs.

Considerations:
 - Need to figure out whether to extract the `CREATE TABLE` logic to somewhere else (as part of the grest setup scripts) and directory structure where this query belongs.
 - Because this would be an automated query to run on schedule we might separate those kinds of queries from the rest that run on request.

Performance on a full run (prefix with `EXPLAIN ANALYZE` to test):
``` 
Planning Time: 6.161 ms
Execution Time: 264105.294 ms (4.4 minutes)
```

Performance on update run:
```
 Planning Time: 5.219 ms
 Execution Time: 268819.863 ms (4.48 mins)
```

More execution time testing from multiple instances would be useful to get better approximations.


rdlrt:
Thanks, just marking do-not-merge for now to confirm timings , as well as go through tickets to see if any updates needed (i.e. any details we should capture in addition to this one)

files/grest/rpc/stake_address/stake_distribution.sql
```diff
@@ -0,0 +1,106 @@
+CREATE TABLE IF NOT EXISTS GREST.STAKE_DISTRIBUTION (
```
rdlrt, line 1: I think this will likely be a trigger eventually or a cron-query, so we prolly might wanna move it to cache-queries folder

DamjanOstrelic: Yep, I was thinking the same. I put it in here because everyone can just copy and run it without worrying about anything for testing phase.

I think the best option is to move the creation of ALL tables we want to eventually create in our `grest` schema to be a part of the `grest-setup.sh` script.

rdlrt: Hmm, given that this will run multiple times at scheduled intervals, I feel it can stay as a scheduled query, regardless of state of setup script. The setup script would download these cache queries and schedule them if needed

files/grest/rpc/stake_address/stake_distribution.sql
```diff
@@ -11,15 +11,15 @@ CREATE TABLE IF NOT EXISTS GREST.STAKE_DISTRIBUTION (
     PRIMARY KEY (STAKE_ADDRESS)
 );
 
-INSERT INTO GREST.STAKE_DISTRIBUTION
+EXPLAIN ANALYZE INSERT INTO GREST.STAKE_DISTRIBUTION
```
rdlrt, line 14: Looks like this got back :smile:

files/grest/rpc/stake_address/stake_distribution.sql
DamjanOstrelic: Ah, I see what you mean now... I thought you were referring to the stake_distribution query as the scheduled one.

Yeah I think 2 sets of scheduled queries make sense:
1. grest schema setup queries -> ensures we always have correct structure even if we drop and recreate db
2. data queries that populate grest schema like the one below

Did I understand correctly?

rdlrt: I originally meant having both as part of same (since it wouldnt be recreating if one already exists)..but yea both seperately works too

files/grest/rpc/stake_address/stake_distribution.sql
```diff
@@ -0,0 +1,156 @@
...
+CREATE TABLE IF NOT EXISTS GREST.CONTROL_TABLE (
+    control_type text PRIMARY KEY,
+    value text NOT NULL
+);
+
+DO $$
+DECLARE
+    _last_accounted_block_height bigint;
+BEGIN
+    SELECT
```
rdlrt, line 22: Might be unnoticeable (ms) but we could use MAX function here to pluck biggest value from index

files/grest/rpc/stake_address/stake_distribution.sql
```diff
@@ -0,0 +1,156 @@
...
+    UTXO numeric,
+    REWARDS numeric,
+    WITHDRAWALS numeric,
+    REWARDS_AVAILABLE numeric,
+    RESERVES numeric,
+    TREASURY numeric
+);
+
+CREATE TABLE IF NOT EXISTS GREST.CONTROL_TABLE (
+    control_type text PRIMARY KEY,
```
rdlrt, line 14: Query type (eg: stake distribution), Block hash  (nullable in this case), Block ID

This control table can be filled by different queries with their properties, and easier to update row based on their last checks(?)

DamjanOstrelic: This was totally the first draft only so I would expect changes here yeah. I remember us mentioning adding an entry for db-sync version for example. So I just made the entires as general as possible, i.e. control_type doesn't include just queries, and value is a text field so it can be filled with arbitrary data.

We can add block hash if we want to as a nullable field - unless we know only query entries will go in this table.

rdlrt: I see - I guess the mapping done that way is easier to include bits indeed...in that case, we could add 3 columns:
key, last_value, artifacts - which could then include vvaried (JSON, content depending on query) things

DamjanOstrelic: That sounds good, I'll update it here. Another thing I was thinking about we should have a centralised query where this control table is created.

We _can_ have it on every query that needs it, but then if we ever change the structure of it we would have to change it on every query instead of just one place.

rdlrt (approved):
Approving as agreed on last meeting, to add/habdle delta inserts seperately

PR #995, opened 2021-07-25
Rewards history RPC function

DamjanOstrelic:
Marked draft because I'm not sure whether to include deposit refunds. I think it makes returning a nice JSON much harder. And deposit refunds are not really rewards.

Not sure if I got the docs part correct 😄


rdlrt:
Sorry for being a bit of a pain, but knowing how this has caused issue(s) for other explorer/wallet providers, would it make sense to have stake address history table which includes type of distribution (thus, rewards/catalyst/deposit/withdrawal/etc)?

If so, this could become easier for taxation too (one of the important use cases for this type of data)

DamjanOstrelic:
> Sorry for being a bit of a pain, but knowing how this has caused issue(s) for other explorer/wallet providers, would it make sense to have stake address history table which includes type of distribution (thus, rewards/catalyst/deposit/withdrawal/etc)?
> 
> If so, this could become easier for taxation too (one of the important use cases for this type of data, and possible reduce providing duplicate queries). Other option would be grest api provides singular queries for each of above, but then it might have more endpoints to worry about and maintain

Yeah, I think it's not a problem. The only thing is that in that case, the `catalyst/deposit/withdrawal` part should be real-time as it's based on transactions and not epochs, so we would need to come up with a good schedule for those. Because of that, I would divide the rewards history query as it only needs to be run once in an epoch and the other ones, but make them fill the same table.

DamjanOstrelic:
@rdlrt We talked about this in the last meet... We have no way of distinguishing between the various `treasury` payments... Deposit refunds, catalyst rewards, (itn rewards?)... So we should probably sync up on what our plan is here. I think it would be simple to do a structure like:

```
address | amount | epoch | type
addr...    | 3.5        |   267    | reward
addr...    | 1.5        |   268    | treasury
addr...    | 3.5        |   269    | withdrawal
```
if that's what we want for now...

rdlrt:
> @rdlrt We talked about this in the last meet... We have no way of distinguishing between the various treasury payments...
The withdrawals/deposits will be tracked much easier from input-output-hk/cardano-db-sync#721 .

But for the rest , I agree with your example with simplification

files/grest/rpc/stake_address/rewards_history.sql
```diff
@@ -0,0 +1,41 @@
...
+                json_agg(js) json_final
+            FROM (
+                SELECT
+                    json_build_object('epoch_no', reward.epoch_no, 'stake_pool', pool_hash.view, 'reward_amount', ROUND((reward.amount / 1000000), 2)) js
+                FROM
+                    reward
+                    INNER JOIN stake_address ON reward.addr_id = stake_address.id
+                    INNER JOIN pool_hash ON reward.pool_id = pool_hash.id
+                WHERE
+                    stake_address.view = _stake_address_bech32
```
rdlrt, line 33: Just noticed filter is missing  , tho epoch_no is now divided into earned/spendable

DamjanOstrelic: Yep, got it working last night, was just waiting for the ada/lovelace decision. Will update soon

rdlrt: ye - avoiding ADA keeps it consistent with chain unit too

DamjanOstrelic: Updated now, along with adding a type for the rewards according to: 
https://github.com/input-output-hk/cardano-db-sync/commit/5965ee3e87730ec3f8c93b9287320c11f94d2160

> `member` or `leader` should have  `poolId` entry while
> for `treasury` or `reserves` should always have NULL for the `poolId` field.

Some things to clear up:
1. If `addr...` is passed instead of `stake...`, do we want to do a lookup or require only stake address to be passed in?
2. If an epoch is specified, do we consider it as the `spendable` or `earned` epoch for filtering (currently `earned`)?

rdlrt: 1. For now, my view is prefer to only have stake... address - keeping it simple - until dbsync stabilises in terms of changes, as addr to stake addr mapping is also currently available seperately
2. earned indeed

DamjanOstrelic: In that case, I think this is ready for removing the draft label

rdlrt (approved):

PR #997, opened 2021-08-01
Account balance RPC

DamjanOstrelic:
Almost perfect... There are edge cases with some special deposit withdrawals. E.g. `stake1uywjy7h05jmhx9y3wzy94td6xz4txynuccgam0zfn800v8qq33z29` -> adastat and cardanoscan calculate available rewards as `1.005749`, this query just zeroes it out as `withdrawals > rewards`.

But other than that, I think it's pretty good.

Testing (mainnet):

```
select grest.account_balance('addr1vxcnz5w5ccv3nwerjm944hl6hjzdc6axm893c88a9jcsrpsd22m4p');

select grest.account_balance('addr1q8q2dtulm8t4pfq8vwlxljne95vjd3s00jfkjazx3g7genkq56helkwh2zjqwca7dl98jtgeymrq7lynd96ydz3u3n8qyz3r65');

select grest.account_balance('DdzFFzCqrhsuQV5ohP1EBh9d4UjmmWDb2Vui2s2SACSBDeEW74kHhp3zj9hvLFDb9mK7hRr7fbctu1UYatPaF1omXiXPwf12bn7DQvbT');

--deregistered
select grest.account_balance('stake1uyruvdz88v34kx3wu4zw87n8kn9kzgcluq7duv0efaxcn2skac3ry');
select grest.account_balance('addr1q9f67ad5zkanl068leetzpfr4gggaey2dt53kman9uq2mvs8cc6ywwertvdzae2yu0ax0dxtvy33lcpumccljn6d3x4qtwt7cs');


--deregistered
select grest.account_balance('stake1uywjy7h05jmhx9y3wzy94td6xz4txynuccgam0zfn800v8qq33z29');
select grest.account_balance('addr1q82knuya3axzgqq08wdflcw5p6q442rr06422rp4mnqgm2cayfawlf9hwv2fzuygt2km5v92kvf8e3s3mk7ynxw77cwqayujrp');
```


rdlrt (approved):
LGTM, drop a case on tg to check if that is similar to discrepancy that you see - but other than that looks awesome

DamjanOstrelic:
> LGTM, drop a case on tg to check if that is similar to discrepancy that you see - but other than that looks awesome

Still needs a more rigid check on the rewards epochs included

rdlrt:
Results for tests:
- The output may not be consistent across dbsync instances, might need a way to find out which instance is latest and ping them instead.
- The descrepancy around orphan'd rewards should be resolved by input-output-hk/cardano-db-sync#721 , we will need to append our queries accordingly soon

DamjanOstrelic:
> * When I query a stake address the first time , it takes 1-3 seconds, eventual calls are almost instant. Wonder if this could create an issue if a consumer has to loop through a lot of addresses (shouldn't be an issue to DB itself, but the performance from consumer pov could be a bit mediocre)
This sounds like it comes down to postgres performance across instances... All calls have been instant on my end, with the Byron address taking 1 full second (with subsequent runs 0.25 seconds). We probably need to set some approach/standard on how to best mitigate the differences.

> * Possible proposal: We keep blocks/transaction/address-balance RPCs in blockchain folder instead of maintaining seperate one, as they can be inter-related to traverse from explorer pov + can be tagged together in OpenAPI
Agreed, I'll change it.

rdlrt:
Merging current state as best-effort until fixes upstream, will need to update in future as there are a number of schema changes being added

PR #1001, opened 2021-08-05
Allow variables to inherit from environment if already set

rdlrt:
(comment from tg:
RESTAPI_PORT/HAPROXY_PORT are currently unused, they would be once we decide endpoint-based polling)


Scitz0 (changes requested):
Still static values in code to dbsync. Need to use the newly added variables for this. Other than this 👍

Scitz0 (approved):

PR #1002, opened 2021-08-07
Update totals rpc

rdlrt:
- Include more details from ada_pots
- Rename active_stake to supply and use utxo as circulation


DamjanOstrelic (approved):
👌

PR #1007, opened 2021-08-08
[cntools] Added check_KES to toggle startup KES check

Straightpool:
Fixes the following edge case:
- Node holds block producer VRF keys to check leaderlog
- Node does not have access to block producer EKG port
- Node can thus not receive KES EKG data breaking the KES startup check of cntools 
- The simple toggle in cntools.config allows disabling CHECK_KES at startup, by default the check is true


Scitz0 (approved):
Looks good to me

PR #1008, opened 2021-08-09
Change result object from JSON to Table for rewards_history and totals

rdlrt:
- PostgREST natively adds certain parameters for filtering (eg: `select` for columns, `limit` and `offset` for pagination, etc), in addition to bringing result as a dictionary. Thus, would be good to stick to RETURN TABLE where possible, as other RETURN objects do not follow configuration limits (eg: `max-rows` )

- Rename `_stake_address_bech32` to `_stake_address` (consistent with `_address` on account_balance) as all parameters we use are bech32 format


DamjanOstrelic (approved):
👍

PR #1010, opened 2021-08-09
Latest block rpc

DamjanOstrelic:
Open to discussion about what info to add/remove, but I think it's a good starting point.

### DIFF
--- a/files/grest/rpc/blockchain/account_balance.sql
+++ b/files/grest/rpc/blockchain/account_balance.sql
@@ -1,3 +1,5 @@
+DROP FUNCTION IF EXISTS grest.account_balance (text);
+
 CREATE OR REPLACE FUNCTION grest.account_balance (_address text)
     RETURNS json STABLE
     LANGUAGE PLPGSQL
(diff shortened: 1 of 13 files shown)


rdlrt:
Just thinking, what is the use case for latest block? If it is to represent tip information, we should give back slot/epoch/height.
The complete (rest of information) should be in /rpc/blocks endpoint

rdlrt (approved):

files/grest/rpc/blockchain/account_balance.sql
```diff
@@ -1,3 +1,5 @@
+DROP FUNCTION IF EXISTS grest.account_balance (text);
+
 CREATE OR REPLACE FUNCTION grest.account_balance (_address text)
```
Scitz0, line 3: Insignificant but if droped first then OR REPLACE can be removed as well.

rdlrt: good point, added commit to keep all queries uniform

PR #1012, opened 2021-08-10
First cut of get transactions for given address

hodlonaut:
First cut of get transactions for given address


rdlrt:
@DamjanOstrelic @hodlonaut - should we consider addition of assets to this PR (It could be a sub-item already as part of account assets, but could be slightly useful)? Ofcourse that further complicates addresses that are too large, which could be difficult to manage in absence of adding indexes to base DB.

DamjanOstrelic:
> @DamjanOstrelic @hodlonaut - should we consider addition of assets to this PR (It could be a sub-item already as part of account assets, but could be slightly useful)? Ofcourse that further complicates addresses that are too large, which could be difficult to manage in absence of adding indexes to base DB.

Yes, I think it's a good idea, we should at least try it.

files/grest/rpc/blockchain/txs_for_addr.sql
```diff
@@ -0,0 +1,90 @@
...
+                WHERE
+                    txo.index = txi.tx_out_index
+                    AND txo.tx_id = txi.tx_out_id
+                    AND txi.tx_in_id = txb.id) in_addrs, (
+                SELECT
+                    jsonb_agg(txo.address) AS addr_list FROM tx_out txo
+            WHERE
+                txo.tx_id = txb.id GROUP BY txo.tx_id) out_addrs, CASE WHEN EXISTS (
+                SELECT
+                    NULL FROM delegation d
```
DamjanOstrelic, line 38: Could this be an issue if multiple addresses are in the transaction and it's a delegation by the address we are not interested in?

Might need to join on `addr_id` as well here. Same for withdrawals.

DamjanOstrelic:
I think it would be very nice to include the result values of the transactions for the address (`+/-` lovelace) as well.

files/grest/rpc/blockchain/txs_for_addr.sql
```diff
@@ -0,0 +1,90 @@
+DROP FUNCTION IF EXISTS grest.addr_txs (text, uinteger, character varying);
+
+CREATE FUNCTION grest.addr_txs (_address text, _limit uinteger DEFAULT 1000, _orderby character varying DEFAULT 'desc' ::character varying)
+    RETURNS TABLE (
+        in_addrs jsonb,
+        out_addrs jsonb,
+        tx_hash text,
+        block_idx uinteger,
```
rdlrt, line 8: Should be OK to remove this variable from the table (?)

files/grest/rpc/blockchain/txs_for_addr.sql
rdlrt: I think a 'yes' / 'no' is still valid at transaction level imo

rdlrt:
Update:
Need to split this into /rpc/addr_txs (resulting in tx + block hash and height) and /rpc/tx_info (which can give as much details as possible about input transaction hashes supplied).

Have been playing with combinations for cache query (incl. returning input/outputs), looks like calls split for tx_info vs addr

PR #1020, opened 2021-08-15
Add db size metric (monitoring)

DamjanOstrelic:
Do we want this to track the growth of the DB?

We can also add the sizes of separate schemas (`public, grest`)to track them individually?

### DIFF
--- a/files/grest/grafana/guild-dashboard.json
+++ b/files/grest/grafana/guild-dashboard.json
@@ -16,6 +16,7 @@
   "editable": true,
   "gnetId": null,
   "graphTooltip": 0,
+  "id": 24,
   "links": [
     {
       "icon": "external link",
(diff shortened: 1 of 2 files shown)


Scitz0 (approved):
Good suggestion. I think it is good to track grest and public schema separately if possible.

rdlrt (commented):
If we're capturing schemas seperately, should we still have entire database too? I think major time from grest-poll is due to sleep time for capturing CPU (as it is difference between 2 points), so it's not a big difference to capture all 3 as seperate metrics too

DamjanOstrelic:
> If we're capturing schemas seperately, should we still have entire database too? I think major time from grest-poll is due to sleep time for capturing CPU (as it is difference between 2 points), so it's not a big difference to capture all 3 as seperate metrics too

Yeah, good point, we should at least simply add them together rather than querying the total size independently in that case.

I also wanted to add disk usage % monitoring for instances - what do you think?.

rdlrt:
> > If we're capturing schemas seperately, should we still have entire database too? I think major time from grest-poll is due to sleep time for capturing CPU (as it is difference between 2 points), so it's not a big difference to capture all 3 as seperate metrics too
> 
> Yeah, good point, we should at least simply add them together rather than querying the total size independently in that case.
> 
> I also wanted to add disk usage % monitoring for instances - what do you think?.

disk usage % - imo - is prolly only good for individual monitoring alerts - and it may also depend on mounts (eg: where pg data folder is, where dbsync ledger-state is, node is)

DamjanOstrelic:
> > > If we're capturing schemas seperately, should we still have entire database too? I think major time from grest-poll is due to sleep time for capturing CPU (as it is difference between 2 points), so it's not a big difference to capture all 3 as seperate metrics too
> > 
> > 
> > Yeah, good point, we should at least simply add them together rather than querying the total size independently in that case.
> > I also wanted to add disk usage % monitoring for instances - what do you think?.
> 
> disk usage % - imo - is prolly only good for individual monitoring alerts - and it may also depend on mounts (eg: where pg data folder is, where dbsync ledger-state is, node is)

Yep, my thinking was to ensure that instances have free space left and add alerts as we don't know the disk size (even with disk size and db size, we still don't know what else is on the instance) for each instance and it can vary.

DamjanOstrelic:
@rdlrt @Scitz0 This has been updated for schema sizes and 3 panels added to the dashboard.

Please check if we need to do some changes, if not I think we can remove the `do not merge` and go to alpha.

rdlrt (approved):

PR #1021, opened 2021-08-16
Epoch info cache

DamjanOstrelic:
Co-author: @redoracle 

Combines epoch info and epoch params into one table that is automatically updated through 2 triggers:

1) updates current epoch info
2) inserts new epoch row on epoch transition 

We can then create endpoints for anything epoch related without doing any `JOIN`s on each request.

Still to-do:
- [x] Trigger for epoch insert
- [ ] RPC functions


files/grest/cache_queries/epoch_info_cache_setup.sql
```diff
@@ -0,0 +1,204 @@
...
+CREATE TEMP TABLE temp_active_stake AS
+SELECT
+    epoch_no,
+    SUM(es.amount) AS i_active_stake
+FROM
+    epoch_stake es
+GROUP BY
+    es.epoch_no;
+
+CREATE TEMP TABLE temp_last_block_time AS SELECT DISTINCT ON (epoch_no)
```
rdlrt, line 62: Not sure if I understand this one, does this differ greatly from public.epoch table?

files/grest/cache_queries/epoch_info_cache_setup.sql
```diff
@@ -0,0 +1,204 @@
...
+    epoch_no,
+    time DESC;
+
+CREATE TEMP TABLE temp_epoch_start_end_time (
+    epoch serial PRIMARY KEY,
+    start_time timestamp without time zone NOT NULL,
+    end_time timestamp without time zone NOT NULL
+);
+
+INSERT INTO temp_epoch_start_end_time
```
rdlrt, line 77: Also start and end time are pre-populated in epoch table I believe?

files/grest/cache_queries/epoch_info_cache_setup.sql
```diff
@@ -0,0 +1,204 @@
+DROP TABLE IF EXISTS grest.epoch_info_cache;
+
+DROP TABLE IF EXISTS temp_active_stake;
```
rdlrt, line 3: Unless really needed, would be good to avoid writing to public schema, same for other tables below

files/grest/cache_queries/epoch_info_cache_setup.sql
```diff
@@ -0,0 +1,204 @@
...
+    RETURN NEW;
+END;
+$epoch_info_update$
+LANGUAGE plpgsql;
+
+DROP TRIGGER IF EXISTS epoch_info_update_trigger ON public.block;
+
+CREATE TRIGGER epoch_info_update_trigger
+    AFTER INSERT ON public.block
+    FOR EACH ROW
```
rdlrt, line 202: Discussed with damjan that we'd prolly want to reduce frequency of this call using intervals, as most (not all) of the information will likely be almost same throughout the epoch, possibly even liaise with cache query intervals, can use 15m for initial mock

rdlrt (approved):

DamjanOstrelic:
@rdlrt Added the new epoch trigger - tested on guild network only. I think it's ready to go to alpha and RPC can be separate PRs?

rdlrt:
@DamjanOstrelic - yip , we just need to add to API spec

PR #1022, opened 2021-08-16
Create registered_pools.sql

hodlonaut:


DamjanOstrelic (approved):
👍 I added the API doc update only. I'll let you decide whether you want to do it on the other PRs or delegate it to me 😄 

Just one question: Does this belong to blockchain queries or pool queries? @rdlrt

And one note: Some explorers only show pools with some stake - there are around ~150 pools that have 0 stakes but are registered. It's just a note, I wouldn't change anything on this RPC because of it.

rdlrt:
Discussed with @hodlonaut about possibly merging #1022 , #1023 and #1024 as addendum to rpc/pool_updates , as we could have single endpoint for these with date, type of action

DamjanOstrelic:
> Discussed with @hodlonaut about possibly merging #1022 , #1023 and #1024 as addendum to rpc/pool_updates (if pool_updates has pool_id , it would be limited to single pool - else all pool), as we could have single endpoint for these with date, type of action

Does that allow users to query lists of pools based on status though? I thought that was the main goal of these queries. What you mentioned about date and type of action makes sense to me to add to the `rpc/pool_updates`, but it seems like a separate use-case.

rdlrt:
> Does that allow users to query lists of pools based on status though? 

Yes ofcourse - as part of suggested change. As I said, if pool_updates has input pool ID , only then return single pool . The other (second) parameter could be type of registration, instead of `latest` that we currently have.

PR #1027, opened 2021-08-18
Epoch info/params RPCs + Swagger doc updates

DamjanOstrelic:


rdlrt (approved):

docs/Build/grestsvcs.json
```diff
@@ -569,7 +577,7 @@
       }
     },
     "/epoch_info_cache": {
-      "get": {
+      "post": {
```
rdlrt, line 580: GET works, prolly preferable too - something I'd try and update for others - where RETURN is table

docs/Build/grestsvcs.json
```diff
@@ -569,7 +577,7 @@
       }
     },
     "/epoch_info_cache": {
```
rdlrt, line 579: We can actually remove epoch_info_cache (I had added it before myself), I think we only need to add a control table with status

PR #1032, opened 2021-08-22
Account addresses rpc

DamjanOstrelic:
1) Do we want to include UTXO values for each individual address here?
2) Do we want to include the `stake_address` with which all of these addresses are associated to?

### DIFF
--- a/docs/Build/grestsvcs.json
+++ b/docs/Build/grestsvcs.json
@@ -540,6 +540,39 @@
         }
       }
     },
+    "/rpc/account_addresses": {
+      "post": {
+        "tags": [
+          "Blockchain Queries"
+        ],
+        "summary": "Get all addresses associated with an account",
+        "produces": [
+          "application/json",
+          "application/vnd.pgrst.object+json"
+        ],
+        "parameters": [
+          {
+            "required": true,
+            "schema": {
+              "type": "object",
+              "properties": {
+                "_address": {
+                  "format": "text",
+                  "type": "string"
+                }
+              }
+            },
+            "in": "body",
+            "name": "args"
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "OK"
+          }
+        }
+      }
+    },
     "/rpc/account_assets": {
       "post": {
         "tags": [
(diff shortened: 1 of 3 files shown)


Scitz0:
> 1. Do we want to include UTXO values for each individual address here?
I think the flow goes like this to paint a complete picture of an account:
1) Find all addresses used that belong to this account
2) Find all transactions that have a connection to any of these addresses
3) Get all utxo's for the above transactions.
4) Get details for each utxo, e.g. inputs and outputs, each with an array of assets/amount. 
But these would be separate queries in my mind. 

With this said, It might also be interesting to be able to get active unspent utxo's that belong to any of these addresses. Query 4 from above could then be used to get further details about the utxo.  

> 2. Do we want to include the `stake_address` with which all of these addresses are associated to?
I dont think thats necessary, would just result in the same stake address for each row in the table. We could have a separate query that returns the stake address given a base address.

DamjanOstrelic:
> I dont think thats necessary, would just result in the same stake address for each row in the table. We could have a separate query that returns the stake address given a base address. Basically a small part of this query.

Yea, you're right. I was still thinking we could return a JSON, but we switched to returning tables 😄

DamjanOstrelic:
> I think the flow goes like this to paint a complete picture of an account:
> 
> 1. Find all addresses used that belong to this account
> 2. Find all transactions that have a connection to any of these addresses
> 3. Get all utxo's for the above transactions.
> 4. Get details for each utxo, e.g. inputs and outputs, each with an array of assets/amount.
>    But these would be separate queries in my mind.
> 
> With this said, It might also be interesting to be able to get active unspent utxo's that belong to any of these addresses. Query 4 from above could then be used to get further details about the utxo.

I guess somewhere between points 2. and 4., or as point 5. would be to find all historical transactions with their inputs/outputs as well to get a complete picture of an account.

So for this query, are you saying what is there is enough (i.e. completes step 1)?

rdlrt (approved):
Agree with Ola for scope of this PR and the flow (as some of the items in that flow are already underway on seperate PR)

PR #1033, opened 2021-08-24
Restructure rpc directories

DamjanOstrelic:
can we also get rid of `rpc/queries/` directory? And put `cached_tables/` in its place so that they are outside of `rpc/`?


Scitz0 (approved):
LGTM

rdlrt (approved):

PR #1035, opened 2021-08-25
Bump postgres version

rdlrt:
Better for any new instance to directly use pg13, one of the important updates is Re-Index on DB is seamless (does not need an outage)


Scitz0 (approved):
👍

PR #1044, opened 2021-08-29
Adds pool_info_cache table setup

Scitz0:


Scitz0:
Leaving PR as do-not-merge for now until I have cleaned up pool query rpc's and modified them to use this new cache table.

rdlrt:
LGTM, need to add doc + move retiring+retire endpoints to this one 👍🏻

files/grest/rpc/cached_tables/pool_info_cache_setup.sql
```diff
@@ -0,0 +1,203 @@
...
+
+COMMENT ON FUNCTION grest.pool_info_update IS 'Internal function to insert all new pool updates into pool_info cache table';
+
+
+-- Create pool_info_update trigger
+DROP TRIGGER IF EXISTS pool_info_update_trigger ON public.block;
+
+CREATE TRIGGER pool_info_update_trigger
+    AFTER INSERT ON public.block
+    FOR EACH ROW
```
DamjanOstrelic, line 202: Is it better to use `FOR EACH STATEMENT` here?

From docs (https://www.postgresql.org/docs/9.1/sql-createtrigger.html):

> A trigger that is marked FOR EACH ROW is called once for every row that the operation modifies. For example, a DELETE that affects 10 rows will cause any ON DELETE triggers on the target relation to be called 10 separate times, once for each deleted row. In contrast, a trigger that is marked FOR EACH STATEMENT only executes once for any given operation, regardless of how many rows it modifies (in particular, an operation that modifies zero rows will still result in the execution of any applicable FOR EACH STATEMENT triggers).

DamjanOstrelic (approved):
👍

files/grest/rpc/cached_tables/pool_info_cache_setup.sql
Scitz0: But can you use the NEW "object" in the called trigger function then?
I use the id of the block that causes the trigger to fire to compare against.

DamjanOstrelic: I think you can still use it. My understanding is it only affects during-sync execution when you can have batches of blocks inserted, and then `FOR EACH ROW` would execute the procedure for each of the e.g. 5000 blocks in a batch. But after sync, I think we can only get 1 by 1 block inserts so I think there is no difference in that case.

files/grest/rpc/pool/pool_list.sql
```diff
@@ -0,0 +1,27 @@
...
+    SELECT
+        DISTINCT ON (pic.pool_id_bech32) pool_id_bech32,
+        pic.pool_id_hex,
+        pod.ticker_name
+    FROM
+        grest.pool_info_cache AS pic
+    LEFT JOIN
+        public.pool_offline_data AS pod ON pod.pmr_id = pic.meta_id
+    ORDER BY
+        pool_id_bech32, tx_id DESC;
```
DamjanOstrelic, line 23:
```suggestion
        pic.pool_id_bech32, tx_id DESC;
```

DamjanOstrelic (approved):
Some results from my testing and thoughts:

1) `pool_blocks`, `pool_delegators` - work based on comparing data to cardanoscan.io
2) `pool_info` - works but there are a lot of joins - I'm not sure if it can be improved, but at least the active stake sum can be switched to https://github.com/cardano-community/guild-operators/pull/1080 when finished to avoid summing the values on each request. I would like to see block count, op cert and so on all in the pool_info_cache, but I understand we have to stop caching at some point 😄 
3) `pool_list` - for me, there is a small bug because of the `SELECT DISTINCT ON` - suggested change is what fixed it for me.
4) `pool_metadata`, `pool_relays` - works

Perhaps we would like to make 3) and 4) more granular? At the moment, they return full lists but if users would like just data for 1 pool, the function could accept an argument for it rather than using pgREST's `...?pool_id.eq=pool_id` or similar syntax? I understand this info is also in the `pool_info` endpoint, but that might be too much unnecessary data for the user in this case.

Note: `rpc/pool/pool_updates.sql` seems to have been left as an empty file.

files/grest/rpc/pool/pool_list.sql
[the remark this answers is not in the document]
Scitz0: 👍  
vscode forgot to save some files for me. pool_update and pool_list rpc now fixed. 
2. agree on active stake, the others I think should be left outside of pool info cache table according to TG chat.

PR #1049, opened 2021-08-31
Fixed Adrestia architecture doc URL

dehy:


DamjanOstrelic:
Just something to consider as an alternative link as well: https://docs.cardano.org/explore-cardano/cardano-architecture/overview/

Not sure what's better, probably the IOHK one cause it has short component descriptions.

rdlrt (approved):

PR #1057, opened 2021-09-02
multi instances

dehy:
Fixes a bug where multiple instances cannot be installed on the same machine

**Describe the bug**
When installing two instances with different names on a same server (for testing purpose), the second instance on the initial directory and not the custom one.

**To Reproduce**
1. On a fresh server, install an instance with `prereqs.sh -t cnode` (explicit default for demo)
2. Install a second instance with `prereqs.sh -t relay1`
3. There is only one directory : /opt/cardano/cnode

**Expected behavior**
The second instance is located at `/opt/cardano/relay1`

### DIFF
--- a/scripts/cnode-helper-scripts/prereqs.sh
+++ b/scripts/cnode-helper-scripts/prereqs.sh
@@ -2,8 +2,6 @@
 # shellcheck disable=SC2086,SC1090
 # shellcheck source=/dev/null
 
-unset CNODE_HOME
-
 ##########################################
 # User Variables - Change as desired     #
 # command line flags override set values #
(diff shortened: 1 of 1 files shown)


rdlrt:
Not sure if I understand the issue here, the CNODE_HOME reference is only within prereqs.sh , not outside - when using `-t` it would pick a new CNODE_NAME - which is propogated throughout for _HOME.

I will test it out shortly if there's an issue introduced recently

rdlrt:
Nice catch - it is due to sourcing bashrc itself, but given that CNODE_HOME is all we add to bashrc (other than build ones which are already pre-sourced) , there was no reason to source bashrc again. Linked PR should fix it

PR #1060, opened 2021-09-03
Skeleton to fill guild mainnet node instances

rdlrt:
For consistency in naming location, I used `curl -k https://json.geoiplookup.io/<IP>` to use continent_code/country_code/city (stripping out spaces).


ilap (approved):
LGTM

PR #1073, opened 2021-09-06
Changed path for the customise() function

redoracle:
changed to /opt/cardano/cnode/files from /opt/cardano/cnode

So we can preserve files in ../priv


Scitz0 (approved):
👍

PR #1080, opened 2021-09-08
Active stake cache setup

DamjanOstrelic:
At the moment, we have to fire the trigger on every insert (chunks of 2000 entries) as we cannot know when the inserts to `epoch_stake` are completed for an epoch: https://github.com/input-output-hk/cardano-db-sync/issues/797

Once that's added, we can improve the efficiency of the trigger by having it run only once instead.

Note that the initial insert into the cache table takes a while, as it needs to process 150k+ rows.


rdlrt (approved):

DamjanOstrelic:
NOTE: This should be tested on mainnet today before being merged.

DamjanOstrelic:
Need to test one more time on epoch 291/292 transition because of new changes.

PR #1091, opened 2021-09-15
Add CARDANO_NODE_SOCKET_PATH environment variable

Asafffff:
This is required for issuing `cardano-cli` commands

### DIFF
--- a/files/docker/node/dockerfile_stage2
+++ b/files/docker/node/dockerfile_stage2
@@ -1,11 +1,11 @@
 FROM cardanocommunity/cardano-node:stage1
 
 ENV \
-CNODE_HOME=/opt/cardano/cnode \
-DEBIAN_FRONTEND=noninteractive \ 
-LANG=C.UTF-8 \
-USER=root \
-PATH=$CNODE_HOME/scripts:/root/.cabal/bin:/root/.ghcup/bin:$PATH 
+  CNODE_HOME=/opt/cardano/cnode \
+  DEBIAN_FRONTEND=noninteractive \ 
+  LANG=C.UTF-8 \
+  USER=root \
+  PATH=$CNODE_HOME/scripts:/root/.cabal/bin:/root/.ghcup/bin:$PATH 
 
 RUN git clone https://github.com/input-output-hk/cardano-node.git \
   && export BOOTSTRAP_HASKELL_NO_UPGRADE=1 \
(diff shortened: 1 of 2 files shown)


rdlrt:
Did you mean to add it to [stage3 file ](https://cardano-community.github.io/guild-operators/docker/docker/#strategy) instead of stage2?

cryptofarmersllc:
Better to add to docker file stage 3

shibug:
Better to add to docker file stage 3

Asafffff:
> Did you mean to add it to [stage3 file ](https://cardano-community.github.io/guild-operators/docker/docker/#strategy) instead of stage2?

Yes. I'll modify it.
In addition, i saw few inconsistencies between the base cnode / cli's env variables and the env file. For example - this particular `CARDANO_NODE_SOCKET_PATH` is called `SOCKET` in the env file.

`cnode.sh` (idk if it's an external script, or maintained in this repo only) uses `POOL_DIR` instead of a concatenation between `CNODE_HOME/priv/pool/POOL_NAME` which are being used in the env file.

Also, is there some place to see the default node environment variables? Couldn't find a good resource other than reading the scripts themselves.

rdlrt:
The points you raise are all expected and dont need any changes

> In addition, i saw few inconsistencies between the base cnode / cli's env variables and the env file. For example - this particular `CARDANO_NODE_SOCKET_PATH` is called `SOCKET` in the env file.

That is necessary as there may be more than one instance running on the server, if a custom SOCKET has to be set

> `cnode.sh` (idk if it's an external script, or maintained in this repo only) uses `POOL_DIR` instead of a concatenation between `CNODE_HOME/priv/pool/POOL_NAME` which are being used in the env file.

No, the concatenation is POOL_DIR - this derivation is [already in env file](https://github.com/cardano-community/guild-operators/blob/317ec5db7b6aa9d1284b1552f25e2b1cb05e296e/scripts/cnode-helper-scripts/env#L283). And yes, cnode.sh is part of repo (look in scripts/cnode-helper-scripts folder)

> Also, is there some place to see the default node environment variables? Couldn't find a good resource other than reading the scripts themselves.

There is a page dedicated to env file (see [here](https://cardano-community.github.io/guild-operators/Scripts/env/) but essentially the user variables section are what users would want to customise (and can be inherited from Environment variables) - while the others are derived from those user variables.

rdlrt (approved):

PR #1094, opened 2021-09-15
Asset address list RPC

DamjanOstrelic:
example queries for mainnet: 

Low TX count:
`select * from grest.asset_address_list('12077ad90b05fc19d890c71a2f00fe129dd7d169ef3e277acad2ad12', '454e4f4c53');`

Large TX count:
`select * from grest.asset_address_list('2ad8fbdc6f18d97ce61b113301f05ff6b2a241c4e0bf5d06127d30a2', '4f4343'); --9365 txs`

This has a large difference in first/second runs:
4 seconds first run, 112 ms next run


rdlrt:
Took almost the same time here as well : `Total query runtime: 6 secs 29 msec.`

Scitz0 (approved):
👍

PR #1100, opened 2021-09-17
Add EXPOSE statement for c-node port

Asafffff:
`EXPOSE` serves as a good documentation of what port the image is listening on


rdlrt:
an SPO would not want to always expose the port, they may run it in private OR public mode (as documented [here](https://cardano-community.github.io/guild-operators/docker/run/) )

Asafffff:
> an SPO would not want to always expose the port, they may run it in private OR public mode (as documented [here](https://cardano-community.github.io/guild-operators/docker/run/) )

This doesn't explicitly expose the port. It just marks that this container listens by default on port X. If you want the node port to be exposed, you need to provide the real expose syntax (`-p Y:X`) where X is the port we documented in the dockerfile

shibug:
I agree with @Asafffff on this. Having expose instructions on dockerfile of recommended.

PR #1102, opened 2021-09-18
Added missed slots statistics

orpheus-antpool:
An important performance metric SPOs should be aware of


rdlrt:
> An important performance metric SPOs should be aware of

Not sure if I understand, you have leader check (which checks the whether VRF key supplied qualified to create a block) - and you have blocks forged/adopted. Why/How does missed slot add important performance metric? 🙂

orpheus-antpool:
> Not sure if I understand, you have leader check (which checks the whether VRF key supplied qualified to create a block) - and you have blocks forged/adopted. Why/How does missed slot add important performance metric? 

Because the leader checks will not be performed if other operations keep cardano-node busy. A common example is the garbage collector kicking in and lasting more than one second, which leads to a missed slot. If the pool was scheduled to produce a block during that slot, it will silently miss the opportunity to do so. This is a common problem especially at epoch boundaries. More details here: https://forum.cardano.org/t/no-more-missed-slots-during-epoch-after-changing-one-setting/67924

rdlrt:
Thanks for the background. IMO - this sits in same category as tuning/performance monitoring which better sits on monitoring layer (as against minimalistic dashboard to show if node is operating normally). Having an counter higher than 1 can often cause panic to SPOs and cause performance tuning exercise on the basis of gLiveView, but will let @Scitz0 @gufmar confirm what their views are

Regardless, to approve the PR:
1. Needs version bump.
2. The metric definition is best placed in [getNodemetrics](https://github.com/cardano-community/guild-operators/blob/alpha/scripts/cnode-helper-scripts/env#L627-L668) function, for consistency and reduction of additional call to EKG (or Prom) metric, and availability in other sections.
3. Missed Slots counter Like `BLOCKS` section are only valid since node startup

orpheus-antpool:
@rdlrt: Thanks for the feedback. If the others think this is useful, I'll update the PR accordingly. IMO, many SPOs want a minimalist setup for their BP nodes, which is why gLiveView.sh would be ideal to monitor metrics related to tuning too, thereby eliminating the need to install heavier tools like Grafana.

Scitz0:
I'm happy to include this metric with notes from @rdlrt
But I also have some additional notes that tie into @rdlrt concern for users.

1) Need to add a sentence to info section (show_home_info) that explains the metric.
2) Update [gLiveView documentation page](https://github.com/cardano-community/guild-operators/blob/alpha/docs/Scripts/gliveview.md) with the same or more information for what the value show and what to expect. The documentation page doesn't have the same space constraints as gLiveView does. Built in gLiveView docs are already close to the max size I would say. Like for example that it's expected to have this value raise on epoch boundary but during the epoch, it should stay fairly constant. Reward calculation window etc could also add to this metric so people need to be aware that there are circumstances out of their control to limit slot missed increase. But it could be useful to check for abnormal behavior.

orpheus-antpool:
OK, I've pushed an update. The missed slots are not available in EKG, only Prometheus. If EKG is used (default) missed slots will be N/A. May be worth considering Prometheus as default.

rdlrt:
> The missed slots are not available in EKG, only Prometheus.

They are:
```bash
curl -sH "Accept: application/json" http://127.0.0.1:13788 | jq .cardano.node.metrics.slotsMissedNum.int.val
0
```

> If EKG is used (default) missed slots will be N/A. May be worth considering Prometheus as default.

The reason to prefer EKG as default if availble is because in the past we have seen prometheus can have gaps every now and then in reporting back if queried frequently.

orpheus-antpool:
> The reason to prefer EKG as default if availble is because in the past we have seen prometheus can have gaps every now and then in reporting back if queried frequently.

Right, I missed this EKG metric in my jq output and I thought it does not exist. This is fixed now

Scitz0:
Reviewing PR but will most likely take a day or two due to other commitments. I have a local non-committed change to deal with the increased size of into tabs as well I will deploy together with this to not have it complain about height of output.

Scitz0:
Got a moment to look it over. I want to minimize the additional rows added as much as possible. gLiveView will fail to render on smaller screens if there are not enough rows. So each additional will make this worse. Could you revert to your original implementation by adding missed slot metric below KES values with no additional separator line? 

And the checked counter doesn't really add any beneficial value so limit it to display missed slots. If you want you could make it a bit "sexier" by adding percent missed by dividing with total slots since node start. E.g. something like this:
`printf -v missed_slot_pct %.4f "$(bc -l <<<"(${missed_slots}/(${about_to_lead}+${missed_slots}))*100")"`
this would save the percentage of missed slots compared to all slots since node start into missed_slot_pct variable. As I assume about_to_lead doesn't include missed_slots I added this to the total for calculation. 
Final output could then look something like this:
```
├──────────────────────────────────────────────────────────────┤
│ KES current/remaining         : 290 / 13                     │
│ KES expiration date           : 2021-10-07 20:20:16 UTC      │
│ Missed slot leader checks     : 237 (0.0190%)                │
│--------------------------------------------------------------│
```

orpheus-antpool:
Fair point. I've reverted to the previous format.

rdlrt (approved):

PR #1103, opened 2021-09-19
Stake distribution cron job setup

DamjanOstrelic:
- sets up the stake distribution update job to run every 30 mins (the block height difference check is still retained - in case instances go out of sync we don't run it) 
- last commit is formatting which we can revert
- moved all function declarations to the top of the file


scripts/grest-helper-scripts/setup-grest.sh
```diff
@@ -77,12 +140,12 @@ if [[ "${UPDATE_CHECK}" = 'Y' ]] && curl -s -f -m ${CURL_TIMEOUT} -o "${PARENT}"
...
     STATIC_CMD=$(awk '/#!/{x=1}/^# Do NOT modify/{exit} x' "${PARENT}"/setup-grest.sh)
-    printf '%s\n%s\n' "$STATIC_CMD" "$TEMPL2_CMD" > "${PARENT}"/setup-grest.sh.tmp
+    printf '%s\n%s\n' "$STATIC_CMD" "$TEMPL2_CMD" >"${PARENT}"/setup-grest.sh.tmp
     {
-      mv -f "${PARENT}"/setup-grest.sh.tmp "${PARENT}"/setup-grest.sh && \
-      chmod 755 "${PARENT}"/setup-grest.sh && \
-      echo -e "\nUpdate applied successfully, please run setup-grest again!\n" && \
-      exit 0;
+      mv -f "${PARENT}"/setup-grest.sh.tmp "${PARENT}"/setup-grest.sh &&
+        chmod 755 "${PARENT}"/setup-grest.sh &&
```
rdlrt, line 146: If it's a single line, best to add 2 line indents

files/grest/cron/setup-cron-jobs.sh
```diff
@@ -0,0 +1,108 @@
...
+CNODE_PATH="/opt/cardano"
+USER="$(whoami)"
+CURL_TIMEOUT=60
+
+is_empty "$CNODE_NAME" && CNODE_NAME='cnode'
+CNODE_HOME=${CNODE_PATH}/${CNODE_NAME}
+LOG_DIR="${CNODE_HOME}/logs"
+CRON_SCRIPTS_DIR="${CNODE_HOME}/scripts/cron-scripts"
+
+main() {
```
rdlrt, line 99: For a simplistic script, is there a reason we cannot merge this to setup-grest instead of seperate script here?

I can imagine for more complicated scripts, we might split it out - but we already have setup-grest, prereqs - a bit hesitant to have another script for setup

DamjanOstrelic: IMO setup-grest script is already overloaded with code. I think we should not only keep this as a separate script but also refactor setup-grest in the future to extract more modules out.

As for the env variables, when I created the script I was testing it manually, but what about if this script (and any others) are only supposed to be called from the main `setup-grest.sh`? Could we then define all the variables there and any script that gets called would have access to them?

rdlrt (approved):

PR #1107, opened 2021-09-20
gLiveView 1.21.1 - stop refresh for info pages

Scitz0:
This makes it possible to have more info without worrying about row overflow.


gufmar (approved):
Lgtm

PR #1110, opened 2021-09-21
Add account_updates RPC

DamjanOstrelic:
I don't think `withdrawal` belongs to this endpoint, but following our Trello board.

At the moment, only `tx_hash` is returned as additional info for the update action.

### DIFF
--- a/docs/Build/grestsvcs.json
+++ b/docs/Build/grestsvcs.json
@@ -408,6 +408,35 @@
         }
       }
     },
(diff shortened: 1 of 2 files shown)


Scitz0 (approved):
LGTM. Not sure about withdrawal but let's keep it for now. I think tx hash is fine. tx_info can then be used to get more details.

rdlrt (approved):

PR #1112, opened 2021-09-22
Add memory utilization statistics to gLiveView.sh

orpheus-antpool:
SPOs running on 8GB machines need to pay close attention to memory utilization and the efficiency of the garbage collector. This PR adds support to monitor live/heap data.


rdlrt:
Comment about addition:

> SPOs running on 8GB machines need to 

.. our recommendation until changes are made at source is to upgrade, as these SPOs are one resource spike/incorrect shutdown away from node having to rebuild db/ledger-state and hitting into OOM loops as entire UTxO gets loaded to memory, which will naturally accelerate more now in Alonzo era.

I agree it's important to monitor those metrics, but it makes sense to look at it over a period of time (vs point-in-time dashboard like gLiveView), and I'd repeat such monitoring should be done by a proper monitoring service - that does not necessarily have to run on same node as the cardano-node(s).

About the PR, the addition can wrap it stats next line as below:

image

orpheus-antpool:
Fully agreed, monitoring with a logging service is needed. This is not a replacement, just for convenience (simple live view of all info in a single window). Not sure what you mean by wrapping stats, the figure is clipped. Can you please elaborate?

rdlrt:
> the figure is clipped. Can you please elaborate?

Right, Updated image in previous comment 🙂

orpheus-antpool:
I cannot reproduce this issue. Here's how it looks on my end. This is most likely due to different terminal sizes or other hacky printf stuff I missed :) Any ideas?

image

orpheus-antpool:
> > the figure is clipped. Can you please elaborate?
> 
> Right, Updated image in previous comment

Should be fixed now. Somehow, I ended up pushing an intermediate version and was wondering why it works on my end but not yours :)

rdlrt:
Just feel it to be still a bit cramped and reduced margin (for instance, the memory usage goes into double digits we end up either having to resize the view), feels like if this has to be included, the max width might need a bump too - thoughts @Scitz0 ?

Scitz0:
I agree with the comments from @rdlrt, to add more metrics like this the UI would need an overhaul. My suggestion would be to play around with the width and increase this a bit. Then maybe make it a three-column view instead of two. You will find these values in the upper part of the script:
```
width=63
second_col=$((width/2 + 3))
```
These are then used to later align the content dynamically. Here more variables could be added. Above could be kept(with modified width) if you need a two-column split. New variables could be added to create a three-column view etc. I dont have the solution but just some suggestions for how to re-design the interface. The goal should be to minimize dead space, yet still make it clear without obscure abbreviations and keep section metrics together that belong.

orpheus-antpool:
Agreed. I added a third column. Also, I we could use tput ${line} ${column} to fill the reminder of the column instead of printf "%-s$(($width-$column-X))". This way we avoid having to count X :)

scripts/cnode-helper-scripts/gLiveView.sh
```diff
@@ -179,8 +179,9 @@ declare -gA geoIP=()
 [[ -z ${PEER_LIST_CNT} ]] && PEER_LIST_CNT=6
 
 # Style
-width=63
-second_col=$((width/2 + 3))
+width=67
+second_col=$((width/3 + 12))
+third_col=$((second_col + 16))
```
Scitz0, line 184: This wasn't exactly what I had in mind. 
I want to be able to have both a two and three-column view.
So something like this 
* start position of first col not relevant
* +3 is used due to the fact that three chars go away at the start of the line
```
width=67
two_col_second=$((width/2 + 3))
three_col_second=$((width/3 + 3))
three_col_third=$(((width/3)*2 + 3))
```
with these you can build both a two-column view and a three-column view.

Scitz0 (commented):
Also see separate notes added.

Scitz0:
> Also, I we could use tput ${line} ${column} to fill the reminder of the column instead of printf "%-s$(($width-$column-X))". This way we avoid having to count X :)

**tput vs printf**
All external(non BASH internal) calls are much more expensive compared to if you can use internal BASH commands. To demonstrate I made a loop that jumps to line 30, prints something, moves back to column 0 and prints something. I set the loop to 10000 iterations. Using printf it took 0.1-0.2s to complete. Using tput, it took over 12 seconds. We dont call tput that many times in each loop but everything done at each refresh adds up. I know I have used tput a lot in the code already. But I have plans to move away from all repetitive external calls as much as possible. Just haven't had the time for it. 

**layout**
My thought was to rearrange it something like this. Switch to three-column view shown as yellow and keep two column view painted in orange. Might be more/better ways to optimize space but this was my initial thought without thinking that much about the grouping and arrangement. 

glv

orpheus-antpool:
I see your point. One idea is to make the third column in the first section print memory stats (RSS, Live, Heap). Then, reduce the width of the first column in the second section for Processed TX and Mempool TX (we could rename to Total TX and drop bytes from Mempool, which are not really relevant anyway). Finally, instead of Live/Heap print info about the garbage collector (minor/major). The rest remains unchanged and uses two columns (two_col_second) as per your suggestion. How does this sound?

Scitz0:
I will test your latest changes tomorrow and see how it looks. I think I will also create a collection of helper functions to be able to move around and to mimic the functionality of tput with printf for maximum performance.

Scitz0:
Something to also draw inspiration from could be this alternative modified compact version of gLiveView. 
The reduced line above epoch progress is good I think and if we make the three-column view 4-5 rows we could accommodate all metrics in a similar fashion. With the added width some of the abbreviations could be made a bit clearer. 

http://lbcrypt.com/cardano/gAltView.sh

photo_2021-09-25_00-43-25

orpheus-antpool:
@Scitz0 looks good to me. Looking forward to your commit.

Scitz0:
Thank you for your PR, I'm closing this PR in favour of #1115 that I just pushed. Please test this branch and let me know what you think. Leave comments in the new PR. Easier to move into new PR as my changes would have caused conflicts with your PR. 

A combination of your PR, http://lbcrypt.com/cardano/gAltView.sh and my ideas.

PR #1115, opened 2021-09-25
gLiveView 1.22.0 & CNTools 8.6.6

Scitz0:
**gLiveView 1.22.0**
- UI redesigned to be a bit more compact and include GC / Memory metrics. Credits to @orpheus-antpool for his contributions in PR #1112 that lay the ground for this. Inspiration was also taken from http://lbcrypt.com/cardano/gAltView.sh. Now utilizes a three-column upper view and other sections made slightly more compact. 
- To improve performance and reduce as many external (non bash internal) calls as possible tput was replaced with printf ANSI escape sequences cursor movement helper functions.
- Documentation update to reflect changes in metrics.

**CNTools 8.6.6**
- Pool rotation showed the wrong date as new expiration. The function still working as it should. 

TODO
- [x] Look over documentation and make sure the GC / Memory metrics are properly explained. 
- ~~Move cursor movement helper functions to env **?**~~ 
Some movement are very specific to gLV. Move later if needed elsewhere, tput is fine for most cases if performance is not critical.

### DIFF
--- a/docs/Scripts/gliveview.md
+++ b/docs/Scripts/gliveview.md
@@ -44,17 +44,29 @@ A sample output from both core and relay (with peer analysis):
 
 ###### Upper main section
 
-Displays live metrics gathered from EKG. Epoch number and progress is live from the node while date calculation until epoch boundary is based on offline genesis parameters. Reference tip is also an offline calculation based on genesis values used to compare against the node tip to see how far of the tip (diff value) the node is. With current parameters a slot diff up to 40 from reference tip is considered good but it should usually stay below 30. In/Out peers show how many connections the node has established in and out. Live/Heap shows how much memory is used for live/heap data. A large difference between them (or the heap approaching the physical memory limit) means the node is struggling with the garbage collector and/or may begin swapping.
+Displays live metrics from cardano-node gathered through the nodes EKG/Prometheus(env setting) endpoint.
+- **Epoch Progress** - Epoch number and progress is live from the node while date calculation until epoch boundary is based on offline genesis parameters.
+- **Block** - The nodes current block height since genesis start.
+- **Slot** - The nodes current slot height since current epoch start.
+- **Density** - With the current chain parameters(MainNet), a block is created roughly every 20 seconds(`activeSlotsCoeff`). A slot on MainNet happens every 1 second(`slotLength`), thus the max chain density can be calculated as `slotLength/activeSlotsCoeff = 5%`. Normally, the value should fluctuate around this value.
+- **Total Tx** - The total number of transactions processed since node start.
+- **Pending Tx** - The number of transactions and the bytes(total, in kb) currently in mempool to be included in upcoming blocks.
+- **Tip (ref)** - Reference tip is an offline calculation based on genesis values for current slot height since genesis start.
+- **Tip (node)** - The nodes current slot height since genesis start.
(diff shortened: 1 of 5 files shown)


scripts/cnode-helper-scripts/gLiveView.sh
```diff
@@ -952,58 +1058,102 @@ while true; do
...
+        
+        mvThreeFirstBlock
+        printf "Leader : ${leader_fmt}%-${col_2_value_width}s${NC}" "${isleader}"
+        mvThreeSecondBlock
+        printf "Adopted : ${adopted_fmt}%-${col_3_value_width}s${NC}" "${adopted}"
+        mvThreeThirdBlock
+        printf "Invalid : ${missed_fmt}%-${col_4_value_width}s${NC}" "${didntadopt}"
+        closeRow
+        
+        printf "${VL}${STANDOUT} BLOCKS ${NC} %$((width-38))s %-6s | ${FG_GREEN}%-7s${NC} | ${FG_RED}%-7s" "" "Leader" "Adopted" "Invalid" && closeRow
```
rdlrt, line 1131: isnt this a repeat of above?

docs/Scripts/gliveview.md
```diff
@@ -44,7 +44,7 @@ A sample output from both core and relay (with peer analysis):
...
+Displays live metrics gathered from EKG. Epoch number and progress is live from the node while date calculation until epoch boundary is based on offline genesis parameters. Reference tip is also an offline calculation based on genesis values used to compare against the node tip to see how far of the tip (diff value) the node is. With current parameters a slot diff up to 40 from reference tip is considered good but it should usually stay below 30. In/Out peers show how many connections the node has established in and out. Live/Heap shows how much memory is used for live/heap data. A large difference between them (or the heap approaching the physical memory limit) means the node is struggling with the garbage collector and/or may begin swapping.
```
rdlrt, line 47: Should we split this out to bullet points instead?

scripts/cnode-helper-scripts/gLiveView.sh
Scitz0: Correct, fixed in 0b9557c

docs/Scripts/gliveview.md
Scitz0: A good suggestion. Pretty unreadable to have it all in a huge block of text. Best to split into sections and bullet points for each metric.

rdlrt (approved):

orpheus-antpool:
Hi, this looks good so far. Two issues though:
1. Tip (diff) column will overflow if the node is synching
2. EGK metrics for blocks are not displayed properly, see image below

image

Scitz0:
> Hi, this looks good so far. Two issues though:
> 
> 1. Tip (diff) column will overflow if the node is synching
> 2. EGK metrics for blocks are not displayed properly, see image below
> 
> image

Test latest version pushed to alpha, should be fixed there.

orpheus-antpool:
Looks good now. I would use greed/red colors for Adopted/Invalid though, same as before. Also, the number of blocks could be styled in blue color too, to match the rest of the stats.

Scitz0:
Pushing a change that will make value colored depending on status. Adopted/Confirmed will show green if matching Leader, else yellow. Invalid/Missed/Ghosted/Stolen will show blue if zero, else red. I don't think I want to color the header. 

I couldn't find the misspelling you mentioned. Where is this?

orpheus-antpool:
The spelling is a symptom of an artifact, because closeRow will jump and the end of the column instead of printing empty spaces. This is not a problem if the new text is longer than the previous one, but becomes the problem the other way around, because you are not "erasing" the extra characters from the previous text. In the case of "remaininging", the transition from 1 day XX hours to XX hours probably made the new text shorter, hence the artifact.

Scitz0:
Ok thanks, yes I see what you mean. Fixed in 1.22.2 pushed to alpha.

rdlrt:
+1, Thanks @orpheus-antpool for your inputs

PR #1118, opened 2021-09-28
Add CNODEBIN to allow custom cardano-node name; Closes #1071

### LINKED ISSUE #1071: Wrong version detected in case of multiple cardano-node executables
**Describe the bug**
the gLiveView.sh script is detecting a wrong version of cardano-node in case of the running node process is executed using a binary named differently than cardano-node.
In case of multiple version installation, like `cardano-node-X.Y.Z` and `cardano-node` the gLiveView.sh script is not able to detect the correct version assuming that there is an executable named cardano-node.

**To Reproduce**
1. Install cardano-node version 1.27.0 binary in PATH , naming it `cardano-node`
2. Install cardano-node version 1.29.0 binary in PATH , naming it `cardano-node-1.29.0`
3. Execute the `cardano-node-1.29.0`

**Additional context**
I found useful to maintain installed several version of cardano-node and cardano-cli installed on my nodes, an during some testing upgrades or other migrations I don't want to have to stick to a binary named `cardano-node`

The issue is caused by this lines:
- https://github.com/cardano-community/guild-operators/blob/alpha/scripts/cnode-helper-scripts/gLiveView.sh#L459
- https://github.com/cardano-community/guild-operators/blob/alpha/scripts/cnode-helper-scripts/gLiveView.sh#L516

Would be good if using the CNODE_PID variable (detected) the scripts would find the exact executable for the running process using something like:

```
cat /proc/${CNODE_PID}/cmdline |tr '\0' ' ' | cut -f1 -d' ' 
```

rdlrt:


scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -522,12 +522,7 @@ cncliMigrateBlocklog() {
 cncliPTsendtip() {
   [[ ${NWMAGIC} -ne 764824073 ]] && echo "PoolTool sendtip only available on MainNet, exiting!" && exit 1
   [[ -z ${POOL_ID} || -z ${POOL_TICKER} || -z ${PT_API_KEY} ]] && echo "'POOL_ID' and/or 'POOL_TICKER' and/or 'PT_API_KEY' not set in $(basename "$0"), exiting!" && exit 1
-  
-  # Generate a temporary pooltool config
-  if ! cnode_path=$(command -v cardano-node 2>/dev/null); then
-    echo "ERROR: cardano-node not in PATH, please manually set CCLI in env file"
-    exit 1
-  fi
+  cnode_path=$(command -v ${CNODEBIN} 2>/dev/null)
```
rdlrt, line 525: Ensuring CNODEBIN is in path is in env already. Hence, removed condition

scripts/cnode-helper-scripts/env
```diff
@@ -106,19 +107,26 @@ telegramSend() {
...
 if [[ -z "${CCLI}" ]]; then
   if [[ -f "${HOME}/.cabal/bin/cardano-cli" ]]; then
-    # Assumption being made that cardano-cli and cardano-node are both present, if not - prereqs and build instructions were not followed
     export PATH="${HOME}/.cabal/bin":$PATH
     CCLI=$(command -v cardano-cli)
   else
     echo "You do not have a cardano-cli binary available in \$PATH."
     return 1
   fi
-else
```
rdlrt, line 119: We check `command -v` execution above, if not empty - the CCLI_PARENT is already present in ${PATH} . Hence, removed the else condition

scripts/cnode-helper-scripts/cncli.sh
Scitz0: Cant we juse use ${CNODEBIN} directly in this case.
`CNODEBIN=$(command -v cardano-node)`
is already done in env.

Scitz0 (changes requested):
In addition to current changes we have references to `[c]ardano-node` in more places. 

**cnode.sh**
`if pgrep -f "[c]ardano-node.*.${CARDANO_NODE_SOCKET_PATH}"; then`

**sLiveView.sh**
Multiple references to cardano-node in various ways. 

**deploy-as-systemd.sh**
`ExecStop=/bin/bash -l -c \"exec kill -2 \$(ps -ef | grep [c]ardano-node.*.${CNODE_HOME}/ | tr -s ' ' | cut -d ' ' -f2) &>/dev/null\"`

**env**
both SOCKET and CONFIG reference static `[c]ardano-node`

**gLiveView.sh**
`CNODE_PID=$(pgrep -fn "[c]ardano-node*.*--port ${CNODE_PORT}")`

rdlrt:
`sLiveView` atm sits kinda independent of entire script repository, as it does not inherit the other functions in consistent manner. Thus, did not touch sLiveView, but good check about others 😄

Scitz0:
> `sLiveView` atm sits kinda independent of entire script repository, as it does not inherit the other functions in consistent manner. Thus, did not touch sLiveView, but good check about others 😄

True 👍

scripts/cnode-helper-scripts/env
```diff
@@ -250,7 +258,7 @@ BLOCKLOG_DB="${BLOCKLOG_DIR}/blocklog.db"
 [[ -z "${IP_VERSION}" ]] && IP_VERSION=4
 IP_VERSION=$(tr '[:upper:]' '[:lower:]' <<< "${IP_VERSION}")
 [[ -z ${USE_EKG} ]] && USE_EKG='Y'
-CNODE_PID=$(pgrep -fn "[c]ardano-node.*.--port ${CNODE_PORT}")
+CNODE_PID=$(pgrep -fn "$(basename ${CNODEBIN}).*.--port ${CNODE_PORT}")
```
Scitz0, line 261: In all other places ${CNODEBIN} is used directly, any reason to do otherwise here?

rdlrt: Hmm I actually wonder, should we use basename at all places instead?
Reason being that will allow to not assume that the execution command contains full path for cardano-node.

Scitz0: makes sense 👍

Scitz0 (approved):

PR #1120, opened 2021-09-29
Env + docker entrypoint fixes

rdlrt:
- env: Fix quotation for ps commands
- docker entrypoint: Fix path cntools.config, and disable dialog by default


Scitz0 (approved):
👍

PR #1122, opened 2021-09-29
gLiveView 1.22.3

Scitz0:
Latency check settings added. Its now possible to set the tools to use in latency check. Both what tools to use and in what order. In addition to this for cncli you can specify if full handshake duration should be measured or if only connect time should be showed.

-p flag removed due to above change.


gufmar (approved):
Lgtm

redoracle (approved):
Looks great to me. tnx

rdlrt (approved):

PR #1129, opened 2021-10-02
Add missing user variables in scripts on update check

Scitz0:
Implements feature mentioned in discussion #1126

### DIFF
--- a/scripts/cnode-helper-scripts/env
+++ b/scripts/cnode-helper-scripts/env
@@ -392,13 +392,15 @@ checkUpdate() {
       printf "${FG_RED}ERROR${NC}: '${1}' update failed! Please use prereqs.sh to force an update or manually download the file from GitHub"
       return 2
     fi
-    TEMPL_CMD=$(awk '/^# Do NOT modify/,0' "${PARENT}/${1}")
-    TEMPL2_CMD=$(awk '/^# Do NOT modify/,0' "${PARENT}/${1}".tmp)
-    if [[ "$(echo ${TEMPL_CMD} | sha256sum)" != "$(echo ${TEMPL2_CMD} | sha256sum)" ]]; then
-      if [[ $2 = Y ]] || { [[ -t 1 ]] && getAnswer "\nThe static content from ${FG_YELLOW}${1}${NC} does not match with guild-operators repository, do you want to download the updated file?"; }; then
+    OLD_STATIC=$(awk '/#!/{x=1}/^# Do NOT modify/{exit} x' "${PARENT}/${1}")
+    OLD_TEMPL=$(awk '/^# Do NOT modify/,0' "${PARENT}/${1}")
+    GIT_STATIC=$(awk '/#!/{x=1}/^# Do NOT modify/{exit} x' "${PARENT}/${1}".tmp)
+    GIT_TEMPL=$(awk '/^# Do NOT modify/,0' "${PARENT}/${1}".tmp)
+    NEW_STATIC=$(checkUserVariables "${OLD_STATIC}" "${GIT_STATIC}")
+    if [[ "$(echo ${OLD_STATIC} | sha256sum)" != "$(echo ${NEW_STATIC} | sha256sum)" || "$(echo ${OLD_TEMPL} | sha256sum)" != "$(echo ${GIT_TEMPL} | sha256sum)" ]]; then
+      if [[ $2 = Y ]] || { [[ -t 1 ]] && getAnswer "\nThe content from ${FG_YELLOW}${1}${NC} does not match with guild-operators repository, do you want to download the updated file?"; }; then
         cp "${PARENT}/${1}" "${PARENT}/${1}_bkp$(date +%s)"
-        STATIC_CMD=$(awk '/#!/{x=1}/^# Do NOT modify/{exit} x' "${PARENT}/${1}")
-        printf '%s\n%s\n' "$STATIC_CMD" "$TEMPL2_CMD" > "${PARENT}/${1}".tmp
+        printf '%s\n%s\n' "${NEW_STATIC}" "${GIT_TEMPL}" > "${PARENT}/${1}".tmp
         mv "${PARENT}/${1}".tmp "${PARENT}/${1}"
         [[ ! "${1}" == "env" ]] && chmod +x "${1}"
         echo -e "\n${FG_YELLOW}${1}${NC} update successfully applied!"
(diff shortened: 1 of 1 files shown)


scripts/cnode-helper-scripts/env
```diff
@@ -411,6 +413,26 @@ checkUpdate() {
...
 
+# Description : Check that old script User Variables section contain all current settings from GitHub.
+#             : Prints final result to stdout
+#             : $1 = Old/Current script User Variables
+#             : $2 = User Variables section from GitHub
+checkUserVariables() {
+  tmp_head=$(head -n -2 <<< "$1")
+  tmp_tail=$(tail -n -2 <<< "$1")
+  while IFS= read -r line; do
+    if [[ ${line} =~ ^\#?([[:alnum:]_]+\=) ]]; then
```
rdlrt, line 424: Should we add space char too? often users when commenting a variable could.do `# varname=..`
to avoid duplicates

scripts/cnode-helper-scripts/env
```diff
@@ -392,13 +392,15 @@ checkUpdate() {
...
-    TEMPL_CMD=$(awk '/^# Do NOT modify/,0' "${PARENT}/${1}")
-    TEMPL2_CMD=$(awk '/^# Do NOT modify/,0' "${PARENT}/${1}".tmp)
-    if [[ "$(echo ${TEMPL_CMD} | sha256sum)" != "$(echo ${TEMPL2_CMD} | sha256sum)" ]]; then
-      if [[ $2 = Y ]] || { [[ -t 1 ]] && getAnswer "\nThe static content from ${FG_YELLOW}${1}${NC} does not match with guild-operators repository, do you want to download the updated file?"; }; then
+    OLD_STATIC=$(awk '/#!/{x=1}/^# Do NOT modify/{exit} x' "${PARENT}/${1}")
+    OLD_TEMPL=$(awk '/^# Do NOT modify/,0' "${PARENT}/${1}")
+    GIT_STATIC=$(awk '/#!/{x=1}/^# Do NOT modify/{exit} x' "${PARENT}/${1}".tmp)
+    GIT_TEMPL=$(awk '/^# Do NOT modify/,0' "${PARENT}/${1}".tmp)
+    NEW_STATIC=$(checkUserVariables "${OLD_STATIC}" "${GIT_STATIC}")
+    if [[ "$(echo ${OLD_STATIC} | sha256sum)" != "$(echo ${NEW_STATIC} | sha256sum)" || "$(echo ${OLD_TEMPL} | sha256sum)" != "$(echo ${GIT_TEMPL} | sha256sum)" ]]; then
```
rdlrt, line 400: Thinking from UX, this could cause twice prompts:
1. To see that there's an update (when this env gets merged to alpha). User pressing yes will quit their gLV
2. The next attempt will again show an update (this time coz NEW_STATIC is different)
Having said that, I cant think of a way to avoid that from happening and not have quiet update, so I believe the one time double prompt could be managed via comms

scripts/cnode-helper-scripts/env
Scitz0: 👍 improved regex to match both on multiple # and spaces before the variable

scripts/cnode-helper-scripts/env
Scitz0: 1. I have thought about this before as well when there is an update to both env and gLV. First, it updates env, then quits, then updates gLV and quits. Not the best UX. After the update of the script, before quitting, could we re-launch the script with `exec "<path to self>"`? To be done in the calling script as env just return with exit code 1. 
2. True, for the first time switching from old env to this it will give notification twice about the update. But as you said, not sure I see a way around this either.

rdlrt: Adding exec could be a bit tricky as it will return control back to an outdated script that will be handled differently on different OS

Can just be a reminder for ourselves to communicate it when merged to master branch

rdlrt (approved):

Scitz0:
Should we move cntools.config into library in this PR as well?

The only problem I see is if people have made changes to the current config. How should we deal with this? Create a special function that looks in config if it exists, copies the values, and then removes the file?

rdlrt:
> The only problem I see is if people have made changes to the current config. How should we deal with this? Create a special function that looks in config if it exists, copies the values, and then removes the file?

Shouldnt it be cntools.sh (for user variables and make cntools.library really a file that users shouldnt have to touch at all)?
1. If cntools.config exists, source it first (with a comment about cntools.config being deleted in future major versions)...
2. Next cntools user variables (commented by default, if anyone changes any of these, it will override cntools.config)
3. If environment variables are not specified, set defaults

Scitz0:
> Shouldnt it be cntools.sh (for user variables and make cntools.library really a file that users shouldnt have to touch at all)?
> 
> 1. If cntools.config exists, source it first (with a comment about cntools.config being deleted in future major versions)...
> 2. Next cntools user variables (commented by default, if anyone changes any of these, it will override cntools.config)
> 3. If environment variables are not specified, set defaults

👍  sounds good. Will make this change.

Scitz0:
Please test the latest CNTools configuration change so that it works as expected.

rdlrt:
while the change is good, since cntools.sh current version may not have a DO NOT MODIFY line..do you think there is a way to no have this as a breaking change for CNTools?

rdlrt:
Kinda edge-case controlled by - fault on github - for example now since versions are not bumped for gLV, but in terms of UX , might be something we would want to avoid?

image

For an end user, it might show as update prompt 1 for env, prompt 2 for gLV (1.23 to 1.24 for example) - and prompt 3 for1.24 to 1.24 if there are certain user vars missing

Scitz0:
> Kinda edge-case controlled by - fault on github - for example now since versions are not bumped for gLV, but in terms of UX , might be something we would want to avoid?

Yip, should atleast change the message. The default message should be fine I think(?):
`The content from ${FG_YELLOW}${1}${NC} does not match with guild-operators repository, do you want to download the updated file?`


> For an end user, it might show as update prompt 1 for env, prompt 2 for gLV (1.23 to 1.24 for example) - and prompt 3 for1.24 to 1.24 if there are certain user vars missing

If both env and gLV has changes to it, it will trigger two separate update messages for this yes. Before this PR, it exited the script after env update, and then prompted for gLV update and exited a second time. Now it at least continues after env update notice and only exit after gLV update, if there are any. 

Is it preferred not to have separate update messaged for env and called script?

There shouldn't be three update prompts, if user vars are missing this would be handled by gLV update in step 2 in your example. At least I haven't been able to trigger this scenario. If you have a way please let me know.

Scitz0:
Made some additional enhancements. In gLV however, missing user variables are added just before `Do NOT modify code below` section. `Themes` section comes in between. Not sure how to make this better in a good way or if this is just something we have to live with. Even worse is if stuff is missing in the themes section, then it will be added outside of the `setTheme` function. It won't break anything and it's mostly static anyway but something to keep in mind. Changes to theme variables are a breaking change.

PR #1130, opened 2021-10-03
Improve active_stake_cache setup

DamjanOstrelic:
Switching to updating once for each row instead of summing up all the values on each insert. Tested on a mock epoch transition by restarting db-sync (rollback to epoch start).


Scitz0 (approved):
I haven't had time to test it out but quickly going over code changes it looks alright. To me, we can merge and if we find that it's not working as expected later on we can always fix it.

PR #1132, opened 2021-10-04
Add address_txs RPC

DamjanOstrelic:
Currently gets the tx hash list and has an optional `after_block_height` filter.

But we talked about the possibility of bulk addresses received to this endpoint - I need clarification on that.

If we receive bulk addresses, how is that useful for the user if we are only listing tx hashes? Do we need to return a structure of:

```
[
  address1: { tx_hash1, tx_hash2 },
  address2: { tx_hash1, tx_hash2 }
]
```
in that case?

### DIFF
--- a/docs/Build/grestsvcs.json
+++ b/docs/Build/grestsvcs.json
@@ -670,10 +670,10 @@
           {
             "required": true,
             "schema": {
-              "required": ["_payment_address_arr"],
+              "required": ["_payment_addresses"],
               "type": "object",
               "properties": {
-                "_payment_address_arr": {
+                "_payment_addresses": {
                   "type": "array",
                   "items": { "type": "string" },
                   "minItems": 1
--- a/docs/Build/grestsvcs.json
+++ b/docs/Build/grestsvcs.json
@@ -644,9 +644,39 @@
               "required": ["_payment_address_bech32"],
               "type": "object",
               "properties": {
-                "_address": {
+                "_payment_address_bech32": {
                   "format": "text",
                   "type": "string"
+                }
+              }
+            },
+            "in": "body",
+            "name": "args"
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "OK"
+          }
+        }
+      }
+    },
+    "/rpc/address_txs": {
+      "post": {
+        "tags": ["Address Queries"],
+        "summary": "Get the transaction hash list of a payment address array, optionally filtering after specified block height (inclusive)",
+        "produces": ["application/json", "application/vnd.pgrst.object+json"],
+        "parameters": [
+          {
+            "required": true,
+            "schema": {
+              "required": ["_payment_address_arr"],
+              "type": "object",
+              "properties": {
+                "_payment_address_arr": {
+                  "type": "array",
+                  "items": { "type": "string" },
+                  "minItems": 1
                 }
               }
             },
--- a/files/grest/rpc/address/address_txs.sql
+++ b/files/grest/rpc/address/address_txs.sql
@@ -1,6 +1,6 @@
 DROP FUNCTION IF EXISTS grest.address_txs (text[], integer);
 
-CREATE FUNCTION grest.address_txs (_payment_address_arr text[], _after_block_height integer DEFAULT NULL)
+CREATE FUNCTION grest.address_txs (_payment_addresses text[], _after_block_height integer DEFAULT NULL)
   RETURNS TABLE (
     tx_hash text)
   LANGUAGE PLPGSQL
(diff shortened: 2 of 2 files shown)


Scitz0:
I don't think the address is important in this case. The point is to get all transactions that provided address(es) has a connection to. These can then used on tx_info endpoint to get the details for all returned tx hashes. End-user can then match the address with inputs/outputs of tx as he wants.

DamjanOstrelic:
> I don't think the address is important in this case. The point is to get all transactions that provided address(es) has a connection to. These can then used on tx_info endpoint to get the details for all returned tx hashes. End-user can then match the address with inputs/outputs of tx as he wants.

Okay, in that case, it's simpler and I'll just make it accept a list of addresses.

DamjanOstrelic:
Updated with bulk address input support. Now, the endpoint MUST receive an array of addresses (can still be just 1 or even 0). So some example requests would be:
```
-- without after block filtering:
curl -X POST http://127.0.0.1:8050/rpc/address_txs \
-H 'Content-Type: application/json' \
-d '{
"_payment_address_arr": ["addr1qxqzp46gah6tud2lpqswexleyf78mehpp2r6pl9sr34m2zyz4jgk36t20j5hw2a30r7gv0yfvy3vxt7hv9mt9fn8xxkqh708ww", "addr1q8ytda2q4hdjfsf5frkgpp0knwpnhwdndlcp65an9mjw8u0wjefzrj8h5ezugv59k5fq6rca9mnlqczshny3dpkwdn5q5c0yv4"]
}' \
2>/dev/null | jq .
[
  {
    "tx_hash": "15cc6a1b58309ccd4f762976a1ec6aa0f914c936a87c0e782656a5208c3c0b9f"
  },
  {
    "tx_hash": "28b5527f5cad3caab994fda2bff51c2afffac2ed8b8ed442fbe7cf64bf6e6856"
  },
  {
    "tx_hash": "e01af28b007796b63ed3be4cfd128f7ee309583a3204c1ef95c471c4a4c683ed"
  }
]
```
```
-- with after block filtering:
curl -X POST http://127.0.0.1:8050/rpc/address_txs \
-H 'Content-Type: application/json' \
-d '{
"_payment_address_arr": ["addr1qxqzp46gah6tud2lpqswexleyf78mehpp2r6pl9sr34m2zyz4jgk36t20j5hw2a30r7gv0yfvy3vxt7hv9mt9fn8xxkqh708ww", "addr1q8ytda2q4hdjfsf5frkgpp0knwpnhwdndlcp65an9mjw8u0wjefzrj8h5ezugv59k5fq6rca9mnlqczshny3dpkwdn5q5c0yv4"],
"_after_block_height": 6331086
}' \
2>/dev/null | jq .
[
  {
    "tx_hash": "28b5527f5cad3caab994fda2bff51c2afffac2ed8b8ed442fbe7cf64bf6e6856"
  },
  {
    "tx_hash": "e01af28b007796b63ed3be4cfd128f7ee309583a3204c1ef95c471c4a4c683ed"
  }
]
```

```
-- with one address:
curl -X POST http://127.0.0.1:8050/rpc/address_txs \
-H 'Content-Type: application/json' \
-d '{
"_payment_address_arr": ["addr1qxqzp46gah6tud2lpqswexleyf78mehpp2r6pl9sr34m2zyz4jgk36t20j5hw2a30r7gv0yfvy3vxt7hv9mt9fn8xxkqh708ww"]
}' \
2>/dev/null | jq .
[
  {
    "tx_hash": "15cc6a1b58309ccd4f762976a1ec6aa0f914c936a87c0e782656a5208c3c0b9f"
  },
  {
    "tx_hash": "28b5527f5cad3caab994fda2bff51c2afffac2ed8b8ed442fbe7cf64bf6e6856"
  }
]
```
```
-- with one address and block filter:
curl -X POST http://127.0.0.1:8050/rpc/address_txs \
-H 'Content-Type: application/json' \
-d '{
"_payment_address_arr": ["addr1qxqzp46gah6tud2lpqswexleyf78mehpp2r6pl9sr34m2zyz4jgk36t20j5hw2a30r7gv0yfvy3vxt7hv9mt9fn8xxkqh708ww"],
"_after_block_height": 6331086
}' \
2>/dev/null | jq .
[
  {
    "tx_hash": "28b5527f5cad3caab994fda2bff51c2afffac2ed8b8ed442fbe7cf64bf6e6856"
  }
]
```

The not so great thing about it is that you have to send an array of addresses even if it's just a single address you're interested in. This can be avoided by defining the function using `VARIADIC` argument for the payment address array, but `VARIADIC` arguments must come last in the function definition, meaning that the `_after_block_height` would have to come first, and would have to be specified on every request even when filtering is not wanted.
So I am in favour of having it the way it is now, that the user must send an array of values, but can leave out the optional filter.

Scitz0:
> The not so great thing about it is that you have to send an array of addresses even if it's just a single address you're interested in. This can be avoided by defining the function using `VARIADIC` argument for the payment address array, but `VARIADIC` arguments must come last in the function definition, meaning that the `_after_block_height` would have to come first, and would have to be specified on every request even when filtering is not wanted.
> So I am in favour of having it the way it is now, that the user must send an array of values, but can leave out the optional filter.

I think this is fine as well, even for a single address.

rdlrt:
agree

rdlrt (approved):

rdlrt:
A small thing, can we use `_payment_addresses` instead of `_payment_address_arr` ?

PR #1141, opened 2021-10-12
Updates for setup process, and move dbsync steps to dbsync.sh

rdlrt:
- [x] - Split dbsync.sh tasks to be independent of Koios
- [x] - Add dbsync.sh to be part of prereqs.sh
- [x] - Update dbsync documentation page
- [x] - Remove sections for dbsync from setup-grest.sh
- [x] - Use top level folder from already pre-filled env instead (also removes having to duplicate code from prereqs)
- [x] - Make Postgres DB name variable
- [x] - Make HAProxy config file variable (and change default to `$CNODE_HOME/files`)
- [x] - [Re]Move common functions currently in setup-grest that (as functions should be in `env`)
- [x] - Parameterize install steps for HAProxy, Polling, PostgREST, etc
- [x] - Add parameter to [re]-install systemd services for Koios Suite
- [x] - Squash bugs introduced due to changes
    - [x] Install independent components
    - [x] Overwrite configs/systemd
    - [x] Move views to RPC folder
    - [x] Download and run queries
- [x] - Reduce calls to api.github.com (supress description in setup output)
- [x] - Add setup-grest is added to prereqs (to keep CNODE_VNAME substituted similar to other scripts)
- [x] - Update documentations
- [x] - Test setup scenarios (there are breaking changes - testing would be assuming fresh run of setup-grest)
    - [x] Fresh Install of components
    - [x] Force overwrite of existing components
    - [x] Fresh fetch of config/systemd
    - [x] Update-in-place of config/systemd
    - [x] Fresh install of queries
    - [x] Update-in-place to overwrite queries

Excluded for this PR: verifying flows for dependency between components if they're plugged out

### DIFF
--- a/docs/Build/pgrest.md
+++ b/docs/Build/pgrest.md
@@ -5,7 +5,7 @@
 
 [PostgREST](https://postgrest.org/en/latest) is a web server that serves any PostgreSQL database (in our case, useful for `cardano-db-sync`) as a RESTful Web Service. The endpoints of PostgREST in itself are essentially the table/functions exposed via the PostgREST config file. You can read more about exploring the API [here](https://postgrest.org/en/latest/api.html). Understandably, it would of course rely on an existing PostgreSQL server. It is an easy alternative - which will remain up-to-date since it is directly serving the underlying database as an API, as compared to `Cardano GraphQL` component. Some of the other advantages are also performance, elasticity, low overhead, support for JWT / native Postgres DB authentication against the Rest Interface as well.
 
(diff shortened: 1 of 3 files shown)


rdlrt:
During testing `Download and run RPC/cache table queries`, realised that unauthenticated requests to github API is capped at 60 per minute. Since we loop over for each directory to get a list, and then download each sub directories file-list, running it a few times can quickly exhaust the limit. Hence, we'd want to switch to clone instead (in case of which we need to be mindful of content to bring and performance impact, might use submodules instead)

PR #1149, opened 2021-10-19
Small cleanups for setup-grest.sh and err_exit()

dostrelith678:
~Would this work for checking the tip and deciding whether we should or shouldn't deploy RPCs/cache/etc?
I considered the case where db-sync is not running and protected against that, not sure if there are other edge cases that would cause this not to work properly.~ Will keep the Mary era check.

Also, I added a new line to the `err_exit` function so we don't have to worry about it when passing the error message. And cleaned up the `ERROR!!` usage across the repo since the function prints `ERROR: ` itself already.

Does having an empty line in-between functions break the collapse utility in editors? I don't know how to check since mine doesn't do it 😆

### DIFF
--- a/scripts/cnode-helper-scripts/prereqs.sh
+++ b/scripts/cnode-helper-scripts/prereqs.sh
@@ -49,7 +49,6 @@ get_answer() {
 #             : $1 = Error message we'd like to display before exiting (function will pre-fix 'ERROR: ' to the argument)
 err_exit() {
   printf "\e[31mERROR\e[0m: ${1}\n" >&2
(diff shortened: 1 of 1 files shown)


rdlrt:
Added some discussion points on the Koios discussion group.

Current behaviour: Check if proposal for protocol major version 5 was submitted (can only occur post shelley). This already ensures stake tables have populated.

Before we make this change - I'd like to first capture the reasons (eg: cache tables are not going to be valid reasons - as they'll only defer an issue with query to be post setup - eg: when restarting dbsync or when it does a rollback). Let's continue on the group and then add summary here

PR #1160, opened 2021-10-27
Pool history cache table and update function

hodlonaut:
Initial table creation and population, still need to discuss / decide on a trigger (currently thinking after rolling over into a new epoch)


DamjanOstrelic (approved):
👍

PR #1167, opened 2021-11-01
Used NULLIF to deal with possibly zero active stake amounts

hodlonaut:
Better to get null as a result of calculation than division by zero/overflow. Added update of control table with last timestamp of pool history cache update executing.


Scitz0 (approved):
LGTM

PR #1168, opened 2021-11-02
Set up pool_history_cache cron job and fixed control table update

hodlonaut:


scripts/grest-helper-scripts/setup-grest.sh
```diff
@@ -108,6 +108,9 @@
     fi
     get_cron_job_executable "stake-distribution-update"
     install_cron_job "stake-distribution-update" "*/30 * * * *"
+    
+    get_cron_job_executable "pool-history-cache-update"
+    install_cron_job "pool-history-cache-update" "*/10 * * * *"
```
DamjanOstrelic, line 113: Since we will be running the SQL every 10 mins, I would like to have a check in the SQL to make sure the previous run has been completed before starting a new one (epoch transitions is when time stops)

DamjanOstrelic (commented):
From a previous PR, but noticed it here at the start:

`delete from grest.control_table where key = 'pool_history_cache_last_updated';`

I don't think we should ever be deleting stuff from the `control_table`. The insert/update at the end is enough. Is there any reason to delete the entry at the start of the function?

scripts/grest-helper-scripts/setup-grest.sh
Scitz0: makes sense to add checks like this even if the query is fairly quick.

files/grest/rpc/cached_tables/pool_history_cache.sql
```diff
@@ -30,6 +28,16 @@ declare
   _curr_epoch bigint;
   _latest_epoch_no_in_cache bigint;
 begin
+   IF (
+    SELECT
+      COUNT(pid) > 1
+    FROM
+      pg_stat_activity
+    WHERE
+      query != '<IDLE>' AND query ILIKE '%GREST.pool_history_cache_update%') THEN
```
Scitz0, line 37: Can we change this line to:
`state = 'active' AND query ILIKE '%GREST.pool_history_cache_update%'`

Scitz0 (approved):
Great 👍

PR #1171, opened 2021-11-02
Docker update

redoracle:
few changes for enabling cncli and hw wallet support


DamjanOstrelic (approved):

redoracle:
cncli now is within the container and works, path need to be exported. 
I made a new PR to be reviewed to solve this issue + few optimiztions

PR #1175, opened 2021-11-04
tx_status endpoint (list of number of confirmations for given transac…

hodlonaut:
…tions)

### DIFF
--- a/files/grest/rpc/transactions/tx_status.sql
+++ b/files/grest/rpc/transactions/tx_status.sql
@@ -1,0 +1,40 @@
+DROP FUNCTION IF EXISTS grest.tx_status (_tx_hashes text[]);
+
+CREATE FUNCTION grest.tx_status (_tx_hashes text[])
+    RETURNS TABLE (
+        tx_hash text,
+        num_confirmations integer)
+    LANGUAGE plpgsql
+    AS $$
+DECLARE
+    _curr_block_no uinteger;
+BEGIN
+    SELECT
+        max(block_no) INTO _curr_block_no
+    FROM
+        block b;
+
+    RETURN QUERY (
+		WITH matched_txs AS (
+            SELECT
+                t.hash, b.block_no FROM block b, tx t
+            WHERE
+                t.hash::bytea = ANY (
+                    SELECT
+                        DECODE(HASHES, 'hex')
+                FROM UNNEST(_tx_hashes) AS hashes)
+        AND t.block_id = b.id)
+
+	-- returns difference in block numbers, or null if tx was not found
+	SELECT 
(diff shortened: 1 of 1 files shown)


DamjanOstrelic (commented):
I have some general questions:

Is it possible to store the decoded hashes into a variable?

e.g.
`SELECT DECODE(HASHES, 'hex') FROM UNNEST(_tx_hashes) AS HASHES  INTO _decoded_hashes`
Asking because unnesting/decoding is done twice (in the WITH, and return query).

What is sorting by ORDINALITY for?

hodlonaut:
Sample invocation/output against local endpoint:

curl -X POST http://127.0.0.1:8050/rpc/tx_status -H "Accept: application/json"  -H "Content-Type: application/json" -d '{"_tx_hashes":["f144a8264acf4bdfe2e1241170969c930d64ab6b0996a4a45237b623f1dd670e","0b8ba3bed976fa4913f19adc9f6dd9063138db5b4dd29cecde369456b5155e94","6cc8e9059407999899047e9a7b85c3b8bfa2aca49fddc3284368177b54e17757","767b342a0523c668ab39a02a0f1b8e44d600c4d1ecb157a59467442284cf205c","767b342a0523c668ab39a02a0f1b8e44d600c4d1ecb157a59467442284cf245c"]}'
[{"tx_hash":"f144a8264acf4bdfe2e1241170969c930d64ab6b0996a4a45237b623f1dd670e","num_confirmations":103626}, 
 {"tx_hash":"0b8ba3bed976fa4913f19adc9f6dd9063138db5b4dd29cecde369456b5155e94","num_confirmations":95461}, 
 {"tx_hash":"6cc8e9059407999899047e9a7b85c3b8bfa2aca49fddc3284368177b54e17757","num_confirmations":1795535}, 
 {"tx_hash":"767b342a0523c668ab39a02a0f1b8e44d600c4d1ecb157a59467442284cf205c","num_confirmations":5831454}, 
 {"tx_hash":"767b342a0523c668ab39a02a0f1b8e44d600c4d1ecb157a59467442284cf245c","num_confirmations":null}

The select query you mentioned is working fine in my DBeaver.

Ordinality is to ensure the data returned is in the same order as in the original array provided as input (otherwise I found it gets sorted by some other criteria by default).

hodlonaut:
I've simplified the query so it's doing a single unnset/decode, seems to be working fine. Let me know if you run into issues still testing it from your SQL UI/psql.

DamjanOstrelic (approved):
The updated version works for me 👍 

1 more consideration:

When NULL is returned as a number of confirmations, it means the tx was not found right? Would it be better to exclude those in the response altogether? I think using INNER JOIN instead of LEFT OUTER would produce those results.

PR #1186, opened 2021-11-09
Add param to reset the grest node

DamjanOstrelic:
`-r` for reset will:
- remove any grest-related cron entries
- drop all grest-related PostgreSQL triggers
- recreate great schema and set it up for `web_anon` user

It won't:
- recreate any tables/functions in the grest schema (use `-q` for that)

While the `-r` trigger was the goal of the PR, other things are included:
- switch setup SQL to files in the new `grest-helper-scripts/db-scripts/` dir (including genesis table setup)
- fixes to the checkUpdate function
- move control table creation to basic db setup
- separate some stuff into smaller functions


files/grest/cron/jobs/pool-history-cache-update.sh
```diff
@@ -1,3 +1,5 @@
 #!/bin/bash
+DB_NAME=cexplorer
```
DamjanOstrelic, line 2: I guess if we are going to be using DB_NAME in multiple places we should move and source it from `env`?

Scitz0: I see your point but not sure I want to pollute env with Koios specific variables.

DamjanOstrelic: Ah, I see you are taking care of custom db names in  `setup-grest.sh` with `sed` instead.

scripts/cnode-helper-scripts/env
```diff
@@ -141,6 +142,8 @@ fi
 [[ -z "${CNODE_HOME}" ]] && CNODE_HOME=/opt/cardano/cnode
 
 CNODE_NAME="$(basename ${CNODE_HOME})"
+CNODE_VNAME=$(tr '[:upper:]' '[:lower:]' <<< ${CNODE_NAME//_HOME/})
+CNODE_VNAME_UC=$(tr '[:lower:]' '[:upper:]' <<< ${CNODE_VNAME})
```
rdlrt, line 146: varname -> _UC may not be very obvious

scripts/grest-helper-scripts/db-scripts/genesis_table.sql
```diff
@@ -0,0 +1,18 @@
+DROP TABLE IF EXISTS grest.genesis;
```
rdlrt, line 1: Could possibly move this to basics.sql (same as control table)?

scripts/grest-helper-scripts/setup-grest.sh
```diff
@@ -94,25 +111,62 @@
...
+
     get_cron_job_executable "stake-distribution-update"
+    set_cron_variables "stake-distribution-update"
     install_cron_job "stake-distribution-update" "*/30 * * * *"
     
     get_cron_job_executable "pool-history-cache-update"
+    set_cron_variables "pool-history-cache-update"
     install_cron_job "pool-history-cache-update" "*/10 * * * *"
+
+    if [[ ${NWMAGIC} -eq 764824073 || ${NWMAGIC} -eq 1097911063 ]]; then
```
rdlrt, line 145: Comment -> Not applicable/Future addition for Guild network

scripts/grest-helper-scripts/setup-grest.sh
```diff
@@ -147,7 +201,35 @@
     mkdir -p ~/tmp
   }
 
+  # Description : (Re)create genesis table.
+  #             : SQL sourced from grest-helper-scrips/db-scripts/genesis_table.sql.
+  create_genesis_table() {
+    local genesis_table_sql_url="${DB_SCRIPTS_URL}/genesis_table.sql"
```
rdlrt, line 207: (Consistency: either control table have seperate sql or create genesis definition included here)

scripts/grest-helper-scripts/db-scripts/genesis_table.sql
DamjanOstrelic: Definitely an option, I'll think about it

scripts/cnode-helper-scripts/env
Scitz0: happy for suggestions 😄  .... would spelling it out like CNODE_VNAME_UPPERCASE be preferred?

scripts/grest-helper-scripts/setup-grest.sh
Scitz0: Couldn't guild network just be added later on?
As we don't have a token registry repository now, I thought it would be better just not deploy anything for that yet. Else, what should it do in the meantime? Should be an easy update later on once/if we deploy something.

DamjanOstrelic: I think it was just a comment not to remember the guild network. We could do a mock token registry repo with the same structure, but I don't think we need to do anything for the guild network for this ATM.

Scitz0: Ok, yeah I think it should be easy enough to add later on and have people run setup script again to deploy cron job for it.

rdlrt: Yip just comment within script, to avoid someone noticing it later and thinking we missed it 👍🏻

scripts/cnode-helper-scripts/env
rdlrt: prolly, unless we have any other suggestion

Scitz0 (approved):
To me, this PR is now good for merging.

DamjanOstrelic:
> To me, this PR is now good for merging.

I am just checking on mainnet, think we need to stop running cron jobs before dropping the grest schema.

Scitz0:
> > To me, this PR is now good for merging.
> 
> I am just checking on mainnet, think we need to stop running cron jobs before dropping the grest schema.

Good catch, agree, any currently running cron job should be terminated on setup-grest execution.

scripts/grest-helper-scripts/setup-grest.sh
```diff
@@ -163,7 +163,8 @@
   kill_running_cron_jobs() {
     # Currently just killing any cron entry other than the first one
     # There is probably a better (more precise) way of doing this
-    kill -9 $(pgrep cron | awk 'FNR > 1 {print $1}')
+    echo "Stopping currently running cron jobs..."
+    sudo kill -9 $(pgrep cron | awk 'FNR > 1 {print $1}')
```
Scitz0, line 167: Would it be safer to only kill entries based on path of instance? 
Something like:
`lsof | grep "${CRON_SCRIPTS_DIR}" | awk '{print $2}' | sudo xargs kill -9`

rdlrt (approved):

PR #1190, opened 2021-11-15
Add new statistics introduced in 1.31.0 to gLiveView.sh

orpheus-antpool:
Cardano Node v.1.31.0 adds several new statistics to Prometheus/EKG:
- number of forks
- delay of receiving last block since reference tip
- percentage of blocks received within 3s since node start
- number of late blocks since node start

I've added these statistics to gLiveView.sh to provide better health monitoring of the nodes. They help the community identify block propagation issues and mitigate it by tweaking the topology and/or connectivity of the nodes. To save space, I removed Tip (node) which is redundant given Tip ref and Tip diff.


rdlrt (commented):
Thanks for the PR!
One of the pre-req to this would be changing minimum baseline from 1.30.1 to 1.31.0 in env file

Scitz0:
Thanks for PR, but shouldn't we add cdfOne and cdfFive as well and not just cdfThree?

orpheus-antpool:
We could, I was just worried about eating up too much space, which was a concern last time we discussed about additional metrics :) I can easily add another line in the block propagation section and show cdfOne, cfdThree, cfdFive there. Instead of cdfThree on the fist line, I can add  % of blocks served (served/total since node start), which gives an idea about how useful the node is to others (i.e. do they find out about new blocks from me or from somebody else).

Scitz0:
> We could, I was just worried about eating up too much space, which was a concern last time we discussed about additional metrics :) I can easily add another line in the block propagation section and show cdfOne, cfdThree, cfdFive there. Instead of cdfThree on the fist line, I can add  % of blocks served (served/total since node start), which gives an idea about how useful the node is to others (i.e. do they find out about new blocks from me or from somebody else). Any thoughts on this?

Sounds good, give it a try and I'll test when pushed.

scripts/cnode-helper-scripts/gLiveView.sh
```diff
@@ -56,7 +56,7 @@ setTheme() {
 # Do NOT modify code below           #
 ######################################
 
-GLV_VERSION=v1.23.1
+GLV_VERSION=v1.23.2
```
Scitz0, line 59: Looks great, only thing as this is a bigger change and not a fix/patch, would be to bump version to 1.24.0 instead.

Scitz0 (approved):
👍

scripts/cnode-helper-scripts/gLiveView.sh
orpheus-antpool: Sounds good, looking forward to the release.

PR #1193, opened 2021-11-16
[Docker] Basic Koios setup

DamjanOstrelic:
This PR enables running (basic) grest/Koios node as a Docker service. There are still several issues that need to be tackled in order to bring the Docker setup up to speed with the current grest development. Most notable things missing:
- cron job setup for updating cached tables
- genesis table setup as the node config files and the postgres db are two separate containers

The grest setup is executed from inside the `postgres` container after starting it:
```
su - postgres
cd /
./setup-grest.sh -d
```
If the database is synced past the Mary hardfork, this will deploy the exiting endpoints.


redoracle (approved):
LGTM

scripts/grest-helper-scripts/setup-grest.sh
```diff
@@ -579,6 +582,7 @@
   update_check "$@"
   common_init
   set_environment_variables
+  [[ "${DOCKER_RUN}" == "Y" ]] && setup_db_basics && deploy_query_updates && exit 0
```
rdlrt, line 585: shouldn't this be after parse_args?

scripts/grest-helper-scripts/setup-grest.sh
```diff
@@ -24,7 +24,8 @@
 		
 		Install and setup haproxy, PostgREST, polling services and create systemd services for haproxy, postgREST and dbsync
 		
-		-f    Force overwrite of all files including normally saved user config sections 
+		-f    Force overwrite of all files including normally saved user config sections
+    -d    Docker flag - skip checks/installations not required by Docker and deploy the query updates
```
rdlrt, line 28: identation inconsistent

files/docker/db-sync/guildsetup/cfg/env
```diff
@@ -0,0 +1,119 @@
...
+# Description : Check and apply delta updates using different combinations to retain custom config
+#             : $1 = name of script to update
+#             : $2 = [Y|N] ignore question and auto update
+#             : $3 = [Y|N] do a complete file comparision
+#             : $4 = [Y|N] wait to acknowledge update
+#             : $5 = optional alternative scripts folder instead of default 'cnode-helper-scripts'
+# return code : 0 = no update
+#             : 1 = update applied
+#             : 2 = update failed
+checkUpdate() {
```
rdlrt, line 21: Just a note , since we hit something like this before - duplicating functions, especially when using same name - can mean if in weeks/months in  future the two env are sourced and not the same, the results may not be expected - can often lead to hiccups, not fully sure I understood the reason to duplicate the file, but am sure you guys would've discussed it. If needed would it make sense to rename the function to distinguish?

files/docker/db-sync/guildsetup/scripts/docker-entrypoint.sh
```diff
@@ -1,9 +1,19 @@
...
 
-apk update 2>/dev/null
-apk add socat curl gawk jq 2>/dev/null
+###################### Customisations - START ##################################
+apt-get update 2>/dev/null
+apt-get install -y socat curl gawk jq sudo 2>/dev/null
+
+# Sets the postgres user password unlocking sudo privileges
+# Should be improved to use env variables or similar
+echo -e "postgres\npostgres" |  passwd postgres
```
rdlrt, line 9: Default postgres setup deploy allows setting up db in a way that creds are not really needed (which is how guide instructions atm are), preferring socket

rdlrt (changes requested):
Overall looks like docker setup is being done **quite** differently to standard setup-grest flow, resulting not only in some duplicates/conflicting usage, but also will lead to inconsistency in what docs say versus what docker scripts does.

I am OK if that's where we'd like docker branches and scripts to progress towards (alongwith maintaining corresponding docs), in that case - just a couple of minor changes on setup-grest

files/docker/db-sync/guildsetup/scripts/docker-entrypoint.sh
DamjanOstrelic: Hm, yes, I suppose we can duplicate that on docker as well... This was just a quick fix to get it going.

scripts/grest-helper-scripts/setup-grest.sh
DamjanOstrelic: > Overall looks like docker setup is being done **quite** differently to standard setup-grest flow, resulting not only in some duplicates/conflicting usage, but also will lead to inconsistency in what docs say versus what docker scripts does.
> 
> I am OK if that's where we'd like docker branches and scripts to progress towards (alongwith maintaining corresponding docs), in that case - just a couple of minor changes on setup-grest

 I'm really not sure how we should approach Docker stuff. We'll have to discuss it further in the calls, but some ideas were:
 
 1) Completely separate monolith vs docker setup scripts / environments
 This would mean that we have to maintain both parts simultaneously. It might not be such a large overhead as the docker setup script just needs to deploy query updates.
 The other thing is the configs (HAProxy, pgREST etc...), which have would have to be maintained separately as well in this case. But we could have a base for those configs that serves both docker and monolith which would help with that.
 
 2) Integrate docker setup in the `setup-grest.sh`
 In a way, this is also doing the same thing as 1) just in a different way. We still have to maintain both docker and monolith parts of the setup, but we do it in the same script. The comment about configs still applies here.

So right now, I would be in favour of separating the two as much as possible, as it seems that Docker setup will always have to play catch-up to the monolith setup. I cannot see any way around it, and if there is no way around it, then I would prefer to have these two distinct setups maintained separately.

The PR itself is a bit of a POC to show that the Docker setup can work, but there is still a lot to improve on it.

files/docker/db-sync/guildsetup/cfg/env
DamjanOstrelic: For sure, I also don't like it, and had a comment there to explain. For now, I just duplicated the part of the `env` that the `setup-grest.sh` script needs to execute the query deployments.

PR #1198, opened 2021-11-23
Updates to HAProxy config (and minor alignments from recent discussions)

rdlrt:
HAProxy config:
- Add logging placeholder with custom format
- Prevent logs for tip check [lightweight query, and repeated for polling]
- Add backend for Unauthorized access, used by monitoring instances only
- Add CPU map to only use first two cores, specify ulimit
- Reduce DDoS limits from 100/10s to 50/10s
- Enable cache
- Enforce server nomen clature for ssl-enabled servers (will be required for grest-poll.sh updates)

setup-grest:
- Move grest-poll.sh and checkmetrics.sh updates to deploy_config function (so executed with `-f`)
- Supress update prompt for setup-grest script, as it is a lot more independent from env (and skip update can be done via command line param)
- Adapt to PostgREST CI change for binary nomen clature
- Expect baseline for dbsync status to be synched until Alonzo fork (instead of Mary)

grest.conf:
- Update Default max-observations to 1000

### DIFF
--- a/scripts/grest-helper-scripts/setup-grest.sh
+++ b/scripts/grest-helper-scripts/setup-grest.sh
@@ -273,7 +273,7 @@
   deploy_postgrest() {
     echo "[Re]Installing PostgREST.."
     pushd ~/tmp >/dev/null || err_exit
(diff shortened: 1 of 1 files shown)


Scitz0 (approved):
LGTM

PR #1202, opened 2021-11-28
doc: fix deploy-as-systemd

kgeonw:
FIX Deploy as a systemd service, now it has a separately script


DamjanOstrelic:
On `alpha`, the docs are correct and cnode will be deployed with `cnode -d`, but those updates are not yet on the master branch. 
So we definitely don't want to update the `alpha` branch for this.
But the docs are served from `master` (so they are wrong), but there should be a new merge of `alpha` into `master` soon with an announcement to go along.

PR #1204, opened 2021-11-29
Added Manual software pre-requirement for chrony

ResponsibleStaking:
Of course, this could also be part of the automatic install.
Coincashew does it this way: https://www.coincashew.com/coins/overview-ada/guide-how-to-build-a-haskell-stakepool-node/how-to-setup-chrony

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -4,7 +4,12 @@ The architecture for various components are already described at [docs.cardano.o
 
 #### Manual software pre-requirements
 
-Highly accurate system time is critical for operating a stake pool and therefore considered mandatory. Please install and configure chrony.
+
+While we do not intend to hand out step-by-step instructions, the tools are often misused as shortcuts to avoid ensuring base skillset mentioned on home page. Some of the common gotchas that we find SPOs to miss out on:
+
+    - It is imperative that pools operate with highly accurate system time , in order to propogate blocks to network in a timely manner and avoid penalties to yourself (or at times other competing blocks). Please refer to sample guidance [here ](https://ubuntu.com/server/docs/network-ntp) for details - the precise steps may depend on your OS.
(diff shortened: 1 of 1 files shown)


rdlrt:
Looks good, thanks - only thing I would potentially add is maybe a sample site pointing to details (not simplistic instructions like in case of coincashew), eg: [ubuntu site offers nice balance of explaination and instructions, that could be referred](https://ubuntu.com/server/docs/network-ntp)

docs/basics.md
```diff
@@ -2,6 +2,10 @@
 
 The architecture for various components are already described at [docs.cardano.org](https://docs.cardano.org/explore-cardano/cardano-architecture/overview) by CF/IOHK. We will not reinvent the wheel :smile:
 
+#### Manual software pre-requirements
+
+Highly accurate system time is critical for operating a stake pool and therefore considered mandatory. Please install and configure chrony.
```
rdlrt, line 7: Since we're adding manual reqs, could potentially few more known gotchas that folks skip:

```suggestion

While we do not intend to hand out step-by-step instructions, the tools are often misused as shortcuts to avoid ensuring base skillset mentioned on home page. Some of the common gotchas that we find SPOs to miss out on:

    - It is imperative that pools operate with highly accurate system time , in order to propogate blocks to network in a timely manner and avoid penalties to yourself (or at times other competing blocks). Please refer to sample guidance [here ](https://ubuntu.com/server/docs/network-ntp) for details - the precise steps may depend on your OS.
    - Ensure your Firewall rules at Network as well as OS level are updated according to the usage of your system, you'd want to whitelist the rules that you really need to (eg: limit ports that are open to world to node, SSH, and potentially secured webserver/proxy depending on components you run).
    - Update your SSH Configuration to prevent password-based logon.
```

rdlrt (approved):

PR #1207, opened 2021-11-30
Update cnode docker entrypoint

DamjanOstrelic:


rdlrt:
Redundant as fixed in #1208

PR #1214, opened 2021-12-03
Initial Testing template

rdlrt:
Todo:
 - [x] Move check tip to function
 - [x] Add handling of ssl to SCHEME if haproxy.cfg defines endpoint with name
 - [x] Move hard coded values to environment variables that can be overridden
 - [x] Add chk_endpt_get (parameters: [endpoint starting from /] [rpc|view]
 - [x] Add chk_endpt_post (parameters: [endpoint starting from /] [data to submit to POST request]
 - [x] Add comparisons for RPC structure itself (prefer local, as it only impacts failover 'from' local instance, and allows to test different branches
       This comparison is purely for RPC function name, number of parameters, and name of parameters - derived from koiosapi.yaml on github
 - [x] (External) Automate sync of koiosapi from alpha branch every 3 hours on monitoring instance
 - [x] Remove loopback of failover (you do not want to mark yourself available, doing failover to remote instance if local struct does not match - redundant hop)
       For previous iteration, if instance is DOWN but one of the peer is UP, haproxy would continue to mark instance as available making 2 additional hops within haproxy loop
       The behaviour is logged on monitoring instance - so can be easily caught if abused, but more often than not would be done unintentionally
 - [x] Ensure the postgREST limit returned is 1000
 - [x] Add updateCheck for grest-poll itself, checked hourly (or at first run post haproxy restart)
 - [ ] Add interval to download spec URL and API_COMPARE
 - [ ] Elect few endpoints that will indirectly test data
       - [x] Control Table (TODO: Version addition to Control Table)
       - [ ] Query a sample from cache table (based on inputs from control table)
       - [ ] We may not want to perform extensive dbsync data scan itself, except if bug/troublesome data on dbsync (eg: stake not being populated ~10 hours into epoch)
 - [ ] If required and entire polling takes more than 3 seconds, Randomise some of the checks between the elected endpoints
 - [x] Add `-d` flag for debug mode - to run all tests without quitting

Resultant dependency changes required by above: 
- [x] Re-vamp env to define functions first and remove checks for CNODEBIN in offline mode.
- [x] Test & handle env script's impacts to other tools


Scitz0 (approved):
LGTM

rdlrt:
> LGTM so far, is plan to complete all TODOs before merge?

I'd think we want to achieve below (from current pending) before we merge:
  - [x] Add updateCheck for grest-poll itself, checked hourly (or at first run post haproxy restart)
  - [x] Add interval to download spec URL and API_COMPARE
  - [ ] Elect few endpoints that will indirectly test data
        - [x] Control Table (TODO: Version addition to Control Table)
        - [ ] Query a sample from cache table (based on inputs from control table)
        - [ ] We may not want to perform extensive dbsync data scan itself, except if bug/troublesome data on dbsync (eg: stake not being populated ~10 hours into epoch)
  - [ ] If required and entire polling takes more than 3 seconds, Randomise some of the checks between the elected endpoints
  - [x] Add -d flag for debug mode - to run all tests without quitting

rdlrt:
As discussed during the call, merging this to alpha (to prevent re-work on fresh instances being synched atm)

PR #1216, opened 2021-12-05
Update env for variable port number

Seaside-Staking:
change fixed port number 6000 to variable ${CNODE_PORT}. fixed 6000 breaks gLiveview when not using port 6000


Scitz0 (commented):
This is not the way to change port. Just uncomment and change the port in user variables section.

Seaside-Staking:
The code later on uses 6000 and not CNODE_PORT variable

Seaside-Staking:
Ok just got it, when uncommenting port it still uses the right variable. Thank you! deleting branch

PR #1229, opened 2021-12-14
update cntools for 1.32.1 support

Scitz0:
Initial commit for CNTools node 1.32.1 support, eg. ascii -> hex in cardano-cli.

- [x] 1.32.1 enforced in env
- [ ] ~~Update ekg / gLV due to nodeStartTime being deprecated in future version? endpoint still there in 1.32.1. Ignored for now.~~
- [x] Verify that all asset stuff works in CNTools. 
- [ ] ~~more?~~


rdlrt:
> * [ ]  1.32.1 not yet enforced in env in this initial commit, should we do this?

We'd have to, as CLI across previous releases contain breaking changes, and leaving support for old syntax wouldn't be ideal

> * [ ]  Update ekg / gLV due to nodeStartTime being deprecated in future version? endpoint still there in 1.32.1.

There is supposed to be a newer metric available, until it is exposed to EKG - best to wait?

Scitz0:
> We'd have to, as CLI across previous releases contain breaking changes, and leaving support for old syntax wouldn't be ideal

Yip, we should enforce it.


> There is supposed to be a newer metric available, until it is exposed to EKG - best to wait? 

Agree, no need to change this now.

rdlrt (approved):

PR #1230, opened 2021-12-15
Fixed up time conversion for pool history check

hodlonaut:
Needed to specify a flag to indicate that the the value stored in database was UTC to get correct conversion to epoch time.


rdlrt (approved):

Scitz0 (approved):
👍

PR #1236, opened 2021-12-17
Unify RPC vs root endpoints at proxy layer

rdlrt:
- Unify RPC vs root endpoints at proxy layer using ACL file (see comments in grest-poll.sh TODOs for items)
- Update HAProxy configs to use grestrpcs and systemd definition to use variables
- Replace port requirement for stats with socket file, as it will always be local to HAProxy instance
- Update checkstatus.sh according to stats update


Scitz0 (approved):
👍

PR #1237, opened 2021-12-18
[asset_txs] Replace TEMP table with WITH

DamjanOstrelic:
Until we come up with a better solution, this will at least resolve the permission issues on the query for `web_anon` user.


Scitz0 (approved):
Giving me about the same speed on a test system as with a temp table so good to merge.

PR #1240, opened 2021-12-21
Remove hardcoded block rule, fix logging

DamjanOstrelic:
the block was hardcoded so always looking at the same state. The "block height not reached" logline was printed even on update runs .


Scitz0 (approved):
👍

PR #1258, opened 2022-01-01
Minor patches

rdlrt:
- Update GHC version to 8.10.7 (as per upstream changes for node)
  - Update cabal-build-all.sh to add back removal of cardano-config folder, as some components still show old git revision
  - Add epoch_info to grest-poll tests
  - Add Testnet instance URL to API spec
  - For polling GET requests, fail if curl response is invalid, and print response code if DEBUG flag is enabled
  - Bump individual instance flood limit (gateway will still have them), avoids individual instances to know about/configure whitelisted nodes

### DIFF
--- a/docs/Build/node-cli.md
+++ b/docs/Build/node-cli.md
@@ -31,15 +31,15 @@ The above would copy the binaries built into `~/.cabal/bin` folder.
 
 ### Verify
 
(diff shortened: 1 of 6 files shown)


Scitz0 (approved):
👍

DamjanOstrelic (approved):

PR #1265, opened 2022-01-10
Patch for active stake polling check

rdlrt:
Use (epochlength)/12 as a reference point post epoch transition to do a health check, avoids absolute outage post transition.
Downside: for a period post transition, there could be a small duration of inconsistency where a node may have completed latest epoch stake cache, while others may yet be catching up, for now - the immediate aim is to not have all nodes fail health checks.


Scitz0 (approved):
👍

PR #1267, opened 2022-01-10
CNTools 9.0.0

Scitz0:
Koios integration into CNTools for utxo and pool info.


files/grest/rpc/address/address_info.sql
```diff
@@ -2,43 +2,44 @@ DROP FUNCTION IF EXISTS grest.address_info (text);
 
 CREATE FUNCTION grest.address_info (_address text DEFAULT NULL)
   RETURNS TABLE (
-    balance numeric,
```
DamjanOstrelic, line 5: Any specific reason we are switching this field to string?

rdlrt (approved):
Brilliant overall! Few considerations:
- Add logging for every curl request made to Koios by default to cntools-history.log
- Knowing previous community requests, rather than relying on unset of KOIOS_API variable OR editing it to invalid value, would be good to have a ENABLE_KOIOS flag instead (defaulting to 'Y") in env
- For ` >> ADVANCED >> CHAIN-ANALYSIS >> POOL QUERIES >> METADATA HASH` , not sure if we need to show the script before execution, could possibly reduce the prompt?

Scitz0:
> Brilliant overall! Few considerations:
> 
> * Add logging for every curl request made to Koios by default to cntools-history.log
👍 
> * Knowing previous community requests, rather than relying on unset of KOIOS_API variable OR editing it to invalid value, would be good to have a ENABLE_KOIOS flag instead (defaulting to 'Y") in env
👍 **sounds good as well.** 
> * For ` >> ADVANCED >> CHAIN-ANALYSIS >> POOL QUERIES >> METADATA HASH` , not sure if we need to show the script before execution, could possibly reduce the prompt?
**Could print to less as we do for changelog, in case of big script.**

Scitz0:
Updates added according to notes.

PR #1275, opened 2022-01-15
Stringify all lovelace / token quantities in query outputs

Scitz0:
Epoch params left as numeric


rdlrt (approved):
Breaking change, will need to coordinate via comms

PR #1278, opened 2022-01-17
rely VersionCheck on CNODEBIN

gufmar:
when gLiveView lost and restored connection to a running instance, `getNodeMetrics` didn't use `$CNODEBIN` reference to get the actually running node version. Configurations using a custom `env:$CNODEBIN` location ended up displaying the node version from the default `cabal/bin/cardano-node` folders binary


Scitz0:
👍 I will hijack this PR to also add infinite retries if set to 0 as requested by a user.

rdlrt (approved):

PR #1295, opened 2022-01-29
tx_list fix + tx_info collateral and performance boost

Scitz0:
- Fix for tx_list that didn't filter out pools correctly based on retire status.
- Collateral inputs added to tx_info endpoint + koiosapi spec update
- tx_info rewritten to not join on CTE but instead select on it in a subquery. This gives a big performance boost as CTEs are inefficient on joins. Quick test showed ~6-8x performance boost.

### DIFF
--- a/files/grest/rpc/pool/pool_list.sql
+++ b/files/grest/rpc/pool/pool_list.sql
@@ -8,19 +8,32 @@ CREATE FUNCTION grest.pool_list ()
   AS $$
   # variable_conflict use_column
 BEGIN
-  RETURN QUERY SELECT DISTINCT ON (pic.pool_id_bech32)
-    pool_id_bech32,
-    pod.ticker_name
-  FROM
-    grest.pool_info_cache AS pic
-  LEFT JOIN public.pool_offline_data AS pod ON pod.pmr_id = pic.meta_id
-WHERE
-  pic.pool_status != 'retired'
-ORDER BY
-  pic.pool_id_bech32,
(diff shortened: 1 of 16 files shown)


files/grest/rpc/pool/pool_list.sql
```diff
@@ -8,19 +8,32 @@ CREATE FUNCTION grest.pool_list ()
...
-WHERE
-  pic.pool_status != 'retired'
-ORDER BY
-  pic.pool_id_bech32,
-  pic.tx_id DESC;
-END;
-$$;
 
-COMMENT ON FUNCTION grest.pool_list IS 'A list of all currently registered/retiring (not retired) pools';
+  RETURN QUERY (
```
rdlrt, line 12: can we change this to return table format? (consistency as well as unlocks full potential for PostgREST API)

PS: There is a seperate ticket for this request, so can be looked at seperately for almost all pool queries

DamjanOstrelic: I am confused - the function is defined as `RETURNS TABLE`. I thought we have to `RETURN QUERY` in the actual function and `RETURN QUERY` actually returns a table?

rdlrt: ah - my bad, I misread it

DamjanOstrelic (approved):
Tested on guild network 👍

rdlrt (approved):

PR #1297, opened 2022-01-30
[RPC] accept list of IDs for pool_info

DamjanOstrelic:
Switched to `POST` endpoint to support sending multiple pool IDs at once. This is a breaking change for consumers of this endpoint.

`active_epoch_no` -> this field should probably be renamed as it returns `epoch when the last pool update became active` from what I can tell.


Scitz0 (approved):
👍  I can see merges got a bit mixed but don't matter. Changes related to this PR look good to me.

PR #1298, opened 2022-01-30
GRest Patches

rdlrt:
- Move haproxy.cfg - is_wss acl to after app-secured definition
- Update epoch_info check to look at prev epoch - since cache is checked seperately


Scitz0 (approved):
👍

PR #1300, opened 2022-01-31
Added cast for supply to fix pool_history_cache

hodlonaut:
Please don't merge just yet, will need to check for more supply references


hodlonaut:
Did a search for supply under files/grest and this was the only reference I could locate that seemed to need a cast.

Scitz0 (approved):
👍

PR #1301, opened 2022-02-01
CNTools - 9.0.3 (Allow TX_TTL to be user defined)

rdlrt:
TTL timings in CNTools used atm are a bit borderline given the chain competitiveness.. and transaction being successful depends on external variables beyond node's [default] control. Hence, bump default TTL expiry (instead of 10 minutes) to 1 hour


Scitz0 (approved):
Looks good to me, comment in cntools.library when setting ttl variable still mentions 10min as default. But just a minor thing, change in itself is good.

PR #1305, opened 2022-02-04
pool_list and API Spec fixes

rdlrt:
- pool_list : On postgrest v14.1 , missing alias reference on ORDER BY may fail
- API Specs: _policy_id was missing


hodlonaut (approved):
Looks good

PR #1307, opened 2022-02-05
introduce blockLog.sh

gufmar:


scripts/cnode-helper-scripts/blockLog.sh
```diff
@@ -0,0 +1,122 @@
...
+blockHeightPrev=$(curl -s -H 'Accept: application/json' http://127.0.0.1:$(cat $CONFIG | jq .hasEKG)/ | jq -r .cardano.node.metrics.blockNum.int.val)
+blockTimeDeltaSlots=0
+
+getDeltaMS() {
+	echo $(echo "$2 $4" | awk -F '[:, ]' '{print ($1*3600000+$2*60000+$3*1000+$4)-($5*3600000+$6*60000+$7*1000+$8) }')
+}
+
+while true
+do
+	blockHeight=$(curl -s -H 'Accept: application/json' http://127.0.0.1:$(cat $CONFIG | jq .hasEKG)/ | jq -r .cardano.node.metrics.blockNum.int.val)
```
Scitz0, line 41: This is already done outside of loop:
`ekgport=$(cat $CONFIG | jq .hasEKG)`
No need to waste resources on this jq call at each loop. 
Also, if Option B is utilized, there is no need to grab ekg port as env does this for you. Both host and port is available as ${EKG_HOST}:${EKG_PORT}. 
So code could be changed to:
[[ -z ${EKG_PORT} ]] && EKG_PORT=$(jq .hasEKG $CONFIG)
And then use ${EKG_PORT} everywhere in script.

scripts/cnode-helper-scripts/blockLog.sh
```diff
@@ -0,0 +1,122 @@
...
+						;;
+					SendFetchRequest)
+						[ -z "$blockTimeSfr" ] && blockTimeSfr=$(date -d $(jq -r .at <<< $blockLogLine) +"%F %T,%3N")
+						;;
+					CompletedBlockFetch)
+						[ -z "$blockTimeCbf" ] && blockTimeCbf=$(date -d $(jq -r .at <<< $blockLogLine) +"%F %T,%3N")
+						[ -z "$blockTimeCbfAddr" ] && blockTimeCbfAddr=$(jq -r .data.peer.remote.addr <<< $blockLogLine)
+						[ -z "$blockTimeCbfPort" ] && blockTimeCbfPort=$(jq -r .data.peer.remote.port <<< $blockLogLine)
+						[ -z "$blockDelay" ] && blockDelay=$(jq -r .data.delay <<< $blockLogLine)
+						[ -z "$blockSize" ] && blockSize=$(jq -r .data.size <<< $blockLogLine)
```
Scitz0, line 71: Any way to reduce repetitive external jq calls helps to reduce resource imprint the script has. I noticed this with gLiveView, that a single call that grabs all values is better. Then let bash deal with splitting it into different variables if needed. See getNodeMetrics() in env. It grabs all values and pipes it out as tsv, then this is split into an array so each result in array contain selected values. 
https://github.com/cardano-community/guild-operators/blob/master/scripts/cnode-helper-scripts/env#L466

Scitz0 (commented):
Just some comments on decreasing resource usage.

scripts/cnode-helper-scripts/blockLog.sh
gufmar: considered this, thanks

scripts/cnode-helper-scripts/blockLog.sh
gufmar: I implemented it like this

Scitz0 (approved):
Looks good

rdlrt (approved):

PR #1314, opened 2022-02-07
Fix prereqs (ogmios option) for alpha branch users

rdlrt:
Other utilities do not use `unzip` - this results in those who were using correct commands (eg: `./prereqs.sh -b alpha -t delme -n guild -o`) to receive a failure when attempting to extract ogmios.


Scitz0 (approved):
👍

PR #1316, opened 2022-02-07
fix eLoop

gufmar:
avoid endless loop on node restarts (temp EKG timeout)

### DIFF
--- a/scripts/cnode-helper-scripts/blockLog.sh
+++ b/scripts/cnode-helper-scripts/blockLog.sh
@@ -44,9 +44,9 @@ getDeltaMS() {
 
 while true
 do
-  blockHeight=$(curl -s -H 'Accept: application/json' http://${EKG_HOST}:${EKG_PORT}/ | jq -r .cardano.node.metrics.blockNum.int.val)
-  if [ -z $blockHeight ] || [ "$blockHeight" == "null" ]; then
-    echo "WARN: can't query EKG on http://127.0.0.1:$(cat $CONFIG | jq .hasEKG)/ ..."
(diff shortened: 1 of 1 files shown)


scripts/cnode-helper-scripts/blockLog.sh
```diff
@@ -45,7 +45,7 @@ getDeltaMS() {
 while true
 do
   blockHeight=$(curl -s -H 'Accept: application/json' http://${EKG_HOST}:${EKG_PORT}/ | jq -r .cardano.node.metrics.blockNum.int.val)
-  if [ -z $blockHeight ] || [ "$blockHeight" == "null" ] || [ "$blockHeightPrev" == "null" ] ; then
+  if [ -z $blockHeight ] || [ "$blockHeight" == "null" ]; then
     echo "WARN: can't query EKG on http://127.0.0.1:$(cat $CONFIG | jq .hasEKG)/ ..."
```
Scitz0, line 49: This error message should also use ${EKG_PORT} ?

Scitz0:
If it fails to get `blockHeight`, the response will be `null` with the current code. This is then written to `blockHeightPrev`. If at the next loop ekg does respond with a valid value it will try to do math on `blockHeightPrev` when it contains the string `null`.

You could add a deafult value of 0 to `jq -r .cardano.node.metrics.blockNum.int.val //0` when setting `blockHeight`. And then instead of checking for null, do `-eq 0` comparison. Then 0 will be written to the previous value just like on the first start.

scripts/cnode-helper-scripts/blockLog.sh
gufmar: ok, done

Scitz0 (approved):
👍

PR #1318, opened 2022-02-08
parse IPv6 segments

gufmar:
parse IPv6 addresses, using same internal : separator as peer data already does for RTT:address:port

`RTT:ip:v6:addr::es:port:`


Scitz0 (approved):
👍

PR #1320, opened 2022-02-08
[RPC] Asset history

DamjanOstrelic:
Should we return asset quantities as strings as well?

### DIFF
--- a/files/grest/apispec/koiosapi.yaml
+++ b/files/grest/apispec/koiosapi.yaml
@@ -703,6 +703,26 @@ paths:
           $ref: "#/components/responses/NotFound"
       summary: Asset Information
       description: Get the information of an asset including first minting & token registry metadata
+  /asset_history: #RPC
+    get:
+      tags:
+        - Asset
+      parameters:
+        - $ref: "#/components/parameters/_asset_policy"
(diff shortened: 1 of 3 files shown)


Scitz0:
> Should we return asset quantities as strings as well?

Yes we should, just look at hosky tokens etc. Not sure what their supply is but I'm sure its a big number :)

files/grest/apispec/koiosapi.yaml
```diff
@@ -2038,7 +2058,6 @@ components:
             example: 23
     asset_summary:
       type: array
-      description: An array of assets contained on input UTxO
```
rdlrt, line 2041: Description deleted?

DamjanOstrelic: Yes, cause it was copied from another endpoint. It doesn't fit:
image

I wasn't sure what to replace it with so I just removed it (it would be the subtitle under schema in the image)

rdlrt: > cause it was copied from another endpoint

gotcha 👍🏻

rdlrt (approved):

PR #1321, opened 2022-02-10
Minor patch: Align mem stats for getmetrics with RHEL article

rdlrt:
Simplify memused calculation as per [article from redhat](https://access.redhat.com/solutions/406773), which also aligns it more accurately with top output


hodlonaut (approved):
Thumbs up

PR #1329, opened 2022-02-14
Fixes for polling, env and CNTools-9.0.4

rdlrt:
Issues being fixed:
1. env should set most variables even if cardano-node is not in repository
2. checkUpdate <full-path>/grest-poll.sh did not work due to assumptions on current folder (eg: haproxy starts in it's home folder , resulting in failure when sudo runs from `/` path)
3. Pool > Info and Pool > Rotate were broken when using Koios (since the pool_info endpoint wasn't updated)
4. Auto-update mechanism for grest-poll (hourly, resulting in corresponding download of koiosapi.yaml and grestrpcs files) did not work in all cases before

Changes:
- env: Move Defaults vars to be called before any checks
- env: checkUpdate - split ${1} references to differentiate between filename and dirname
- grest-poll.sh: Specify folder instead of assuming relative path
- grest-poll.sh: Remove UPDATE_CHECK in grest-poll.sh, as these are supposed to auto update (and can be toggled using SKIP_UPDATE for testing)
- grest-poll.sh: Add handling for leaf_paths being integers/arrays
- setup-grest.sh: Add common_updates function, which will always run checkUpdate against static files, and rebuild grestrpcs/koiosapi spec
- CNTools-9.0.4: cntools.library - Update pool_info endpoint references

### DIFF
--- a/docs/Scripts/cntools-changelog.md
+++ b/docs/Scripts/cntools-changelog.md
@@ -6,6 +6,10 @@ All notable changes to this tool will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [9.0.4] - 2022-02-14
+#### Fixed
+- Update request for pool_info endpoint from Koios
+
 ## [9.0.3] - 2022-02-01
 #### Added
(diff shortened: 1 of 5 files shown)


Scitz0 (approved):
LGTM

DamjanOstrelic (approved):

PR #1335, opened 2022-02-17
Minor patch for topoUpd and getmetrics scripts

rdlrt:
- topologyUpdater.sh: Exit if curl command fails (eg: due to timeout) and CUSTOM_PEERS are not set
- getmetrics.sh: Use tuple CSV to grep numeric output for psql instead of awk filtering (eg: my instance had `\timing` turned on, disrupting psql output filtering for line number)


hodlonaut (approved):
Looks good

PR #1340, opened 2022-02-20
checkUpdate fix

Scitz0:
## Description
After change to checkUpdate it now expects either full path to script to update from calling script or that script is called from same folder. Various scripts updated to provide full path to function.


gufmar (approved):
LGTM

rdlrt (approved):

PR #1344, opened 2022-02-22
/pool_history - Return information about pool stake, block and reward history in a given epoch

hodlonaut:
First cut of the simplistic pool history endpoint. Decided not to return just everything since returning pool id seemed a bit weird/redundant, let me know if there's anything else that catches your eye (I searched guild operators alpha branch check-out for references to another small endpoint and came up with just koiosapi.yaml reference and the rcp sql file that were modified).

### DIFF
--- a/files/grest/apispec/koiosapi.yaml
+++ b/files/grest/apispec/koiosapi.yaml
@@ -862,7 +862,11 @@ paths:
         - $ref: "#/components/parameters/_epoch_no"
       responses:
         "200":
-          $ref: "#/components/responses/OK"
+          description: Array of pool history information
+          content:
+            application/json:
+              schema:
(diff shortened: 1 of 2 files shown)


Scitz0 (approved):
LGTM but needs schema/example API data as well.

Scitz0 (dismissed):
I just came to think of something, we decided to switch to return all lovelace values as strings instead of number.

So these fields: active_stake, pool_fee_fixed, pool_fees and deleg_rewards.

With this said, I also noticed that pool_fees and deleg_rewards has number in ADA and not lovelace. We use lovelace everywhere else and I think this should be returned for all endpoints. ADA conversion if wanted is up to end user. But looks like this is how its saved in pool_history_cache.

rdlrt (approved):

PR #1331, opened 2022-02-16
Koios API docs - missing endpoint schemas

DamjanOstrelic:
Finalising API doc missing schemas:

Block:
 - [x] /blocks
 - [x] /block_info
 - [x] /block_txs

Address:
 - [x] /address_info
 - [x] /address_txs
 - [x] /address_assets
 - [x] /credential_txs
 
 Account:
 - [x] /account_list
 - [x] /account_info
 - [x] /account_rewards
 - [x] /account_updates
 - [x] /account_addresses
 - [x] /account_assets
 
 Pool:
 - [x] /pool_delegators
 - [x] /pool_blocks
 - [x] /pool_updates
 - [x] /pool_relays
 - [x] /pool_metadata

PR #1355, opened 2022-02-26
Add gaps found by Dudi

Scitz0:
- [x] Address Information:
  - AddressType (shelley, byron, etc.)
    What matters on protocol is presence of stake address, byron vs shelley address differentiation is best handled on lib (or utilities)
  - Datum Hash => OK
  - Order UTxO by time: We will be adding block height/time allowing sorting by key values (rather than making that a server side load which will slower responses)
  - Address Script => OK 

- [x] Address/Credential Transaction:
  - Tx Index => Performance impact for large addresses (over 300K txs), hence - we provide ways to grab a list of tx which can be sent in bulk to tx info
  - Block Height => OK
  - Block Time => OK
  - Order by Block Time/Height (no ordering, user action if needed using postgrest ordering)

- ~~Block List:~~
  - Output & Fee (Not sure if output is needed, this is a value which is statistical. It's an additional join on DB which should atleast need an outcome defined first to consider) => Instead, we will add bulk hash query to block_info (similar to address, tx and pool)
  - Confirmation Count => OK (for block_info)
  - Next/Previous hash => You already have a list ordered with prev/next :) Fields are already present in block_info

- [x] Block Info:
  - Output => OK
  - Fees => OK
  - Confirmation Count => OK

- [x] Epoch Info:
  - Epoch start time => OK (mathematical), real epoch start is defined by first block 
  - Epoch end time => OK (mathematical), real epoch end is defined by first block 

- [x] Asset Info:
  - Initial Mint Tx Hash => OK
  - Mint and Burn count => OK

- [x] Script Info:
  => OK

- [x] koiosapi
  - connected to #1331, add after merge of this PR.


files/grest/rpc/address/address_info.sql
```diff
@@ -25,12 +26,16 @@ BEGIN
   SELECT
     SUM(tx_out.value)::text,
     SA.view,
+    bool_or(tx_out.address_has_script),
     COALESCE(
       JSON_AGG(
         JSON_BUILD_OBJECT(
           'tx_hash', ENCODE(tx.hash, 'hex'), 
           'tx_index', tx_out.index,
+          'block_height', block.block_no,
```
DamjanOstrelic, line 35: Existing endpoints that reference block height are referenced as `height` only. Which one should we use to align across endpoints?

Scitz0: yeah, height can mean more than block, like slot height.

files/grest/rpc/assets/asset_info.sql
```diff
@@ -59,27 +79,22 @@ BEGIN
...
-    ) AS total_supply,
-    (
+    ) AS token_registry_metadata
+  FROM 
+    multi_asset MA
+    LEFT JOIN LATERAL (
       SELECT
-        MIN(B.time) AS date
+        MIN(B.time) AS date,
+        SUM(MTM.quantity)::text AS amount,
```
DamjanOstrelic, line 88: should we call this total supply?

Scitz0: It is in the final output, but if you want to rename this internal column as well I'm fine with that.

files/grest/rpc/transactions/tx_info.sql
```diff
@@ -471,6 +473,86 @@ BEGIN
...
+      _all_plutus_contracts AS (
+        SELECT
+          tx_id,
+          JSON_AGG(data) AS list
+        FROM (
+          SELECT
+            redeemer.tx_id,
+            JSON_BUILD_OBJECT(
+              'address', INUTXO.address,
+              'script_hash', ENCODE(script.hash, 'hex'),
```
DamjanOstrelic, line 512: Are these 2 fields the same?

Scitz0: One is an address and one is a hash of script. Or did you mean other fields than INUTXO.address and script.hash ?

rdlrt:
Before the testing itself, we had few failures for setup script execution
```
    Downloading pSQL executions from subdir script
      Deploying Function :   native_script_list
        ERROR: NOTICE:  function grest.native_script_list() does not exist, skipping
DROP FUNCTION
CREATE FUNCTION
ERROR:  could not find a function named "grest.script_list"
      Deploying Function :   plutus_script_list
        ERROR: NOTICE:  function grest.plutus_script_list() does not exist, skipping
DROP FUNCTION
CREATE FUNCTION
ERROR:  could not find a function named "grest.script_list"
      Deploying Function :   script_redeemers
```

rdlrt:
Added updates/fixes where failed against schema:
image

rdlrt (approved):

PR #1358, opened 2022-02-28
Remove uptime query from Node

### LINKED ISSUE #1357: [gLiveView] No longer connects to node v1.34.0
**Describe the bug**
After re-compiling cardano-node to use the 1.34.0 release, gLiveView seems to fail to connect to cardano-node. The same issue occurs on both my BP and relay servers.

**To Reproduce**
1. Update the running cardano-node instance to v1.34.0
2. Run `gLiveView.sh`
3. See the error `Connection to node lost, retrying (2/3)!`

**Expected behavior**
gLiveView continues to connect to the cardano-node instance as normal

**Version:**
 - OS: Ubuntu Server 20.04 LTS
 - Product version: gLiveView 1.26.0
 - Cardano Node version: 1.34.0
 - Network you're connecting to: mainnet

**Additional context**
Attaching to the node service via journalctl shows that both my BP and relay nodes seem to be running as expected. Additionally if I press the up arrow key while gLiveView is trying to connect, I see gLiveView as it should appear, showing current node status as of the time I started gLiveView, but frozen.

So from what I'm able to gather so far this is not a node issue but a gLiveView problem

rdlrt:
- gLV-1.26.2: Remove nodeStartTime and use uptime (in seconds) instead, using 5s of uptime - to avoid detecting restart loop of node - eg: due to invalid bin path - as `Starting` status
  - env: Remove calls to nodeStartTime from EKG/prom, use OS `ps` command calls instead
  - Remove sLiveView as the backlog has increased quite a bit, and gLiveView has gone much further comparatively (in adoption as well as features)

PS: This does not mean tools are tested for node-1.34.0, but the obvious blocker for gLV to proceed is tackled (closes #1357). Since this isnt a breaking change for those running node-1.33.x, the minimum node version is not bumped yet


hodlonaut (approved):
Looks good

PR #1359, opened 2022-03-01
Prepare epoch start/end times in cache

DamjanOstrelic:
## Description

Adding epoch start/end time to epoch_info_cache.
~This is a PoC ATM - working on mainnet/testnet, still needs a way to work on guildnet.~ Ready for test and merge


DamjanOstrelic:
This will be switched to using an SQL helper function to calculate start/end times without storing them.

rdlrt (approved):

PR #1361, opened 2022-03-02
gLV-1.26.3

rdlrt:
Instead of checking uptime to ensure node is up, use CNODE_PID

That should leave gLiveView in `Starting..` state as long as CNODE_PID resolves


Scitz0 (approved):
👍

PR #1362, opened 2022-03-02
Gaps API doc update

DamjanOstrelic:
## Description

API doc update for upcoming gaps PR + some consistency changes.
~Still need to complete `inputs/outputs` fields of the tx_info endpoint.~


Scitz0 (approved):
👍

rdlrt:
This will cause a merge conflict with template specs [here](https://github.com/cardano-community/koios-artifacts/blob/main/specs/koiosapi.yaml).

Probably best to set the standard:
1. Remove apispec references from grest, and replace with spec references for corresponding network in koios-artifacts repo
2. Use ##<unique param|responseBodies-name>## in template specs for examples instead of actual value
3. Map the above in [spec examples](https://github.com/cardano-community/koios-artifacts/blob/main/specs/spec_examples.json).

If needed, we can go ahead with this for now and sync back the corrections after this PR

DamjanOstrelic:
> This will cause a merge conflict with template specs [here](https://github.com/cardano-community/koios-artifacts/blob/main/specs/koiosapi.yaml).
> 
> Probably best to set the standard:
> 
> 1. Remove apispec references from grest, and replace with spec references for corresponding network in koios-artifacts repo
> 2. Use ##<unique param|responseBodies-name>## in template specs for examples instead of actual value
> 3. Map the above in [spec examples](https://github.com/cardano-community/koios-artifacts/blob/main/specs/spec_examples.json).
> 
> If needed, we can go ahead with this for now and sync back the corrections after this PR

Right, I forgot 🙈

DamjanOstrelic:
We agreed that updates on this PR relating to examples need to be switched over to the koios-artifacts repo, but can go with this for now

PR #1367, opened 2022-03-07
Various minor cleanups

rdlrt:
### env:
- Reduce jq calls for parsing config and genesis files - for options that will always be present
- For mainnet/testnet/staging/guild networks, dont need to check for SHELLEY_TRANS_EPOCH variable to pre-exist as they're constant
- NETWORKID is no longer used for environments, can use NWMAGIC instead

### prereqs.sh:
- Add updateWithCustomConfig for setup-grest.sh

### getmetrics.sh:
- Remove duplicate reference for env
- Use tip from variable already available

### cntools.library:
- Remove HASH_IDENTIFIER (Ddz issue which required this seperation was resolved a while ago)

### cntools.sh:
- Replace NETWORKID check with NWMAGIC

### setup-grest.sh/dbsync.sh:
- Use genesis json variables already present in env instead of parsing config again


Scitz0 (approved):
👍  nice

scripts/cnode-helper-scripts/prereqs.sh
```diff
@@ -530,6 +530,7 @@ updateWithCustomConfig "topologyUpdater.sh"
 updateWithCustomConfig "logMonitor.sh"
 updateWithCustomConfig "cncli.sh"
 updateWithCustomConfig "blockPerf.sh"
+updateWithCustomConfig "setup-grest.sh"
```
Scitz0, line 533: Need to download file as .tmp as well.

Scitz0 (changes requested):

Scitz0 (approved):

Issue #1375, opened 2022-03-20
Installation ghcup with Error: syntax error near unexpected token `(' 


MichaelTMS:
**Describe the bug**
I got an error while build docker stage1 image. 

```
Install ghcup (The Haskell Toolchain installer) ..
bash: line 393: syntax error near unexpected token `('
bash: line 393: `			warn "[P] Yes, prepend  [A] Yes, append  [N] No  [?] Help (default is \"P\")."'
Upgrading ghcup ..
Installing GHC v8.10.7 ..
./prereqs.sh: line 267: ghcup: command not found
./prereqs.sh: line 268: ghcup: command not found
./prereqs.sh: line 269: ghc: command not found
Installing Cabal v3.6.2.0..
./prereqs.sh: line 278: ghcup: command not found
```

**Screenshots**
![image](https://user-images.githubusercontent.com/90009297/159186573-07239c88-acd8-4e29-9eab-4172654407e2.png)

**Version:**
 - OS: Ubuntu 20.04.4 LTS 
 - Cardano Node version: 1.34.1

rdlrt:
Thanks for reporting , looks like ghc setup had an update in last few days...the mentioned line 293 is coming from ghcup bootstrap script that's downloaded from haskell site...the corresponding change they added seems to be relevant for windows, but has kinda caused a broken build for interactive session on bash.

We might have to patch it for now removing interactive support (as I doubt anyone really uses ghcup interactive anymore).

PR #1376, opened 2022-03-21
Update ghcup deployment steps as per upstream env variables support

rdlrt:
## Description

ghcup installer has had a few changes to support environment variables for different cases. Also, this included changes to read statements they previously used (in hindsight, the sed for read was a bit too aggressive anyways) - causing failures for execution against `0.1.17.5` version of ghcup installer. Now that there are better environment variables provided, we're able to use them and remove the hack we had to implement before for quiet mode of install

## Motivation and context

- Remove hack (that was used due to absence of control in ghcup installer) with better solution
- Current prereqs.sh was broken as a result of changes upstream

## Which issue it fixes?

- #1375

## How has this been tested?

- Only tested on CentOS locally, but PR checks should check for other OS now

### DIFF
--- a/scripts/cnode-helper-scripts/prereqs.sh
+++ b/scripts/cnode-helper-scripts/prereqs.sh
@@ -107,7 +107,6 @@ shift $((OPTIND -1))
 [[ -z ${INSTALL_VCHC} ]] && INSTALL_VCHC='N'
 [[ -z ${INSTALL_OGMIOS} ]] && INSTALL_OGMIOS='N'
 [[ -z ${CNODE_NAME} ]] && CNODE_NAME='cnode'
-[[ -z ${INTERACTIVE} ]] && INTERACTIVE='N'
 [[ -z ${CURL_TIMEOUT} ]] && CURL_TIMEOUT=60
 [[ -z ${UPDATE_CHECK} ]] && UPDATE_CHECK='Y'
 [[ -z ${SUDO} ]] && SUDO='Y'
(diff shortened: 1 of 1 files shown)


scripts/cnode-helper-scripts/prereqs.sh
```diff
@@ -107,7 +107,6 @@ shift $((OPTIND -1))
 [[ -z ${INSTALL_VCHC} ]] && INSTALL_VCHC='N'
 [[ -z ${INSTALL_OGMIOS} ]] && INSTALL_OGMIOS='N'
 [[ -z ${CNODE_NAME} ]] && CNODE_NAME='cnode'
-[[ -z ${INTERACTIVE} ]] && INTERACTIVE='N'
```
rdlrt, line 110: * this is only being removed from here as it was a duplicate of line 101

chadle-git:
I received this error on a new build running Ubuntu 20.04

$ ./prereqs.sh ghcup-patch
Using apt to prepare packages for "Ubuntu" system
  Updating system packages...
  Installing missing prerequisite packages, if any..
Install ghcup (The Haskell Toolchain installer) ..
bash: line 393: syntax error near unexpected token `('
bash: line 393: `                       warn "[P] Yes, prepend  [A] Yes, append  [N] No  [?] Help (default is \"P\")."'
Upgrading ghcup ..
Installing GHC v8.10.7 ..
./prereqs.sh: line 267: ghcup: command not found
./prereqs.sh: line 268: ghcup: command not found
./prereqs.sh: line 269: ghc: command not found
Installing Cabal v3.6.2.0..
./prereqs.sh: line 278: ghcup: command not found
Creating Folder Structure ..
Environment Variable already set up!
IOG fork of libsodium installed to /usr/local/lib/
Downloading files...

Scitz0 (approved):
Lgtm

PR #1377, opened 2022-03-21
Update db size metrics to include total relation (instead of table)

rdlrt:
## Description
Current `getmetrics.sh` only captures table size, resulting in large difference of actual db size versus what's seen in grafana monitoring


Scitz0 (approved):
👍

Issue #1332, opened 2022-02-16
SELECT GREST.STAKE_DISTRIBUTION_CACHE_UPDATE_CHECK(); taking hours on testnet instances


pyromaniac3010:
**Describe the bug**
We are monitoring the haproxys from our grest-instances via zabbix and getting constantly errors for the grest local backend on our testnet instances.
I get a lot of "Stake Distribution cache too far from tip !!" timeouts in haproxy for the local grest backend on my **testnet** instances. 
After checking postgresql I see that there is a running query "SELECT GREST.STAKE_DISTRIBUTION_CACHE_UPDATE_CHECK();
" which is running for a very long time:
```
SELECT pid, age(clock_timestamp(), query_start), usename, query
FROM pg_stat_activity
WHERE query != '<IDLE>' AND query NOT ILIKE '%pg_stat_activity%'
ORDER BY query_start desc;
  pid  |       age       | usename  |                                        query
-------+-----------------+----------+-------------------------------------------------------------------------------------
 27662 | 02:18:49.286034 | cardano  | SELECT GREST.STAKE_DISTRIBUTION_CACHE_UPDATE_CHECK();
  1490 | 02:41:39.990768 | cardano  | LISTEN "pgrst"
```
All instances were setup and optimized according to the docs (including postgres optimizations) and I used alpha branch, becaus master branch does not work for grest.
I was using AWS r5a.xlarge for testnet and r5a.2xlarge for mainnet.
I did not experience these problems with mainnet yet. Only my testnet instances report the local haproxy backends going down.
I switched to t3a.2xlarge for one testnet instance but still get these long running querys and the local haproxy backend goes down.

**To Reproduce**
1. setup grest with testnet

**Expected behavior**
It should not take many hours on testnet.

**Screenshots**

**Version:**
 - OS: Ubuntu 
 - Product version: 
 - Cardano Node version: cardano-node 1.33.0 - linux-x86_64 - ghc-8.10
 - Network you're connecting to: testnet

**Additional context**

DamjanOstrelic:
One more thing to check for useful info here are logs under `CNODE_HOME/logs/stake-distribution-update.log`

pyromaniac3010:
@DamjanOstrelic
Here the stake-distribution-update.log from one testnet grest setup:
[stake-distribution-update.log](https://github.com/cardano-community/guild-operators/files/8087469/stake-distribution-update.log)

DamjanOstrelic:
[...]

PR #1378, opened 2022-03-21
[Placeholder] - Stake Distribution Cache Optimisation

rdlrt:
## Description
- [x] Replace temp tables by 'with' statements [No optimisation, only to address very high loadavg write IOps]
- [x] Investigate UTxO performance issue
- [x] Consider alternative methods
- [x] Implement Re-designed stake distribution cache

## Motivation and context

Just some notes from discussions offline:
1. Current stake distribution does not scale well as history of chain increases, in particular - lookup for all UTxOs associated with every registered stake account is getting unmanagable.
2. For every `SELECT GREST.STAKE_DISTRIBUTION_CACHE_UPDATE_CHECK();` run , the above results in 10-12GB of space being consumed in `pg_tmp` folder depending on estimates detected by query planner , which also depends on tiles currently cached. Other instances end up evading this with ~120-150MB of temp space.
image 
image
3. Furthermore (in addition to above difference based on materialize, on certain instances - problem is magnified further. During the run , it also triggers an AUTOVACUUM on epoch[_stake] table, resulting in wait times of hours instead of few minutes post completion of temp table creation. We still haven't been able to drill down to the exact reason of what triggers this yet), but it's certainly not something we can expect to keep going with forever relying on tuning/config
4. The presence of stake_address and pool_hash vs their id makes a huge difference in size of the table itself, but it hardly moves the timings - and ends up being not so relevant (tho we'd want prolly want to avoid insert of 670MB when it can be done in 35MB instead - the downside being lookup in RPCs will need to use the ID), it's a very low hanging fruit - TBC post redesign if it actually helpful at all.

Now, even if we do find ways to tune configuration to avoid point#3 above, it will still be a very inconsistent and not a predictable solution. Hence, we're looking at alternative ways to design this cache distribution. Current proposal:

a. We already maintain active_stake_cache on our ends for stake addresses as long as 2-3 epoch's ago (mostly 2 epochs, except post epoch transition where time taken by dbsync to populate epoch_stake could take a while). This could be used as a baseline for the current `total_balance` field in `grest.stake_distribution_cache`  table, but with further adaptions required as below.
b. We'd need to extract UTxO balance from above as per below:
- adjust previous rewards/MIR(treasury/reserves)/withdrawals (the data is already present as part of overall query, so does not add much to costs) - remove the balance as of 2 epochs ago.
- Add the delta UTxO movements (getting updates on previously used inputs here can be a problem - TBC), from active stake - based on current run on instances exhibiting point#2 and point#3 above, the run for checking utxo map for tx in past 2 epochs was ~5 minutes. Newly registered stake addresses are already part of current filters, they should be retained (not having an entry in active stake marks it at 0)
c. The existing withdrawals, rewards, rewards_available, reserves, treasury, can remain untouched for now, as they dont use up much time.
d. Once this is all done, consider if removing stake_address and pool_id in bech32 format is desirable at all (if not much advantages to remove, we should aim to let current RPCs untouched).

This will likely complicate readability of the query significantly, but we should be able to:
- Remove dependency on TEMP tables
- Improve run-time by getting it to even out amongst instances drastically (current aim would be to get table updates done within 10-20 minutes on machines exhibiting slow behaviour, even if it doesnt improve it by much for machines that dont encounter the slowness).
- Reduce IOPs heavily
- Avoid triggering vacuum limits on every run for default configuration settings.

**Additional Notes:**

- We had treasury, reserves in the cache table but was not used anywhere across grest schema (to verify). Thus, removed the 2 extra joins.
- There was initially temptation to remove `INSERT INTO` due to potential conflict in results and stale entries in the stake cache table (vs truncate/delete/drop table), but doing so will result in a period where stake distribution will return empty, not ideal. Hence, dropping that idea for now.

## Which issue it fixes?
- #1332 

## How has this been tested?
- Verify few stake account entries against output from `cardano-cli` and cardanoscan.io (since it's independent of `cardano-db-sync`)


DamjanOstrelic:
Posting my thinking about the redesigning idea and potential algorithm with questions along the way:

After any given epoch snapshot, there are only 2 ways of changing the balance of an account in the following epoch:
1) transactions
2) rewards distribution (incl. MIR)
(reward withdrawals are not required for the live stake to be updated as rewards are automatically re-staked even if not withdrawn. If there is a withdrawal, the value will be transferred to UTxO)

1.
Get `account_active_stake_cache `amount (`total_balance `at end of last epoch if cache processed, else end of second to last epoch) -> this is the safe baseline mentioned above.

2. (optionally do not run until current epoch cache completed)
Depending on 1., get all newly registered accounts from the last epoch or second to last epoch to 5 blocks behind the current tip (baseline balance for newly registered accounts can be 0)

(what if an account is deregistered? We remove it from cache?)

3.
Scan the transactions for the given delta and sum up for each account.
(I'm not sure how simple the following would be)
```
If account address is input -> negative number
If account address is output -> positive number
```

4.
Add (subtract) summed up delta amounts to baseline. This would not quite be the updated total_balance, as it's missing the possible rewards distributed in the current epoch.
(But what if the rewards have been withdrawn to UTxO? In that case, it is the total_balance as any withdrawal in the current epoch means that no more rewards are available. But what if we are looking 2 epochs back and the withdrawal was in the previous epoch? Then in the current epoch, we still have rewards available to be withdrawn. That's why it is simpler to run live stake update only when active_stake_cache is completed - this would mean around 6 hours lag behind the tip for live stake cache on epoch transition which is perhaps **not acceptable**. But perhaps this is just solvable by looking for AVAILABLE rewards in the next step and whatever we find is what's missing.)

5. Scan available rewards (not withdrawn) for each account and add them to the total balance.

6. We need to differentiate between UTxO and non-UTxO balances to keep the current table structure. So we have 2 options:
	a. Store only the live_stake value (i.e. total_balance) without differentiation.
	b. A good guess at this point of total UTxO value is simply: `UTxO = total_balance - rewards_available`
        (So far I don't see issues with option b.)

7.
I put this as a separate step but in reality, it would probably be handled in step 5. -> make sure we get values for fields in the current table structure: 
`rewards | withdrawals | rewards_available | reserves | treasury`

rdlrt:
> get all newly registered accounts from the last epoch or second to last epoch to 5 blocks behind the current tip (baseline balance for newly registered accounts can be 0) (what if an account is deregistered? We remove it from cache?)

We are already have left join of currently delegated accounts (awdp), that wouldn't change - thereby, we already get updates for newly registered/deregistered accounts

> (optionally do not run until current epoch cache completed)

We should use control_table to check last run of epoch cache, and quit if last epoch run 

> Add (subtract) summed up delta amounts to baseline. This would not quite be the updated total_balance

Yes, as mentioned we'd need to add withdrawal/rewards/MIR for delta indeed.

> points 4-6

I am off the opinion that we + rewards (type=member,treasury, governance) for given epoch, and subtract any withdrawal tx in past 2 epochs - as it currently does (6hr  on mainnet post epoch transition is fine given cardano's implementation, we're anyways gonna have that lag on other stake endpoints accordingly)

6. I would go with option b, as we're already kinda calculating  each (the runtime will likely have almost no addition to provide this data)

DamjanOstrelic:
> We already have left join of currently delegated accounts (awdp), that wouldn't change - thereby, we already get updates for newly registered/deregistered accounts

Still, what about deregistered accounts? If we don't truncate the cache table before inserting, they just don't get touched but still exist in the cache? I think this would cause issues (i.e. wrong live stake values for a pool after summing up delegated balances). This is an issue in current implementation as well if I'm not mistaken.

rdlrt:
> > We already have left join of currently delegated accounts (awdp), that wouldn't change - thereby, we already get updates for newly registered/deregistered accounts
> 
> Still, what about deregistered accounts? If we don't truncate the cache table before inserting, they just don't get touched but still exist in the cache? I think this would cause issues (i.e. wrong live stake values for a pool after summing up delegated balances). This is an issue in current implementation as well if I'm not mistaken.

If you mean the deregistered stake addresses will be leftover, agree - we can/should add a delete row for any accounts from table that are not in awdp (alternatively, simply re-create table - as `most` rows, not all, will anyways get an UPDATE due to rewards). But I think our first step should be get hands on and see performance difference with current implementation, validate our theories wrt performance gains

rdlrt:
Added updates to addresses the updated PR description.

Timings against mainnet instances:
- The result so far is that on mentioned slow machine (where stake distribution cache was taking 20-22 hours due to vacuum trigger), the update finished in ~20 minutes.
- On machine where update stake distribution was taking 40-45 minutes, the update finished in ~10 minute.

Note that in the existing (previous and current) implementation, we did not delete stale accounts - but we ideally should. This has currently not been added, but might potentially add few minutes to execution.

files/grest/rpc/01_cached_tables/stake_distribution_cache.sql
```diff
@@ -1,301 +1,238 @@
...
-SELECT awdp.stake_address,
-  pi.pool_id,
-  CASE
-    WHEN (
-      COALESCE(atrew.REWARDS, 0) - COALESCE(atw.WITHDRAWALS, 0)
-    ) < 0 THEN COALESCE(atu.UTXO, 0) + COALESCE(atrew.REWARDS, 0) - COALESCE(atw.WITHDRAWALS, 0) + COALESCE(atres.RESERVES, 0) + COALESCE(att.TREASURY, 0) - (
-      COALESCE(atrew.REWARDS, 0) - COALESCE(atw.WITHDRAWALS, 0)
+  SELECT block_no FROM PUBLIC.BLOCK
+    WHERE block_no IS NOT NULL
+      AND block_no = (SELECT MAX(BLOCK_NO) FROM PUBLIC.BLOCK) INTO _last_accounted_block_height;
```
DamjanOstrelic, line 20: Is `MAX(BLOCK_NO)` correct here? We may want to look 5 blocks behind the tip?

files/grest/rpc/01_cached_tables/stake_distribution_cache.sql
```diff
@@ -1,301 +1,238 @@
...
+  
+  SELECT (last_value::integer - 2)::integer INTO _active_stake_epoch FROM GREST.CONTROL_TABLE WHERE key = 'last_active_stake_validated_epoch';
```
DamjanOstrelic, line 22: Hmm, if we validated epoch 195 for example, it means we have all the entries in the `active_stake` table for the stake distribution snapshot that was captured at the end of epoch 193 and is active in epoch 195. Just making sure that I got the correct idea:
Our baseline is always `last_active_stake_validated_epoch - 2`?

files/grest/rpc/01_cached_tables/stake_distribution_cache.sql
```diff
@@ -1,301 +1,238 @@
...
+
+  SELECT id INTO _last_active_stake_blockid FROM PUBLIC.BLOCK
+    WHERE epoch_no = _active_stake_epoch
+    ORDER BY block_no DESC LIMIT 1 ;
```
DamjanOstrelic, line 26: A possible edge case here is if a block_no is NULL it would be selected as first when ordering by DESC.

But since it is selected from older epochs, I don't think it can happen. 👍

files/grest/rpc/01_cached_tables/stake_distribution_cache.sql
```diff
@@ -1,301 +1,238 @@
...
+        INNER JOIN tx ON tx.id = tx_out.tx_id
+        LEFT JOIN tx_in ON tx_out.tx_id = tx_in.tx_out_id AND tx_out.index::smallint = tx_in.tx_out_index::smallint
+      WHERE TX.BLOCK_ID > _last_active_stake_blockid
+      GROUP BY awdp.stake_address_id
+    ),
+    account_delta_rewards AS (
+      SELECT awdp.stake_address_id, COALESCE(SUM(reward.amount), 0) AS REWARDS
+      FROM REWARD
+        INNER JOIN accounts_with_delegated_pools awdp ON awdp.stake_address_id = reward.addr_id
+      WHERE REWARD.SPENDABLE_EPOCH >= _active_stake_epoch
```
DamjanOstrelic, line 82: I think we are missing an upper boundary check here that `SPENDABLE_EPOCH` is not greater than `CURRENT_EPOCH`.

files/grest/rpc/01_cached_tables/stake_distribution_cache.sql
rdlrt: Not necessarily, even if block is rolled back and tx is included elsewhere - it would prolly end up to be same value

files/grest/rpc/01_cached_tables/stake_distribution_cache.sql
rdlrt: Yip that's exactly what we'd be looking for as baseline...to use active stake as baseline and then start tracking tx whose block height is greater than tx id of last block of that epoch

files/grest/rpc/01_cached_tables/stake_distribution_cache.sql
rdlrt: Yip, but good point - should avoid in case a bug is encountered on dbsync end

files/grest/rpc/01_cached_tables/stake_distribution_cache.sql
rdlrt: That's interesting, I didn't think `spendable_epoch` for next epoch is recorded by dbsync yet :

``` sql
select from reward where spendable_epoch > (select max(no) from epoch) limit 10;
# --
# (0 rows)
```

DamjanOstrelic: From mainnet DB just now:
```
cexplorer=# select max(no) from epoch;
 max
-----
 329
(1 row)

cexplorer=# select max(spendable_epoch) from reward;
 max
-----
 330
(1 row)
```

rdlrt: Cool, wasnt aware of this one - it's a good outcome then

Scitz0 (approved):
👍  great work guys

DamjanOstrelic (approved):
🥇

Issue #1379, opened 2022-03-21
prereqs.sh: Improve addition to .bashrc


HeptaSean:
The code used to add `export CNODE_HOME` to `.bashrc` in `prereqs.sh` is not defensive enough:
https://github.com/cardano-community/guild-operators/blob/alpha/scripts/cnode-helper-scripts/prereqs.sh#L290-L296

If `.bashrc` does not end in a newline, it will add at the end of the last line and not on a new line.

We had one user in the forum who was bitten by this:
https://forum.cardano.org/t/how-to-set-up-a-pool-in-a-few-minutes-and-register-using-cntools/48767/1748
Apparently, he did not even cause it himself, but `ghcup` (which is installed from `prereqs.sh` directly before that) does its additions without final newline (in some circumstances?).

The script should either check for final newline or unconditionally insert one or more blanks before starting modifying the `.bashrc`.

Moreover, the check if the export is already there only looks for the string `CNODE_HOME`. If it is in a comment or even only used, but not exported, the script will think it's already there. (Similarly applies to the `LD_LIBRARY_PATH` check in the next paragraph.)

A check at least for `^export CNODE_HOME` would be more resilient.

HeptaSean:
Looks good to me! Thanks for the fast fix!

Grepping for the whole export also looks good. You are right that the script should respect a user commenting that out explicitly.

PR #1380, opened 2022-03-21
[prereqs.sh] - Handle new line for export of CNODE_HOME on Darwin

rdlrt:
Address comments from #1379 , to address ghcup installer change on Darwin under `adjust_bashrc`


Scitz0 (approved):
👍

PR #1385, opened 2022-03-27
fix(cncli): replace hardcoded cncliSync host

mkungla:
Use env $PT_HOST from User Variables instead of hardcoded value.
defaults to 127.0.0.1

### DIFF
--- a/scripts/cnode-helper-scripts/cncli.sh
+++ b/scripts/cnode-helper-scripts/cncli.sh
@@ -207,7 +207,7 @@ cncliInit() {
 #################################
 
 cncliSync() {
-  ${CNCLI} sync --host 127.0.0.1 --network-magic "${NWMAGIC}" --port "${CNODE_PORT}" --db "${CNCLI_DB}"
+  ${CNCLI} sync --host "${PT_HOST}" --network-magic "${NWMAGIC}" --port "${CNODE_PORT}" --db "${CNCLI_DB}"
 }
 
 #################################


rdlrt (approved):

Scitz0:
I'm not sure I agree with this change to be honest. PT_HOST was added specifically for pooltool sendtip to shield bp and allowing one to post data from any relay to pooltool. 

Syncing cncli db from anything other than localhost will be extremely slow. At least that was my experience way back.

If we want to make this configurable it should at least get its own setting and not mix it with pooltool.

Scitz0:
If you want to run cncli wrapper script on a relay to check leaderlogs etc. you can just deploy cncli services like normal but hardcode values for POOL_ID, POOL_VRF_SKEY and POOL_VRF_VKEY.

mkungla:
@Scitz0 sorry if I messed up something with this change I'm new to **guild-operators** scripts. I proposed that change since I encountered issue having hard coded `127.0.0.1` when deploying  `cnode--cncli-sync.service` with `deploy-as-systemd.sh`, but now I understand your point that `PT_HOST` is perhaps not right one to use here.

My problem was since system service calls this

https://github.com/cardano-community/guild-operators/blob/a8de40f351b8aa40b707ef643d56ff0ab11b36d8/scripts/cnode-helper-scripts/deploy-as-systemd.sh#L168

so `cncli` can not call my local node via 127.0.0.1 loopback because it listens specific interface e.g. `10.0.0.2` and `fe80::8400....` set here
https://github.com/cardano-community/guild-operators/blob/a8de40f351b8aa40b707ef643d56ff0ab11b36d8/scripts/cnode-helper-scripts/cnode.sh#L14-L15

Scitz0:
I see, I talked it over with priyank and we will create a new variable for this. It has to be made clear in description for variable though that its not recommended to use that for remote syncing.

PR #1388, opened 2022-03-29
[RPC] Cast metadata keys to strings

DamjanOstrelic:
## Description

Endpoints returning metadata keys will now return their values as strings instead of integers.

Associated schema update PR: https://github.com/cardano-community/koios-artifacts/pull/19

## Motivation and context

Easier client-side handling.

## How has this been tested?

Check that the changed endpoint returns the key value as a string.


Scitz0 (approved):
👍

PR #1389, opened 2022-03-29
Add _pool_bech32_id argument for /rpc/pool_metadata endpoint

rcmorano:
## Description

I've added support for pool filtering while retrieving metadata [in the same way](https://github.com/cardano-community/guild-operators/blob/alpha/files/grest/rpc/pool/pool_info.sql#L1) it's implemented in the `/rpc/pool_info` endpoint

## Motivation and context

I thought it would be handy :)

### DIFF
--- a/files/grest/rpc/pool/pool_metadata.sql
+++ b/files/grest/rpc/pool/pool_metadata.sql
@@ -1,4 +1,4 @@
-CREATE FUNCTION grest.pool_metadata ()
+CREATE FUNCTION grest.pool_metadata (_pool_bech32_ids text[])
     RETURNS TABLE (
         pool_id_bech32 character varying,
         meta_url character varying,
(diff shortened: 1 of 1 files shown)


rdlrt:
Thanks, I agree it's useful - but we'd probably want to make it optional (in case of `pool_info` it has joins on other tables making it much more expensive  to instance, as compared to `pool_metadata` which is single join).
Not having that param allows viewing all metadata - for URL/tickers/etc, useful for checking clusters and so on

rcmorano:
Ah sorry, it was late here and it seems I was sleepy :sweat_smile: I tried w/o any args and it actually did work for me because I didn't drop the oirignal function hehe  

I've added changes so it keeps original behaviour when no arg is passed :)

DamjanOstrelic (approved):
LGTM, I think it's a welcome change as well. Unless someone has a different opinion, we can probably roll it out this Saturday. 👍 

It will need an API spec update on the [koios-artifacts repo](https://github.com/cardano-community/koios-artifacts), but I can update that part once this is approved.

Scitz0 (approved):
Looks fine to me

PR #1391, opened 2022-04-01
Bump shutdown time for cnode.sh to 60s

rdlrt:
There is not scientifically proven theory around that shutdown in 5s causes node to not go down clean enough, and 30s works better (personally, I've found that when at times doing SIGINT at terminal , and letting it take it's time still resulted in wierd shutdown state, but assuming the experience of others is true, 5s is possibly too short)


hodlonaut (approved):
Looks okay

PR #1393, opened 2022-04-03
Minor patch: Replace 'select max(..) from ..' calls with constants 

rdlrt:
Replace 'select max(..) from ..' calls with constants  as they're repeated for every record.

The change is primarily to reduce hits in pg_stats for epoch table - but also slightly improves performance further, as slower instance timing reduced from ~25m to ~17m, while faster one reduced from ~10m to ~8m on mainnet) than query itself.

The change remains compatible (no change to values/query itself) with previous commit


Scitz0 (approved):
Looks good

PR #1394, opened 2022-04-04
Fix issues from last commit, and adjust epoch boundary conditions

rdlrt:
Stake Distribution updates:
- There was an ordering issue in previous commit
- Active Stake distribution for delta rewards to use _active_stake_epoch as baseline and skip next 2 epochs (was causing duplicate reward)
- Remove unnecessary join on tx_in for output

PS: Due to reasons I cannot fathom, use of `_latest_epoch` on delta rewards causes query to run exponentially longer. Until we figure out why, used subquery for the same with added comment


DamjanOstrelic (approved):

Scitz0 (commented):
👍

PR #1399, opened 2022-04-11
grest [sdc]: Certain stake accounts were not updated

rdlrt:
I am not fully sure what role the removed clause below plays, but having it prevents updates to existing account tables - preventing accurate results to be inserted to sdc table

- stake_dist_cache: Remove redundant(?) WHERE clause
- Order to prefer pmr_id over registered tx_id in pool_info

(Past week few users reported the issue where their stake was not being updated, we were not able to replicate it before since we re-created the stake distribution cache table as part of previous change)


Scitz0 (approved):
I guess this is best approved by those involved in creating the code to begin with, but from my perspective, I cant see why these filters need to be there.

PR #1400, opened 2022-04-20
[SDC] Update job for newly registered accounts

DamjanOstrelic:
## Description

Since changing our approach to handling the stake distribution cache in [this PR](https://github.com/cardano-community/guild-operators/pull/1378) to use the active_stake as the baseline for calculations, the leftover edge case is newly registered accounts. These accounts do not have an entry in the `epoch_stake` table and therefore, the delta calculated for them as their current `live_stake` ends up being negative. This self-heals with time for a given account but takes 2 epochs to get an entry to the `epoch_stake` table.

Check that the code works correctly, and consider possible edge cases if any. There is a comment in the code regarding the `epoch_no` rule for the two jobs to follow:
```
          -- Account must be present in epoch_stake table for the last validated epoch
          -- Up for discussion whether this should be switched to _latest_epoch
```

Consider whether we need any `grest.control_table` entries for this new job.

## Motivation and context

The goal of this PR is to create a new stake distribution update job exclusive to these newly registered accounts, as well as exclude these accounts from the primary job that uses the `epoch_stake` table.

## How has this been tested?

Every commit has been gradually tested on the guild network. Cases covered include:
 - checking the primary job does not pick up a newly registered account
 - checking the newly added job only picks up the newly registered account(s)
 - checking the newly added job does not pick up a newly registered account after 2 epochs when the `epoch_stake` table has an entry for it
 - checking that the newly added job inserts the correct balance values for a newly registered account
 - checking that the newly added job correctly updates existing balance values for an already inserted account (but still without an `epoch_stake` entry)
 - checking that the newly added job kills the SDC job if it's running
 - check that SDC job decides to exit if new accounts job is running


rdlrt (commented):
Havent taken it for test, but need to add check against pg_stat_activity for both SQLs to prevent running simultaneously (would prefer to defer to next run when that happens)

PR #1403, opened 2022-04-26
Add setup-grest versioning

DamjanOstrelic:
## Description

Adding grest versioning to track instance updates. Version is only logged/updated if `setup-grest.sh -q` flag is used. Additionally, if the reset flag `-r` is used, it is logged as an `artifact`. This can be later extended to log other flags being used or similar.

`grest-poll.sh` gets an update to check the instance grest version before further checks.

Small update for SDC SQL to not try create the table if already exists (avoids errors during `-q`-only deploys).
Small update to `grest.update_control_table()` to make sure the `artifacts` field is updated on conflict.


rdlrt (approved):
LGTM, only to be merged post Saturday 8am UTC (to prevent hourly update of grest-poll checking for version atm)

PR #1405, opened 2022-04-29
Multiple polishing updates for grest pre version release:

rdlrt:
- Update haproxy config: Split grest_core into postgrest and failover backends. While this keeps queries compatible with grest_core, it will allow changes upstream on haproxy config of monitoring backends
- Bump prometheus/grafana versions (relevant only for monitoring instances)
- Add upcoming `/ready` endpoint (postgREST internal status endpoint) status check for grest-poll.sh script, absence of the endpoint in v9 for now does not impact availability, as we check the returned header

### DIFF
--- a/scripts/grest-helper-scripts/grest-poll.sh
+++ b/scripts/grest-helper-scripts/grest-poll.sh
@@ -59,7 +59,7 @@ function chk_upd() {
 }
 
 function log_err() {
-	echo "$(date +%DT%T)	- ERROR: " "$@" >> "${LOG_DIR}"/grest-poll.sh_"$(date +%d%m%y)"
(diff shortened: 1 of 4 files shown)


scripts/grest-helper-scripts/grest-poll.sh
```diff
@@ -72,6 +72,14 @@ function usage() {
   exit 1
 }
 
+function is_up() {
```
DamjanOstrelic, line 75: Function name mismatch:
```suggestion
function chk_is_up() {
```

DamjanOstrelic (approved):
After fixing the function name, all looks good on my guild instance with new setup

PR #1414, opened 2022-05-27
gRest BugFix: /pool_info => wrong metadata sort

rdlrt:
## Description

1. Join for pool_offline_data and pool_info_cache based on hash_id match was slightly vague, as you can end up having a mismatch of meta and pmr ID (can be troublesome later for ordering)
2. ordering of meta_id should be descending

The above could result in invalid metadata for those who may have updated their pool metadata


Scitz0 (approved):
👍

DamjanOstrelic (approved):

PR #1415, opened 2022-05-28
[gRest - EIC] Add rewards data to epoch cache and /epoch_info

DamjanOstrelic:
## Description

Adding 2 columns to epoch info cache:
1. total rewards earned for the epoch
2. average block reward for the epoch

Since it's a cache table change, to test it's best to manually delete the `grest.epoch_info_cache` table and run the changed SQL. That will produce a new table with the 2 extra fields. The population can be then done with the cron script `epoch-info-cache-update.sh` or with SQL `SELECT  grest.epoch_info_cache_update(0);`.

Check the values for the two new fields are correct after the initial population. The current and previous epochs (last 2) should have `NULL` values until updated in further epochs.
Check that after the epoch switch to epoch `N`, the update job fills epoch `N - 2` for the new columns.

Check that the changed endpoint `epoch_info` works with new fields.

Closes cardano-community/koios-artifacts#22


files/grest/rpc/01_cached_tables/epoch_info_cache.sql
```diff
@@ -155,6 +168,17 @@ BEGIN
...
+          SUM(r.amount) as amount
+        FROM
+          reward r
+        WHERE
+          r.earned_epoch = e.no
+        GROUP BY
+          e.no
+      ) reward_pot ON TRUE
     WHERE
       e.no >= _epoch_no_to_insert_from
```
Scitz0, line 183: Joins in general aren't executed in any particular order as they are ordered in code. But... lateral joins always execute last as they loop over all results from previous joins. Where filtering is then applied to the final result, this is at least my understanding of it. 

To reduce rows lateral join has to loop over, you could move the where filter to the inner join on block. This would force it to filter out these results before lateral join is applied. I don't know if the query planner is smart enough to figure this out by itself or what the actual gain might be. But could be worth testing. 

So inner join would then be:
`INNER JOIN block b ON b.time = e.start_time AND e.no >= _epoch_no_to_insert_from`
and WHERE filter can be removed.

DamjanOstrelic: Thanks, good point. I will give it a try

rdlrt (approved):

PR #1432, opened 2022-06-10
Update CNODE_HOME to CNODE_PORT in kill command for cnode.sh

rdlrt:
## Description
In cases where folks may have more than one node running under CNODE_HOME, having that variable to kill would not be appropriate. Instead , using CNODE_PORT ensures that the PID that comes back is unique (as you should - under no circumstances - be ending up with two cardano-node processes sharing same port)


hodlonaut (approved):
Looks good to me.

PR #1433, opened 2022-06-13
Updates for node-1.35.0

rdlrt:
Placeholder branch/PR where updates to scripts will be made once release is finalised (later than `1.35.0-rc2`)

### DIFF
--- a/scripts/cnode-helper-scripts/env
+++ b/scripts/cnode-helper-scripts/env
@@ -973,8 +973,8 @@ if [[ -n "${CNODE_PID}" ]]; then
 fi
 
 node_version="$(${CCLI} version | head -1 | cut -d' ' -f2)"
-if ! versionCheck "1.32.1" "${node_version}"; then
-  echo -e "\nGuild scripts has now been upgraded to support cardano-node 1.32.1 or higher (${node_version} found).\nPlease update cardano-node (note that you should ideally update your config too) or use tagged branches for older node version.\n\n"
+if ! versionCheck "1.35.0" "${node_version}"; then
+  echo -e "\nGuild scripts has now been upgraded to support cardano-node 1.35.0 or higher (${node_version} found).\nPlease update cardano-node (note that you should ideally update your config too) or use tagged branches for older node version.\n\n"
   return 1
 fi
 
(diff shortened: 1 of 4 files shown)


scripts/cnode-helper-scripts/env
```diff
@@ -882,7 +882,7 @@ if ! command -v "jq" &>/dev/null; then
   return 1
 fi
 
-read -ra CONFIG_CONTENTS <<<"$(jq -r '[ .AlonzoGenesisFile, .ByronGenesisFile, .ShelleyGenesisFile, .Protocol, .TraceChainDb]| @tsv' "${CONFIG}" 2>/dev/null)"
+read -ra CONFIG_CONTENTS <<<"$(jq -r '[ .AlonzoGenesisFile, .ByronGenesisFile, .ShelleyGenesisFile, .Protocol, .TraceChainDb, .EnableP2P]| @tsv' "${CONFIG}" 2>/dev/null)"
```
Scitz0, line 885: This is not necessarily a problem with this specific change but jq and tsv in general. If the requested field doesn't exist, the output is empty. And because of this, addressing a specific entry in the resulting array may not get you the value you want. In this example, if EnableP2P doesn't exist in config, the array will only contain 5 entries. 
`P2P_ENABLED="${CONFIG_CONTENTS[5]:=false}"` will work anyway and set value to false as its the last entry. But lets say a 7th metric is grabbed or if some of the previous in list is missing, then it will be a mess with the result. 

So... the solution is to always add a default value in jq call using for example `//"null"` to return the string null if empty.
`read -ra CONFIG_CONTENTS <<<"$(jq -r '[ .AlonzoGenesisFile //"null", .ByronGenesisFile //"null", .ShelleyGenesisFile //"null", .Protocol //"null", .TraceChainDb //"null", .EnableP2P //"null"]| @tsv' "${CONFIG}" 2>/dev/null)"`

All following references to CONFIG_CONTENTS array would then need to be updated to instead of doing -z check for empty, do something like this:
`[[ -z ${CONFIG_CONTENTS[0]} || ${CONFIG_CONTENTS[0]} = null ]]`

gufmar: makes sense
added a commit to have a default for each config key (usually null)

`.Protocol` defaults to "Cardano"
`.EnableP2P` to false if absent

It looks like all "following references" are covered in the couple of lines after the jq call. 
All genesis files are checked against -f (ile exists) which correctly fails for null
And `.TraceChainDb` is now exclusively checking for true setting

PR #1434, opened 2022-06-16
Use declared variable instead of inline subquery for latest epoch in sdc

Scitz0:


rdlrt (approved):
LGTM, tho we're still not sure of the cause (the previous timings were exact reverse of current)

PR #1435, opened 2022-06-18
Fix for pool_info described in issue 47(koios-artifacts)

Scitz0:
## Description
Solved by:
- removing meta_id ordering
- fallback on offline data json in case of missing entry for meta_id to grab next non null entry
- code refactored to use a CTE. The idea is that all lateral joins will have fewer entries to loop over by this approach if there are pools with many updates.

## Which issue it fixes?
https://github.com/cardano-community/koios-artifacts/issues/47

## How has this been tested?
verified with a few pools on mainnet and testnet but further tests needed.

### DIFF
--- a/files/grest/rpc/pool/pool_info.sql
+++ b/files/grest/rpc/pool/pool_info.sql
@@ -31,7 +31,7 @@ DECLARE
   _epoch_no bigint;
   _saturation_limit bigint;
 BEGIN
-  SELECT epoch.no INTO _epoch_no FROM public.epoch ORDER BY epoch.no DESC LIMIT 1;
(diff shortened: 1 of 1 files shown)


DamjanOstrelic:
Just a quick note, to reference an issue on another repo (since we are working with the issues from koios-artifacts repo here):

cardano-community/koios-artifacts#47

DamjanOstrelic (approved):
LGTM 👍 Not sure if you have any other test cases lined up for consideration still

Scitz0:
> cardano-community/koios-artifacts#47

Right, good point 👍 
Will update to remove any cross reference to wrong issue.

rdlrt (approved):

PR #1436, opened 2022-06-20
Bump haproxy version

rdlrt:
- Bump Haproxy Version to 2.6
- Make use of systemd reload optimisations (hitless)
- Temporarily disable check for `/ready` endpoint from PostgREST, as it needs a seperate admin server port as well as version `9.0.1`


DamjanOstrelic (approved):
LGTM 👍

PR #1440, opened 2022-06-26
Add consensus flag to CNCLI leaderlog command

rdlrt:
CNCLI v5 added a consensus flag to support breaking changes from node-1.35

While this is already being handled in #1433 , for mainnet (where node to be used is 1.34.x) - the change needs to be added prior to that PR. Hence, cherry-picked the commit.

This will likely require a `cncli.sh leaderlog force`


Scitz0 (approved):
👍

PR #1442, opened 2022-06-28
Fix leaderlogs not running for next epoch

reqlez:
## Description
Changed praos and tpraos consensus variables to remove "--consensus" flag.
Added --consensus flags to cncli leaderlogs commands.

## How has this been tested?
Tested running leaderlogs for current and next epoch, okay!


Scitz0 (approved):
Thanks for the fix 👍

PR #1446, opened 2022-07-01
Fix dbtip to read UNIX timestamp (v1.0.1) and human readable format (> 1.0.5)

### LINKED ISSUE #1445: grest updates on alpha for dbsync 13 broke haproxy backend grest_postgrest on all current koios installations with dbsync 12
Since https://github.com/cardano-community/guild-operators/commit/07f0ca4daaec588d16206ee9b5bfd9dc1024260b (about 2 hours ago) went live on alpha channel all our local koios installations report that the haproxy backend "grest_postgrest" is down.
I think it has to do with the servers downloading grest-poll.sh regularly from the repo and saving it locally.
Looks like the new grest-poll.sh is not compatible with an current installation of dbsync 12.
How to fix that?

rdlrt:
Closes #1445


DamjanOstrelic (approved):
LGTM

PR #1451, opened 2022-07-04
Lighten grest.account_active_stake_cache

rdlrt:
## Description

- [x] Update grest.account_history to remove dependency on grest.account_active_stake_cache
- [x] Retire minimalistic (single-table) functions used in active_stake_cache SQL that are only used once
- [x] Delete historical view (beyond 4 epochs) for grest.account_active_stake_cache, in most cases - only 3 is sufficient, except on a new sync instance when epoch_stake will not be populated and you may need 4th
- [x] Minor bug-fixes in grest-poll.sh (echo for dbtip/currtip logging) and dbsync.sh (ensure \${CONFIG}.tmp file is not empty)

Closes cardano-community/koios-artifacts#52


Scitz0 (approved):
Looks good but haven't tested the changes myself.

PR #1452, opened 2022-07-04
[ASC] Build last epoch Active stake cache

dostrelith678:
## Description

Initial work for capturing stake snapshots of two previous epochs. The stake snapshot at the end of the **_previous epoch_** will become active in the **_next epoch_** (skipping the **_current epoch_**). This means that shortly after transitioning into the current epoch, we can already provide data of the active stake snapshot for the next epoch.

This also means we don't have to wait for db-sync to populate **_current epoch_** values after transition (~6 hours on mainnet). At the moment, we are fully relying on db-sync stake snapshot data, which is gradually populated and until it is, our API is not fully functional as some endpoints depend on that data.

This PR includes a setup of a new table to hold the snapshot data for the last two epochs (anything older than that gets deleted as it's already available in `epoch_stake` table) and a function that calculates the stake snapshot at the end of the previous epoch and stores it. It does in the same principle as the SDC job, which is to use `epoch_stake` table as a baseline for account balances and calculate the delta of rewards, inputs and outputs from the baseline to the end of the previous epoch (while the SDC job upper boundary is the current tip).

I suggest manually creating the table and creating the procedure `GREST.CAPTURE_LAST_EPOCH_SNAPSHOT`.
Simply running `CALL GREST.CAPTURE_LAST_EPOCH_SNAPSHOT()` will populate the new table and this should then be compared with the actual snapshot db-sync will populate once it gets there.
If we are in epoch `n`, then this function will capture the state at the end of epoch `n-1`. This snapshot should become active in epoch `n - 1 + 2` or `n + 1`, meaning it should be compared with the snapshot in table `epoch_stake where epoch_no = n + 1` once populated.

This means that you'd have to wait for the next epoch to come along and db-sync to populate the values.

For testing, you can instead specify the epoch you want to calculate the snapshot for by making a slight change to `line 216`:
`SELECT MAX(NO) - 1 INTO _previous_epoch_no FROM PUBLIC.EPOCH;`
->
`SELECT 4963 INTO _previous_epoch_no FROM PUBLIC.EPOCH;`
You can then compare the result with `epoch_stake where epoch_no = 4965`.

In this case, some premade `SELECT`s for comparison is:

Compare counts of snapshots:
```
SELECT COUNT(*) FROM EPOCH_STAKE WHERE EPOCH_NO = 216;
SELECT COUNT(*) FROM grest.STAKE_SNAPSHOT_CACHE WHERE EPOCH_NO = 214;
```

Compare total active stake of snapshots:
```
SELECT SUM(amount) FROM EPOCH_STAKE WHERE EPOCH_NO = 216;
SELECT SUM(total_balance) FROM grest.STAKE_SNAPSHOT_CACHE WHERE EPOCH_NO = 214;
```

Get all entries that have different values (should be 0 results):
```
select sa.view, amount, total_balance, amount - total_balance as diff
from grest.stake_snapshot_cache ssc
  inner join stake_address sa ON sa.view = ssc.stake_address
    inner join epoch_stake es ON es.addr_id = sa.id
where es.epoch_no = 216
    AND ssc.epoch_no = 214
    AND amount != total_balance;
```

To-do:
- ~handle deleting old epochs data~
- test MIR reward correctly captured in snapshot
- ~test tx input/output correctly captured in snapshot~
- ~test refund reward correctly captured~
- ~test withdrawals correctly subtracted~
- ~test newly delegated accounts captured correctly~

## Which issue it fixes?

Closes cardano-community/koios-artifacts#49
## How has this been tested?

Below is example code used to test epochs on guild and testnet network. Testnet epochs from 74 to 215 have been tested. Guild epochs from 74 to 3290.
[reg-test.zip](https://github.com/cardano-community/guild-operators/files/9126529/reg-test.zip)


files/grest/rpc/01_cached_tables/active_stake_cache.sql
```diff
@@ -22,6 +19,14 @@ CREATE TABLE IF NOT EXISTS GREST.ACCOUNT_ACTIVE_STAKE_CACHE (
   PRIMARY KEY (STAKE_ADDRESS, POOL_ID, EPOCH_NO)
 );
 
+CREATE TABLE IF NOT EXISTS GREST.LAST_TWO_EPOCHS_ACCOUNT_ACTIVE_STAKE_CACHE (
```
rdlrt, line 22: OK to rename it to something smaller? Essentially what we're getting here is go and set snapshots (artificially), maybe stake_snapshot_cache?

files/grest/rpc/01_cached_tables/stake_distribution_cache.sql
```diff
@@ -28,7 +28,7 @@ BEGIN
       AND block_no IS NOT NULL
     ORDER BY block_no DESC LIMIT 1;
 
-  SELECT MAX(id) INTO _last_account_tx_id FROM PUBLIC.TX WHERE block_id = _last_active_stake_blockid; 
+  SELECT MAX(id) INTO _last_account_tx_id FROM PUBLIC.TX WHERE block_id <= _last_active_stake_blockid;
```
dostrelith678, line 31: This is for an edge case where the last block of the epoch has no transactions and the variable `_last_account_tx_id` becomes `NULL`. However, adding `<=` has worse performance. Could be better optimised.

files/grest/rpc/01_cached_tables/stake_snapshot_cache.sql
```diff
@@ -0,0 +1,445 @@
...
+          FROM pool_update pu
+          WHERE pu.hash_id = pr.hash_id 
+            AND (
+              pu.registered_tx_id > pr.announced_tx_id
+                OR (
+                  pu.registered_tx_id = pr.announced_tx_id
+                    AND pu.cert_index > pr.cert_index
+                )
+            )
+            AND registered_tx_id <= _upper_bound_account_tx_id
```
dostrelith678, line 87: missing `pu.`

files/grest/rpc/01_cached_tables/stake_snapshot_cache.sql
```diff
@@ -0,0 +1,445 @@
...
+
+  SELECT MAX(id) INTO _lower_bound_account_tx_id FROM PUBLIC.TX
+    WHERE block_id <= _last_active_stake_block_id;
+
+  SELECT MAX(id) INTO _upper_bound_account_tx_id FROM PUBLIC.TX
+    WHERE block_id <= _previous_epoch_last_block_id;
+
+/* Registered and delegated accounts to be captured (have epoch_stake entries for baseline) */
+  WITH
+    latest_non_cancelled_pool_retire as (
```
dostrelith678, line 68: Both `latest_non_cancelled_pool_retire`, `minimum_pool_delegation_tx_ids` CTEs could explore using `GROUP BY` instead of `SELECT DISTINCT`.

files/grest/rpc/01_cached_tables/stake_snapshot_cache.sql
```diff
@@ -0,0 +1,404 @@
+/* Keeps track of stake snapshots taken at the end of epochs n - 1 and n - 2 */
+DROP TABLE IF EXISTS GREST.stake_snapshot_cache;
+
+CREATE TABLE GREST.stake_snapshot_cache (
+  addr_id integer,
+  STAKE_ADDRESS varchar,
+  POOL_ID varchar,
+  AMOUNT numeric,
+  EPOCH_NO bigint,
+  PRIMARY KEY (addr_id, EPOCH_NO)
```
dostrelith678, line 10: Change to primary key (addr_id) and UNIQUE constraint (addr_id, epoch_no)

files/grest/rpc/01_cached_tables/stake_snapshot_cache.sql
```diff
@@ -0,0 +1,404 @@
...
+    
+    CREATE INDEX _idx_pool_hash_id ON latest_accounts_delegation_txs (pool_hash_id);
+
+
+/* Registered and delegated accounts to be captured (have epoch_stake entries for baseline) */
+  WITH
+    accounts_with_delegated_pools AS (
+      SELECT DISTINCT ON (STAKE_ADDRESS.ID)
+        stake_address.id as stake_address_id,
+        stake_address.view as stake_address,
```
dostrelith678, line 210: remove this column

files/grest/rpc/01_cached_tables/stake_snapshot_cache.sql
```diff
@@ -0,0 +1,404 @@
+/* Keeps track of stake snapshots taken at the end of epochs n - 1 and n - 2 */
+DROP TABLE IF EXISTS GREST.stake_snapshot_cache;
+
+CREATE TABLE GREST.stake_snapshot_cache (
+  addr_id integer,
+  STAKE_ADDRESS varchar,
+  POOL_ID varchar,
```
dostrelith678, line 7: change to integer pool ID

files/grest/rpc/01_cached_tables/stake_snapshot_cache.sql
```diff
@@ -0,0 +1,404 @@
...
+            )
+        )
+        -- Account must be present in epoch_stake table for the previous epoch
+        AND EXISTS (
+          SELECT TRUE FROM EPOCH_STAKE
+            WHERE EPOCH_STAKE.EPOCH_NO = _previous_epoch_no
+            AND EPOCH_STAKE.ADDR_ID = STAKE_ADDRESS.ID
+        )
+    ),
+    pool_ids as (
```
dostrelith678, line 230: remove this as bech32 not needed.

files/grest/rpc/01_cached_tables/stake_snapshot_cache.sql
```diff
@@ -0,0 +1,404 @@
...
+        INNER JOIN accounts_with_delegated_pools ON accounts_with_delegated_pools.stake_address_id = withdrawal.addr_id
+      WHERE withdrawal.tx_id > _lower_bound_account_tx_id
+        AND withdrawal.tx_id <= _upper_bound_account_tx_id
+      GROUP BY accounts_with_delegated_pools.stake_address_id
+    )
+
+      INSERT INTO GREST.stake_snapshot_cache
+        SELECT
+          awdp.stake_address_id as addr_id,
+          awdp.stake_address,
```
dostrelith678, line 287: remove this column

files/grest/rpc/01_cached_tables/stake_snapshot_cache.sql
```diff
@@ -0,0 +1,404 @@
...
+          pi.pool_id,
```
dostrelith678, line 288: replace with `awdp.pool_hash_id`

rdlrt (approved):

PR #1454, opened 2022-07-06
Active epoch no to pool_delegators

Scitz0:
See koios-artificacs:
closes https://github.com/cardano-community/koios-artifacts/issues/58


dostrelith678 (approved):
Just need to remove the index addition as `stake_address` is already primary key.

PR #1456, opened 2022-07-07
Updates to cnode.sh startup script and systemd

rdlrt:
## Description

Updates to cnode.sh startup script and systemd
- Add -s to cnode.sh to handle shutdown via SIGINT
- Update systemd creation script to not `cnode.sh -s` for shutdown to add burst limit for startup attempts, remove stdout/stderr redirection.

(PS: This will not resolve/workaround bug from 1.35.0 for incorrect shutdown)

### DIFF
--- a/scripts/cnode-helper-scripts/cnode.sh
+++ b/scripts/cnode-helper-scripts/cnode.sh
@@ -29,6 +29,7 @@ usage() {
 		
 		Cardano Node wrapper script !!
 		-d    Deploy cnode as a systemd service
+		-s    Stop cnode using SIGINT
 		
 		EOF
   exit 1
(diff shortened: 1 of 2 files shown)


Scitz0 (approved):
👍

reqlez:
I was just asked to update this file on my testnet pool, now gLiveView 1.27.0 showing this for time, just heads up:

Uptime: -5:-54:001

reqlez:
1657216196
1657211050

above 2 lines.

Scitz0:
So this result looks alright to me, the first value is subtracted by the second which is lower, and the result is the time in seconds the process has been up.

`uptimes=$(( $(date +%s) - $(date -d "$(TZ=UTC ps -p ${CNODE_PID} -o lstart | tail -1)" +%s) ))`

5146s, so 1h ~25m when you ran the above command. I don't see how this change could cause an issue on your system.

reqlez:
I get it... but... 
image

reqlez:
So when i run this functions i get:
image

PR #1473, opened 2022-08-03
Temp table snapshot

dostrelith678:


rdlrt (commented):
The commit between the 2 branches seem to overlap a lot...wonder if that's because it was created from base branch itself, since the file being modified is just 1 - might be better to rebase onto grest-last-epochs-stake-snapshot branch

rdlrt (approved):
As discussed during call

Issue #1470, opened 2022-08-02
504 Gateway Time-out for Tx Metalabels Request


edridudi:
**Describe the bug**
Tx Metalabels Requests are timing out (infrequently)

**To Reproduce**
GET https://api.koios.rest/api/v0/tx_metalabels?limit=10

**Expected behavior**
No Timeout

**Screenshots**
```
Aug 02, 2022 11:59:12 AM okhttp3.internal.platform.Platform log
INFO: --> GET https://api.koios.rest/api/v0/tx_metalabels?limit=10
Aug 02, 2022 11:59:12 AM okhttp3.internal.platform.Platform log
INFO: --> END GET
Aug 02, 2022 11:59:45 AM okhttp3.internal.platform.Platform log
INFO: <-- 504 Gateway Time-out https://api.koios.rest/api/v0/tx_metalabels?limit=10 (33376ms)
Aug 02, 2022 11:59:45 AM okhttp3.internal.platform.Platform log
INFO: content-length: 92
Aug 02, 2022 11:59:45 AM okhttp3.internal.platform.Platform log
INFO: cache-control: no-cache
Aug 02, 2022 11:59:45 AM okhttp3.internal.platform.Platform log
INFO: content-type: text/html
Aug 02, 2022 11:59:45 AM okhttp3.internal.platform.Platform log
INFO: 
Aug 02, 2022 11:59:45 AM okhttp3.internal.platform.Platform log
INFO: <html><body><h1>504 Gateway Time-out</h1>
The server didn't respond in time.
</body></html>

Aug 02, 2022 11:59:45 AM okhttp3.internal.platform.Platform log
INFO: <-- END HTTP (92-byte body)
```

rdlrt:
The current query code (view for select distinct) is not really optimal, we'd want to perhaps improve the `tx_metalabels` endpoint by moving into a cache table instead - the result set being as small as it currently is. The keys returned are not really changing often, so it might be alright for this table to get refreshed every 10/15 minutes

@dostrelith678 @Scitz0 - thoughts?

Scitz0:
Agree, that the current view is not great as it is. I suggest turning the current view into a materialized view (or a cache table)  that is updated on a schedule. If a materialized view is used, we should refresh it using `CONCURRENTLY` option to allow select on it while updating.

Scitz0:
Actually, let me investigate loost index scan first before creating a cache table.
[...]

PR #1474, opened 2022-08-04
Move tx_metalabels to use loose indexscan

rdlrt:
## Description

As found/suggested by @Scitz0 in #1470 , tx_metalabels view was not usable anymore as on some instances it took ~14-32s. Using the suggestion, it's down to ~10-15ms (only difference being it's now pre-sorted)


Scitz0 (approved):
👍

PR #1486, opened 2022-08-11
Separate pool delegators endpoints

dostrelith678:
## Description

Separating between live and active pool delegators as per cardano-community/koios-artifacts#38:

`/pool_delegators` -> returning LIVE delegators and INCLUDING the latest delegation tx for each address in the response.
`/pool_delegators_history` -> returning ACTIVE delegators for a specified epoch (current if not specified) and NOT INCLUDING any delegation txs in the response.

We could think about fallback to `stake_snapshot_cache` if the specified epoch is `> current_epoch`, although this might make the endpoint name `_history` seem a bit misplaced.
We could also consider checking whether `epoch_stake` was populated (`last_validated_active_stake_epoch` rather) for `current_epoch`, and if not, read from the `stake_snapshot_cache`.

## Which issue it fixes?

Closes cardano-community/koios-artifacts#38.

## How has this been tested?

Tested that the values returned are correct, most importantly that the `latest_delegation_tx_hash` in `/pool_delegators` returns the correct latest tx.


rdlrt (approved):

rdlrt:
Postpone this PR to koios-1.0.7 for next week perhaps - to be sure we dont add any oversight last min (and update specs)

Scitz0:
> Postpone this PR to koios-1.0.7 for next week perhaps - to be sure we dont add any oversight last min (and update specs)

👍

PR #1490, opened 2022-08-13
[EIC] Only update last_tx_id of previous epoch on epoch transition

dostrelith678:
## Description

The epoch transition part of the function was overwriting `i_last_tx_id`  values for all epochs instead of just the recently finished one.

~I removed `AND b.epoch_no <> _curr_epoch`, is there any use for it?~


Scitz0:
> I removed `AND b.epoch_no <> _curr_epoch`, is there any use for it?

I assume this is to not populate i_last_tx_id of current epoch as its not finalized yet, that we prefer to have a null value here instead of a temporary one.

Scitz0 (approved):
👍

PR #1501, opened 2022-08-16
Bump gRest version to koios-1.0.7

rdlrt:
## Description

- [x] - Bump setup-grest.sh version
- [x] - Update instructions to use dbsync v13.0.5
- [x] - Consider use of node 1.35.3 (dis-ambiguate for mainnet), as this is applicable purely for koios nodes, not SPO relays/core nodes.
- [x] - Update setup-grest.sh formatting (replace `echo -e` with `printf`)
- [x] - Add updates made to alpha branch, especially bulk inputs
- [x] - Validate stake-snapshot and epoch updates being used - if to be re-used across any relevant endpoints.
- [x] - Update entry in grest-poll.sh to stake stake_snapshot instead of epoch info, aim is to eliminate downtime at epoch transition.
- [x] - Create changelog


rdlrt:
To avoid skipping epochs that may not have any tx (especially with more networks around - but also for guildnet) , some considerations that'd be nice to fix before marking this release:
1. i_last_tx_id should not be empty, it should get populated with latest tx_id until respective epoch
2. We could possible add/modify states in control_table to decide whether or not a run should trigger, makes it independent of chain activity

Currently, on guildnet - the epoch_info_cache does not populate i_last_tx_id for multiple epochs, making preventing further updates

PR #1502, opened 2022-08-17
[EIC] Use last valid tx if no txs in epoch

dostrelith678:
## Description

Epoch info cache had a bug where no tx is populated if there were no txs in the epoch.


rdlrt (approved):
LGTM

PR #1504, opened 2022-08-19
[RPC] Add bulk inputs for selected endpoints

dostrelith678:
## Description

Draft for supporting bulk input for selected endpoints, starting with the initial list from the linked issue:
```
/block_transactions - minor changes to accommodate
/address_info - significant changes to accommodate
/address_assets - significant changes to accommodate
/account_info - significant changes
/account_rewards - minor changes
/account_updates - significant changes to response fields to allow client-side ordering
/account_addresses - minor changes
/account_assets - significant changes, similar to address_assets 
/account_history - minor changes
```
All mentioned endpoints will require schema changes to accommodate new responses.

`/account_updates` loses pre-ordering by switching to JSON so there are some extra fields to allow ordering. We could drop some of them if we want to keep the endpoint lighter.
`/account_rewards` also loses ORDER BY but epoch_nos are already in there.

The most important is to agree on the new response schemas proposed in this PR.

## Motivation and context

Allowing consumers to specify bulk inputs results in fewer API calls to complete specific tasks. Beneficial for both consumers and providers.

## Which issue it fixes?

Closes cardano-community/koios-artifacts#39


rdlrt (commented):
LGTM, but might need performance tests + schema docs (as mentioned)

Wrt performance testing , I'd vote for looking at it post merge - making it easier to use further reviews

dostrelith678:
`/account_rewards` - could think about supporting multiple epoch_no passed in instead of just 2 options (full history, or one specific epoch)

rdlrt:
> `/account_rewards` - could think about supporting multiple epoch_no passed in instead of just 2 options (full history, or one specific epoch).
> 
> Same for `/account_history`.

Is it different to having horizontal filter (from performance pov) when run against all epochs?

dostrelith678:
> > `/account_rewards` - could think about supporting multiple epoch_no passed in instead of just 2 options (full history, or one specific epoch).
> > Same for `/account_history`.
> 
> Is it different to having horizontal filter (from performance pov) when run against all epochs?

Because we are switching to one JSON per account input, I think we lose horizontal filtering...

rdlrt:
> > > `/account_rewards` - could think about supporting multiple epoch_no passed in instead of just 2 options (full history, or one specific epoch).
> > > Same for `/account_history`.
> > 
> > 
> > Is it different to having horizontal filter (from performance pov) when run against all epochs?
> 
> Because we are switching to one JSON per account input, I think we lose horizontal filtering...

ah right, makes sense then

rdlrt (approved):

PR #1518, opened 2022-08-30
add CNODE_NAME to backup filename

getwildr:
## Description
I modified the cntools.sh backup filename by adding the CNODE_NAME (the one that comes from prereqs.sh -t) which comes with env.

backup operations..

## Motivation and context
We may need to deploy many instances within /opt/cardano base dir.
I thought it could help to know which instance a backup comes from by simply reading its filename instead of having to decrypt/list the produced backup.

## Which issue it fixes?
n/a

## How has this been tested?
I did backups from cntools.sh, and produced the following files:
-rw-rw-r-- 1 ubuntu ubuntu  3851 Aug 30 12:03 offline_cntools_backup-20220830120303.mix02.tar.gz.gpg
-rw-rw-r-- 1 ubuntu ubuntu  3733 Aug 30 12:03 offline_cntools_backup-20220830120333.mix02.tar.gz
-rw-rw-r-- 1 ubuntu ubuntu  3512 Aug 30 12:02 online_cntools_backup-20220830120214.mix02.tar.gz.gpg
-rw-rw-r-- 1 ubuntu ubuntu  3395 Aug 30 12:02 online_cntools_backup-20220830120248.mix02.tar.gz
Result was the expected one.

### DIFF
--- a/scripts/cnode-helper-scripts/cntools.sh
+++ b/scripts/cnode-helper-scripts/cntools.sh
@@ -3576,9 +3576,9 @@ function main {
                    "--delete *${ASSET_POLICY_SK_FILENAME}"
                    "--delete *${ASSET_POLICY_SK_FILENAME}.gpg"
                  )
-                 backup_file="${backup_path}online_cntools_backup-$(date '+%Y%m%d%H%M%S').tar"
+                 backup_file="${backup_path}online_cntools_backup-$(date '+%Y%m%d%H%M%S').${CNODE_NAME}.tar"
                  ;;
-              1) backup_file="${backup_path}offline_cntools_backup-$(date '+%Y%m%d%H%M%S').tar" ;;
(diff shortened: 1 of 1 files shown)


Scitz0 (changes requested):
Change looks good to me, can include/merge it together with existing 10.0.4 PR.

Though not sure why we set backup_file inside of case statement. Its the same for both options anyway. Lets move it outside of case statement, and place if after esac line.
Option for yes can be replaced with null command to do nothing.
`1) : ;;`

Can you make these changes to the PR?

getwildr:
@Scitz0 you say _Its the same for both options anyway_ but it's not, its either offline or online according to the selected option.

Scitz0:
> @Scitz0 you say _Its the same for both options anyway_ but it's not, its either offline or online according to the selected option.

You are right 🤦‍♂️ my bad 😄 
I'll approve PR for later merge.

Scitz0 (approved):

PR #1519, opened 2022-08-30
Ask for user input to deploy cardano-submit-api as a service

getwildr:
## Description
Add question + read answer before deploying cardano-submit-api as a service.

run deploy-as-systemd.sh

## Motivation and context
cardano-submit-api binds local IP and listens to a new port.
I strongly believe that end user should decide whether he needs this API service to run or not on his system.

## Which issue it fixes?
n/a

## How has this been tested?
I've run the script and confirmed that systemd service was deployed if answered y and not deployed if answered n.

### DIFF
--- a/scripts/cnode-helper-scripts/deploy-as-systemd.sh
+++ b/scripts/cnode-helper-scripts/deploy-as-systemd.sh
@@ -26,9 +26,13 @@ if grep -q "^PGPASSFILE=" "${CNODE_HOME}/scripts/dbsync.sh" 2> /dev/null || [[ -
 fi
 
 echo -e "\e[32m~~ Cardano Submit API ~~\e[0m"
-echo "launches the main submitapi.sh script to deploy cardano-submit-api as service"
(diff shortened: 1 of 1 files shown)


scripts/cnode-helper-scripts/deploy-as-systemd.sh
```diff
@@ -26,9 +26,13 @@ if grep -q "^PGPASSFILE=" "${CNODE_HOME}/scripts/dbsync.sh" 2> /dev/null || [[ -
 fi
 
 echo -e "\e[32m~~ Cardano Submit API ~~\e[0m"
-echo "launches the main submitapi.sh script to deploy cardano-submit-api as service"
+echo "Please note that cardano-submit-api will listen on ports on your system to serve the API."
```
rdlrt, line 29: That message does not really serve a purpose IMO, deploy-as-systemd file should not be describing what a component is, that could be specified in documentation on node-cli.md page

rdlrt (approved):

PR #1523, opened 2022-09-02
Add CNCLI versions Info In GLV

getwildr:
## Description
Added a function check_cncli_versions that checks CNCLI github repository,
And then tells us whether we have:
 - No CNCLI installed
 - CNCLI Installed up to date
 - CNCLI installed not up to date
 - CNCLI Github not aval or no response.

Run GLV

## Motivation and context
I've seen users like me sayong recently they didn't know CNCLI could be upgraded.

## Which issue it fixes?
n/a

## How has this been tested?
I've run GLV without hard setting the VARS, then setting manually the VARS so that I could test all the use cases.


rdlrt (changes requested):
IMO - this should not be a part of UI itself, could potentially be a part of cncli.sh instead with status flag

Also, note that cncli is not compulsory to run gLiveView

rdlrt:
The UI is used for simplicity and getting important stat in a place, it shouldn't become from scope pov to show the status of different (esp. optional) components. Many miss versions involved or smaller details due to not following announcements and official documentations (home page of which specifies a baseline).

Every small space within UI is carefully added, I think the tool should empower solid SPOs, not lower the bar further for those not using right prereqs flags or not aware of what tools they're running

Will let @Scitz0 take final call

getwildr:
@rdlrt Everytime someone asks for help, or want to discuss an issue on TG/Discord, he pastes a screenshot of GLV.
Because its the perfect tool to have an overview of things.
Thats why I consider the CNCLI notifs deserve to be mentioned.
You use cardano-node: version/hash is displayed,
You use GLV: version is displayed.
You use CNCLI (at least when you do sync/leaderlogs/validate)), and you don't want to display the version : why ? :smile: 
I don't share your opinion.
Yes people miss some announcement, like I do sometimes, because they cannot spend their whole life on TG/Discord.
So I'd say lets rather help them not to miss important info/announcement.

rdlrt:
> You use cardano-node: version/hash is displayed,
You use GLV: version is displayed.

Because you cant have gLiveView running without those 2 components, not the case with any other (and any increase in verticle/horizontal size should be treated carefully)

Scitz0:
On vacation for next 10 days. I will look it over and comment once back home. Please leave open until then.

Scitz0:
Ok, so I have now looked over this PR a bit. I sort of see points made by both sides in the comments above. We need to keep gLiveView as light as possible while also being powerful enough to show data valuable to SPO. 

As we already have a cncli integration I don't think its a bad idea to highlight to the user that a new version of cncli is available if used on the system. With this, I mean both installed and used(service running for this deployment). Another change to PR I want to make is not to include it in the update loop, one call every 2s against github to check the version is a waste of resources. It's enough to just do a one-time check on startup and display before rendering the main UI. This will also reduce the need to find a good place to show the version. Only if not running the latest version will it be displayed on startup.

getwildr:
Hi,

I completely agree with you.

I just didn't understand if you were expecting me to do these changes or if
you were to implement it ?

Scitz0:
Well, I'm currently stretched a bit thin with other commitments so if you feel up to the task based on the above inputs, please go ahead and make changes and I can give input as we move closer to something we call all be happy about :).

getwildr:
@Scitz0 I've tried to adapt and meet your requirements.
Let me know it that's what you were asking for.

scripts/cnode-helper-scripts/env
```diff
@@ -1031,6 +1031,21 @@ test_koios() {
...
+  if [[ $(command -v cncli) ]] ; then
+    vcur=$(cncli -V | sed 's/cncli /v/g')
+    vrem=$(curl -s https://api.github.com/repos/cardano-community/cncli/releases/latest | jq -r .tag_name)
+    if [[ ! -z ${vrem} ]]; then
+      [[ ${vcur} == ${vrem} ]] || upCNCLI="${FG_MAGENTA}CNCLI current version ($vcur) different from repo ($vrem).${NC}"
+    else
+      upCNCLI="${FG_RED}CNCLI : GITHUB API Not Reachable or Not Responding.${NC}"
+    fi
+  else
+    upCNCLI="${FG_RED}CNCLI nos installed. Run prereqs.sh -c [...] to install if needed.${NC}"
```
Scitz0, line 1044: We should just silently ignore for both cases, ie if github is unreachable or if cncli is not installed as this is an optional addon.

scripts/cnode-helper-scripts/env
```diff
@@ -1031,6 +1031,21 @@ test_koios() {
   [[ ${ENABLE_KOIOS} = 'Y' && -n ${KOIOS_API} && $(curl -Isf -m 5 ${KOIOS_API}/tip | head -1) = *"200 OK"* ]] || unset KOIOS_API
 }
 
+check_cncli_versions() {
+  if [[ $(command -v cncli) ]] ; then
```
Scitz0, line 1035: I think I mentioned it but we should also check if a cncli sync service is deployed for this instance. This command could be used to check status. Exit code 0 if running. So
```bash
systemctl is-active --quiet ${CNODE_VNAME}-cncli-sync.service
```
So line could then be modified to:
```bash
if command -v cncli >/dev/null && systemctl is-active --quiet ${CNODE_VNAME}-cncli-sync.service; then
```

scripts/cnode-helper-scripts/gLiveView.sh
```diff
@@ -151,6 +151,13 @@ if [[ ${UPDATE_CHECK} = Y && ${SKIP_UPDATE} != Y ]]; then
     1) $0 "$@" "-u"; myExit 0 ;; # re-launch script with same args skipping update check
     2) exit 1 ;;
   esac
+
+  # check for CNCLI update
+  if [ -z "${CNCLI_STATUS}" ] ; then CNCLI_STATUS=$(check_cncli_versions) ; fi
```
Scitz0, line 156: We dont really need to test if variable is previously set or not, it should always be overwritten on startup no matter what. So this could be changed to just:
```bash
CNCLI_STATUS=$(check_cncli_versions)
```

scripts/cnode-helper-scripts/gLiveView.sh
```diff
@@ -151,6 +151,13 @@ if [[ ${UPDATE_CHECK} = Y && ${SKIP_UPDATE} != Y ]]; then
...
+  if [ -n "${CNCLI_STATUS}" ] ; then
+    printf "$CNCLI_STATUS\n\n" && read -p "Press key to ACK"
```
Scitz0, line 158: We already have a function in env file loaded at startup to use for user interaction called `waitToProceed`.
The default message is good enough and in line with scripts in guild scripts library. So just call this command as is with no params and it will wait for keypress with text:
`press any key to proceed ..`

Scitz0 (changes requested):
Hi,

Thanks for the changes. I added some comments and example replacement code you can look at and test. The reason I'm not just making the changes myself is that I want to encourage community participation in building and extending the tool set, with guidance.

scripts/cnode-helper-scripts/gLiveView.sh
getwildr: That was old code so as to avoid the calling to repo API every 2 secs.

scripts/cnode-helper-scripts/gLiveView.sh
getwildr: Great, about to update this ASAP.

scripts/cnode-helper-scripts/env
getwildr: OK sounds right indeed.

getwildr:
It's more than right for me to proceed this way. Going to make the changes right now.

getwildr:
Hi, are you expecting some more actions from me ?

Scitz0:
> Hi, are you expecting some more actions from me ?

Well yes, but I could just as well do them :)

See the notes above I made after your last commit.

getwildr:
Hi, well that's why I ask,
If you made some comments about my last commit (eg. 0796757)
I don't see them. I thought I had implemented everything you had asked for.
And didn't see any reply from you.
Please tell me if you just didn't see my changes or if I didn't see more
recent comments :-)

Scitz0:
Yes, and after that I made these comments. 
image

I cant say why these are not visible to you. 

I now pushed a change to your branch that moves check from env to gLiveView as this is the only script utilizing the check. I also shortened it according to my suggestion above. 

Please let me know if it works as expected. I didn't have time to test the change.

Scitz0:
Priyank let me know that I missed submitting review, still in pending state 🤦‍♂️ ... that one is on me :)

getwildr:
OK I understand why I hadn't see your comments.
Now I need your help because I don't know what I should do with git now ?
I've tested your latest changes and they're OK.
Do I need to implement in my local branch then push it to my origin branch, so that you see it in the PR,
Or you already made that sort of things according to https://github.com/cardano-community/guild-operators/pull/1523/commits/093e813d3027e79c135a851ed049cfc904f349b7 ?
I'm a bit lost here and don't want to screw the whole github thing.
Please tell How I should proceed now.

Scitz0:
Nope, its all good. I added changes according to commit linked. With your response that code seems good I just need to make sure its up to date with current alpha and merge it.

Scitz0 (approved):
👍

rdlrt (approved):

PR #1540, opened 2022-10-01
shows update proposal with filename

getwildr:
## Description
Just a little plus,
Now, checkUpdate() will display the name of the file that has an update available.
Even though 9 times out of 10 it is about ENV, now it will be clearly said so that there is no doubt.

For instance:
Change env content so that hash changes and check that script self update detection shows "env" in the update available notification.

## Motivation and context
When scripts detects updates, I find interesting to know which script it's about.

## Which issue it fixes?
n/a

## How has this been tested?
Change env content so that hash changes and check that script self update detection message was the following:
>> env script update(s) detected, do you want to download the latest version? (yes/no):


Scitz0 (approved):
LGTM

getwildr:
Just a quick question,
I was about to work on tuning a bit ENV, and existing scripts so that we can work with forked repositories.
e.g. use a G_REPO="cardano-community/guild-operators" that we could modify to make testing easier when working on a fork.
But I prefer to have your opinion first,
Would that be OK for you or do you rather think that is only a tester concern, and he only should do the needed SED/s to be able to work with its fork repo ?

rdlrt:
> I was about to work on tuning a bit ENV, and existing scripts so that we can work with forked repositories.

If you'd like to commit to collaborating long term, we'd be happy to add you as collaborator too

> use a G_REPO="cardano-community/guild-operators" that we could modify

I think when doing a fork, it will only be the org account that needs to change

> Would that be OK for you or do you rather

Yip, should be fine

getwildr:
> > I was about to work on tuning a bit ENV, and existing scripts so that we can work with forked repositories.
> 
> If you'd like to commit to collaborating long term, we'd be happy to add you as collaborator too

Oh, I'd be glad yes thanks.

> > use a G_REPO="cardano-community/guild-operators" that we could modify
> 
> I think when doing a fork, it will only be the org account that needs to change

Sure, just a way to make sure I'm looking at a git related thing. Also, I was wondering: when you fork a project you MUST keep the name . 

> > Would that be OK for you or do you rather
> 
> Yip, should be fine

Great !

PR #1548, opened 2022-10-05
[gRest] /datum_info endpoint

dostrelith678:
## Description

Adding an endpoint for getting datum information based on its hash.
Currently returns only the datum `value` in JSON and `bytes` format. We could add more to it in the future, for now, it seems to be all that's needed.

To consider: make this endpoint accept a list of hashes as bulk input?

## Which issue it fixes?

Closes https://github.com/cardano-community/koios-artifacts/issues/86

## How has this been tested?

Tested endpoint on guildnet.

### DIFF
--- a/files/grest/rpc/script/datum_info.sql
+++ b/files/grest/rpc/script/datum_info.sql
@@ -1,0 +1,22 @@
+CREATE FUNCTION grest.datum_info (_datum_hash text)
+  RETURNS TABLE (
+    hash text,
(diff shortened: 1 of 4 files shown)


rdlrt:
> make this endpoint accept a list of hashes as bulk input

Would agree , as it's a simple table, makes sense to add bulk query ability

rdlrt (approved):
LGTM, need to update specs

PR #1554, opened 2022-10-13
set NetworkName in dbsync config to content of ${NETWORK}.

pyromaniac3010:
this will make dbsync config work for preview and preprod, too.


rdlrt (approved):
LGTM, thanks

PR #1557, opened 2022-10-19
gRest: Prevent triggering no duplicate runs for stake distribution cache of new accounts

rdlrt:
## Description

Add a check for stake distribution cache update for new accounts to exit if the function itself  is already running

## Motivation and context

On guildnet, noticed the instance was down due to number of parallel queries already running

## Which issue it fixes?

Closes cardano-community/koios-artifacts#115

## How has this been tested?

Ran on guildnet


Scitz0 (approved):
👍

PR #1559, opened 2022-10-21
Asset object consistency in different endpoints

Scitz0:
## Description
It was raised that asset objects lacks consistency in different endpoints. 

- [x] address_assets
- [x] address_info
- [x] account_assets, change doesn't make it worse but this endpoint is really slow for some accounts
- [x] asset_history
- [x] asset_summary
- [x] asset_list
- [x] tx_utxos
- [x] tx_info

- [x] API spec

## Which issue it fixes?
https://github.com/cardano-community/koios-artifacts/issues/108

## How has this been tested?
Modified functions has been tested to return results without errors.


rdlrt:
As discussed, for `asset_list`:
- Return policy_id, ascii_name (which is currently hex) and fingerprint
- Downside is it will return duplicate list of policies, but can be easily managed using native (simple) filters, without having to use `->` , which can be difficult to work with

AlexDochioiu (approved):
the API changes look good to me. Not particularly skilled in databases so cannot asses the correctness of queries

AlexDochioiu:
@rdlrt  `asset_list, do we want to remove ascii here and rename hex to asset_name?`

I'd personally keep `asset_name` to always be hex for consistency with all the other endpoints. If you'd rather still keep the ascii format, maybe it should be called `asset_name_ascii`?

Scitz0:
> I'd personally keep asset_name to always be hex for consistency with all the other endpoints. If you'd rather still keep the ascii format, maybe it should be called asset_name_ascii?

I think this was the suggestion that @rdlrt suggested. To rename current hex field to asset_name that returns hex value. And to remove ascii field. I think we are all in agreement on this :)
I will make the final change to asset_list to move it from a view to a function and restructure it to be in line with the rest.

rdlrt:
For `asset_list`:
- Having played around with it a bit, I think it's not as easy to move view to RPC (as it also applied to other endpoints).
- Instead, we might want to keep the asset_list as view and simply align the objects as discussed - which will also help not to make this a breaking change on input for now.

As regards view => RPC, the are constraints are for RETURN TABLE type, we might want to leverage limits/offsets using a feature branch for PostgREST, but would keep that discussion in cardano-community/koios-artifacts#110

rdlrt (approved):

files/grest/rpc/transactions/tx_info.sql
```diff
@@ -371,7 +371,8 @@ BEGIN
             JSONB_BUILD_OBJECT(
               'policy_id', ENCODE(MA.policy, 'hex'),
               'asset_name', ENCODE(MA.name, 'hex'),
-              'quantity', MTM.quantity::text
```
dostrelith678, line 374: Should we retain `quantity` instead of `balance` for the amount minted?

Scitz0: yeah, makes more sense for this endpoint

dostrelith678 (approved):

files/grest/rpc/transactions/tx_info.sql
rdlrt: In that case, does it make sense to rename all balance to quantity?

Scitz0: Ok, yeah might be best to keep it consistent. I thing quantity works in all circumstances.

AlexDochioiu: +1 for renaming all `balance` to `quantity`

Issue #1560, opened 2022-10-24
new ogmios v5.5.6 zip file layout breaks prereqs.sh


pyromaniac3010:
**Describe the bug**
The new https://github.com/CardanoSolutions/ogmios/releases/download/v5.5.6/ogmios-v5.5.6-x86_64-linux.zip does no longer contain subdirectories, which breaks prereqs.sh ogmios setup resulting in `ERROR: ogmios downloaded but binary not found after extracting package!`

**To Reproduce**
1. Execute `prereqs.sh -o` 
2. See error

**Expected behavior**
Successfull download and installation of ogmios.

**Screenshots**

**Version:**
 - OS: Ubuntu 22.04
 - Product version: prereqs.sh master branch

**Additional context**
-

PR #1561, opened 2022-10-24
fix: use bin/ogmios or ogmios as for ogmios zip install

pyromaniac3010:
Fixes https://github.com/cardano-community/guild-operators/issues/1560


Scitz0 (approved):
👍

rdlrt (changes requested):

PR #1564, opened 2022-10-25
[gRest] Cache next epoch nonce when it becomes available

dostrelith678:
## Description

We want to cache the next epoch nonce as soon as it becomes available at the 7/10 time mark inside the epoch. This is relevant mainly for the leaderlogs calculations endpoint `/pool_stake_snapshot`.

Closes https://github.com/cardano-community/koios-artifacts/issues/114


files/grest/cron/jobs/populate-next-epoch-nonce.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+DB_NAME=cexplorer
+NWMAGIC=
+EPOCH_LENGTH=
+PROM_URL=
+
+export CARDANO_NODE_SOCKET_PATH=/opt/cardano/cnode/sockets/node0.socket
```
dostrelith678, line 7: Is this the correct way to do it if we need to use `cardano-cli`? Would need to use `sed` here as well similarly to the above constants.

files/grest/cron/jobs/populate-next-epoch-nonce.sh
```diff
@@ -0,0 +1,17 @@
...
+
+echo "$(date +%F_%H:%M:%S) Running next epoch nonce calculation..."
+min_slot=$((EPOCH_LENGTH * 7 / 10))
+current_epoch=$(curl -s "${PROM_URL}" | grep epoch | awk '{print $2}')
+current_slot_in_epoch=$(curl -s "${PROM_URL}" | grep slotInEpoch | awk '{print $2}')
+next_epoch=$((current_epoch + 1))
+
+[[ ${current_slot_in_epoch} -ge ${min_slot} ]] &&
+  next_epoch_nonce=$(echo "$(cardano-cli query protocol-state --testnet-magic "${NWMAGIC}" | jq -r .candidateNonce.contents)$(cardano-cli query protocol-state --testnet-magic "${NWMAGIC}" | jq -r .lastEpochBlockNonce.contents)" | xxd -r -p | b2sum -b -l 256 | awk '{print $1}') &&
+  psql ${DB_NAME} -c "INSERT INTO grest.epoch_info_cache (epoch_no, p_nonce) VALUES (${next_epoch}, ${next_epoch_nonce}) ON CONFLICT DO NOTHING;" 2>&1 1>/dev/null
```
dostrelith678, line 17: missing "job done" echo or similar

files/grest/cron/jobs/populate-next-epoch-nonce.sh
rdlrt: For consistency, would be good to have same format (User Variables vs Do Not modify sections), where part below are populated by setup-grest.sh , and -f would cause variables to be overwritten using checkUpdate function from env.

Alternatively, we could populate files when setup-grest.sh is run based on env params

dostrelith678: We also need to give it the `cardano-cli` path, so I suggest we use `setup-grest.sh` for this as it is only related to a cron job and is easily handled in that script.

```
[[ ${job} == populate-next-epoch-nonce ]] &&
      sed -e "s@NWMAGIC=.*@NWMAGIC=${NWMAGIC}@" -i "${CRON_SCRIPTS_DIR}/${job}.sh" &&
      sed -e "s@EPOCH_LENGTH=.*@EPOCH_LENGTH=${EPOCH_LENGTH}@" -i "${CRON_SCRIPTS_DIR}/${job}.sh" &&
      sed -e "s@PROM_URL=.*@PROM_URL=http://${PROM_HOST}:${PROM_PORT}/metrics@" -i "${CRON_SCRIPTS_DIR}/${job}.sh" &&
      sed for CARDANO_NODE_SOCKET_PATH &&
      sed for CARDANO_CLI_PATH  
```

Scitz0: I understand @rdlrt point of user variables for consistency but I dont think these scripts change much from default as variables are properly populated by setup-grest.sh

So I think it's fine to continue as @dostrelith678 suggested.

But we need to add CCLI variable populated with the correct absolute path.
Also, is CARDANO_NODE_SOCKET_PATH path replaced by setup-grest to correct path, if cnode is not used as path?

files/grest/cron/jobs/populate-next-epoch-nonce.sh
Scitz0: correct, should add `echo "$(date +%F_%H:%M:%S) Job done!"` like the rest of the cron jobs.

files/grest/cron/jobs/populate-next-epoch-nonce.sh
dostrelith678: Yes, absolute path for CCLI, I am thinking just `$(which cardano-cli)`, assuming user running `setup-grest.sh` knows about it.

`CARDANO_NODE_SOCKET_PATH` I guess depends only on the vname for cnode... Is there any better way to detect it? I guess we could scan the `cardano-node` process for the socket argument. In that case, this step would fail if `cardano-node` is actually not running for some reason.

Scitz0: setup-grest.sh source env file, so CCLI variable and SOCKET variable is already available in setup-grest to be used to replace variable in cron script. I would use this over finding bin from PATH.

scripts/grest-helper-scripts/setup-grest.sh
```diff
@@ -121,7 +121,7 @@ SGVERSION=1.0.8 # Using versions from 1.0.5-1.0.9 for minor commit alignment bef
       sed -e "s@NWMAGIC=.*@NWMAGIC=${NWMAGIC}@" -i "${CRON_SCRIPTS_DIR}/${job}.sh" &&
       sed -e "s@EPOCH_LENGTH=.*@EPOCH_LENGTH=${EPOCH_LENGTH}@" -i "${CRON_SCRIPTS_DIR}/${job}.sh" &&
       sed -e "s@PROM_URL=.*@PROM_URL=http://${PROM_HOST}:${PROM_PORT}/metrics@" -i "${CRON_SCRIPTS_DIR}/${job}.sh"
-      sed -e "s@CCLI_PATH=.*@CCLI_PATH=$(which cardano-cli)@" -i "${CRON_SCRIPTS_DIR}/${job}.sh"
+      sed -e "s@CCLI_PATH=.*@CCLI_PATH=${CCLI}}@" -i "${CRON_SCRIPTS_DIR}/${job}.sh"
       sed -e "s@CARDANO_NODE_SOCKET_PATH=.*@CARDANO_NODE_SOCKET_PATH=${CARDANO_NODE_SOCKET_PATH}@" -i "${CRON_SCRIPTS_DIR}/${job}.sh"
```
Scitz0, line 125: Should replace this with ${SOCKET} as well

Scitz0: actually, never mind, CARDANO_NODE_SOCKET_PATH is fine to use instead of SOCKET 👍

rdlrt (approved):
Not sure why we have CCLI_PATH instead of CCLI, but LGTM otherwise

Issue #1262, opened 2022-01-06
Simplify custom CNODE_VNAME in files and variables


rdlrt:
When adding new files to the repository, currently things work pretty well when using prereqs.sh. But if a new script is added to a folder outside of prereqs (eg: to manually add submitapi.sh without wanting to run prereqs), and if the contents include deploying systemd.sh - it gives a potential for users to accidentally deploy with invalid top level folder. Thus, any new script to be added outside of prereqs.sh needs to have checkUpdate run against it.

One of the approaches to simplify this is as below:
- Allow prereqs.sh to set custom environment variables for interactive sessions (like it currently does, using `${CNODE_VNAME}_HOME`).
- In the `env` file , export static CNODE_HOME variable against `$(dirname $0/../..)`
- The actual scripts should not have to bother about CNODE_VNAME at all, thus - allowing a direct comparison from downloaded github without having to replace environment variables (for STATIC section)
- Reduce VNAME resolution in `checkUpdate` function

Pushed this as an issue for now, rather than PR - as this needs to be done very carefully - with careful testing thinking of edge cases to not make it a breaking change (prereqs run to update once is fine, but user should not have to re-setup any variables and stuff - hopefully without adding too much of backlog to clear out later 🙂 )

rdlrt:
Current plan in linked PR:
- Remove sed for `CNODE_HOME` with alternate name based off `CNODE_NAME` from prereqs and env, for download as well as for update checks.
- Use `$(reallink ${PARENT}/..)` to determine the value for `CNODE_HOME` (`reallink` is part of coreutils package which has been present in prereqs/guild-deploy.
- Introduce variable `USESYSVARS` in env file which will toggle if `CNODE_HOME` should be inherited from shell, default set to `N.`

PR #1572, opened 2022-11-10
Re-Vamp all scripts

rdlrt:
### cabal-build-all.sh:
  - [x] Remove cardano-ping, as it's not used by tools - and cncli offers ping

### guild-deploy.sh:
  - [x] Archive rotatePoolKeys.sh and system-info.sh
  - [x] Change default method of cncli to download from github rather than compile locally (in line with other components)
  - [x] Use function-based definitions for prereqs.sh
  - [x] Re-Vamp options from prereqs.sh 
  - [x] Add top level path as arg
  - [x] modularise further using setup options similar to setup-grest.sh
  - [x] split OS level prereqs into build prereqs and basics)
  - [x] Add option to download pre-built binaries (can use hydra builds for now, and update to use in-house CI later)

### setup-grest.sh:
  - [x] Simplify haproxy config for instance providers (disable health-check by default, avoid creating seperate backend for SSL in comments, remove parts that were only required by monitoring instances)
  - [x] Add guildnet to use same token registry as other non-mainnet networks
  - [x] Remove -f option, as it has exactly same effect as using '-i c'

### grest-poll.sh:
  - [x] Use new environment variables introduced in haproxy 2.6.1 (mainly HAPROXY_SERVER_SSL) to avoid having to rely on custom logic for ssl mode detection
  - [x] Update argument parsing logic to accept 2 arguments or 4 (current) - alongwith possibility to append -d flag for debug
  - [x] Update log_err to print the same message to stdout as being added to log, and re-work those debug messages

### cncli.sh:
  - [x] Make Koios the the default method to get pool and stake snapshot.
  - [x] Remove duplicate test_koios call in cntools (already done when sourcing library)
  - [x] Remove trailing slash in KOIOS_API example

### CNTools:
  - [x] CNToolsAdd modifications for HW CNODE capability
  - [x] Fix for space in name
  - [x] Bump CNTools version and changelog

### Docs
  - [x] Replace references to prereqs.sh with guild-deploy.sh
  - [x] Create migration doc for existing users using prereqs.sh based setups
  - [x] Add rationale with reason for changes
  - [x] Add updated (migration) doc references in script usage itself, perhaps as a reminder after successful execution
  - [x] Create a dedicated small section for $PATH troubleshooting and removing duplicates

### Other Misc:
  - [x] Mark `prereqs.sh` as deprecated and introduce `guild-deploy.sh` to provide same (and more) functionality as per below
  - [x] Consider changes discussed in #1262 (will be breaking, hence - hesitant to integrate yet, but then it would be best time to do so if there are other breaking changes for existing deployment)
    - [x] Remove sed for CNODE_VNAME from prereqs and env
    - [x] Use `${PARENT}/..` to determine the value for `CNODE_HOME`
    - [x] Introduce variable `USESYSVARS` in `env` file which will toggle if CNODE_HOME should be inherited from shell, default set to N.
  - [x] Update github premerge workflows
  - [x] Replace ~/.cabal/bin references with ~/.local/bin
  - [x] Update docker scripts/workflows


Scitz0 (approved):
👍

redoracle (approved):
👍

hodlonaut (approved):
Looks good

PR #1578, opened 2022-11-24
Check for systemctl existence to be compliant with docker deployment

Scitz0:
## Description
Adds a check for the existence of systemctl to make it docker compatible as this binary might not exist there.


redoracle (approved):
lgtm, the one more condition should make it skip.

PR #1584, opened 2022-12-12
fixed getAssetInfo when spaces in keys

getwildr:
## Description
Changed the method to populate mint info array

In advanced/Asset List/Show

## Motivation and context
I was testing minting with guild tools, and faced an issue because of spaces in name key.

## Which issue it fixes?
n/a

## How has this been tested?
List/Show(ed) my policy/asset, and got the expected result.

### DIFF
--- a/scripts/cnode-helper-scripts/cntools.library
+++ b/scripts/cnode-helper-scripts/cntools.library
@@ -633,7 +633,7 @@ isPoolRegistered() {
 #               1: on error ($error_msg contains error message)
 #               2: offline/disabled/no result
 getAssetInfo() {
-  unset 
+  unset
   if [[ ${CNTOOLS_MODE} = "OFFLINE" || -z ${KOIOS_API} || "$#" -ne 2 ]]; then
     return 2
   else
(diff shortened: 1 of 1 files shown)


Scitz0 (commented):
Thanks for the suggested fix. We opted for tabulated output of jq in the first place because repeated jq calls are slow. I'm wondering if there is a way to solve this without reverting to calling jq for each entry we want 🤔

getwildr:
Well I've tried several ways, got stuck with spaces every time.

Scitz0:
> Well I've tried several ways, got stuck with spaces every time.

Could you provide an asset(policy and hex name) that has this problem?
I can just play around a bit as well, else we will go with your solution.

getwildr:
Sure, here you are !

```
Policy ID      : 98a7969b9e6481cd06c3f73b28fc7e8362785d3acd9ed0d171741f62
Policy Expire  : 2022-12-12 14:43:30 UTC, 00:55:51 remaining
Asset Name     : APPLE1 (4150504c4531)
Fingerprint    : asset12zkx3hg8m7f78z3ljkaz498fv3vnrulzf9ln8m
```

Scitz0:
> Sure, here you are ! (preview)
> 
> ```
> Policy ID      : 98a7969b9e6481cd06c3f73b28fc7e8362785d3acd9ed0d171741f62
> Policy Expire  : 2022-12-12 14:43:30 UTC, 00:55:51 remaining
> Asset Name     : APPLE1 (4150504c4531)
> Fingerprint    : asset12zkx3hg8m7f78z3ljkaz498fv3vnrulzf9ln8m
> ```

I'm a bit confused, are you referring to the asset_name_ascii field?
If so it doesn't contain any spaces for this asset?

getwildr:
Oh apologies, I thought you'd retrieve the data with this.
Here is the link to the cexplorer page
https://preview.cexplorer.io/asset/asset12zkx3hg8m7f78z3ljkaz498fv3vnrulzf9ln8m/metadata#data

Scitz0:
I did, but hex name converted to ascii does not contain any spaces, only the name within metadata as json object. Didn't understand that this is what you meant.

Now its more clear. So we need to handle .minting_tx_metadata and .token_registry_metadata in a better way.

getwildr:
That's exactly the thing !

Scitz0:
So what we should do instead is to revert changes you made, then replace @json in jq query with @base64.
Then on row 
```bash
a_minting_tx_metadata=${asset_info_arr[7]}
a_token_registry_metadata=${asset_info_arr[8]}
```
replace these with:
```bash
a_minting_tx_metadata=$(base64 -d <<< "${asset_info_arr[7]}")
a_token_registry_metadata=$(base64 -d <<< "${asset_info_arr[8]}")
```

Could you please make this change and test it?

getwildr:
Seems to do the trick ! Well done.

PR #1586, opened 2022-12-16
various improvements

gufmar:
## Description
- handle edge cases (log rotation, switched event sequence, ...)
- change fetch-request time from first to effective (block serving)
- reporting nodes port (if multiple nodes behind same pub IP)
- handle rollbacks
- tolerate early header announcements (pre slot time)
- performance improvements at grep/jq level


scripts/cnode-helper-scripts/blockPerf.sh
```diff
@@ -13,13 +13,15 @@
 #SELFISH_MODE='Y'      # in case you don't want to share this node's block propagation data, turn the selfish mode on (Y)
 #SERVICE_MODE='N'      # if you deploy-as-service this script it will run with the -s (service) parameter, and surpress console/syslog output. you can overwrite it here, restart the service and watch the console output with 'journalctl -f -u cnode-tu-blockperf'
 
+#CNODE_PORT=6010       # the port on which this node runs (automatically read from cnTools:env   outside cnTools you need to manually set this parameter
```
rdlrt, line 16: Best to keep default value in [sync](https://github.com/cardano-community/guild-operators/blob/setup-upds/scripts/cnode-helper-scripts/env#L14) with env

scripts/cnode-helper-scripts/blockPerf.sh
```diff
@@ -91,8 +93,9 @@ fi
...
-  [[ "$(jq -r .TraceChainSyncClient "${CONFIG}")" != "true" ]] && jq '.TraceChainSyncClient = true' ${CONFIG} > ${CONFIG}.tmp && mv ${CONFIG}.tmp ${CONFIG} && echo "INFO: Enabling node config TraceChainSyncClient" && config_changed=1
-  [[ "$(jq -r .TraceBlockFetchClient "${CONFIG}")" != "true" ]] && jq '.TraceBlockFetchClient = true' ${CONFIG} > ${CONFIG}.tmp && mv ${CONFIG}.tmp ${CONFIG} && echo "INFO: Enabling node config TraceBlockFetchClient" && config_changed=1
+  [[ "$(jq -r .TraceChainSyncClient "${CONFIG}")" != "true" ]] && jq '.TraceChainSyncClient = "true"' ${CONFIG} > ${CONFIG}.tmp && mv ${CONFIG}.tmp ${CONFIG} && echo "INFO: Enabling node config TraceChainSyncClient" && config_changed=1
+  [[ "$(jq -r .TraceBlockFetchClient "${CONFIG}")" != "true" ]] && jq '.TraceBlockFetchClient = "true"' ${CONFIG} > ${CONFIG}.tmp && mv ${CONFIG}.tmp ${CONFIG} && echo "INFO: Enabling node config TraceBlockFetchClient" && config_changed=1
+  [[ "$(jq -r .TracingVerbosity "${CONFIG}")" != "NormalVerbosity" ]] && jq '.TracingVerbosity = "NormalVerbosity"' ${CONFIG} > ${CONFIG}.tmp && mv ${CONFIG}.tmp ${CONFIG} && echo "INFO: Set node config NormalVerbosity" && config_changed=1
```
rdlrt, line 98: 6 jq calls can be replaced with 1, see example [here](https://github.com/cardano-community/guild-operators/blob/setup-upds/scripts/grest-helper-scripts/setup-grest.sh#L302-L314), which can then be used as array (eg: [here](https://github.com/cardano-community/guild-operators/blob/setup-upds/scripts/grest-helper-scripts/setup-grest.sh#L290-L295)

scripts/cnode-helper-scripts/blockPerf.sh
```diff
@@ -109,6 +112,7 @@ if [[ "${CONFIG##*.}" = "json" ]] && [[ -f ${CONFIG} ]]; then
   NWMAGIC=$(jq -r .networkMagic < ${GENESIS_JSON})
   [[ "$(jq -r .TraceChainSyncClient "${CONFIG}")" != "true" ]] && echo "ERROR: please set \"TraceChainSyncClient\":\"true\" in config file " && errors=1
   [[ "$(jq -r .TraceBlockFetchClient "${CONFIG}")" != "true" ]] && echo "ERROR: please set \"TraceBlockFetchClient\":\"true\" in config file " && errors=1
+  [[ "$(jq -r .TracingVerbosity "${CONFIG}")" != "NormalVerbosity" ]] && echo "ERROR: please set \"TracingVerbosity\":\"NormalVerbosity\" in config file " && errors=1
```
rdlrt, line 115: same as above

Scitz0 (commented):
Both option 1 and 2 has preview as network name.

scripts/cnode-helper-scripts/blockPerf.sh
gufmar: it comes from `env`  ...ideally. 
This is only for users running the script outside cntools.

rdlrt: yip, but default values (even if commented) should be same IMO

rdlrt (approved):
LGTM  for this PR.
Possible future considerations:
- use of checkUpdate like other scripts (would allow prompts for updates - if env gets updated without the script - for instance via gLV , it could potentially have issues later on)
- Guild network support - not sure if worthwhile , but could be useful for coordinated experiments

scripts/cnode-helper-scripts/blockPerf.sh
```diff
@@ -138,6 +154,142 @@ else
...
+    while IFS= read -r blockLogLine;do
+      # parse block and propagation metrics from different log kinds
+      lineKind=$(jq -r .data.kind <<< $blockLogLine)
+      case $lineKind in
+        ChainSyncClientEvent.TraceDownloadedHeader)
+          if $missingTbh; then
+            line_tsv=$(jq -r '[
+             .at //0,
+             .data.slot //0,
+             .data.block //0
```
karknu, line 179: Is this needed? It looks like you don't reference line_data_arr[2]?

Scitz0: Could be that its added because line_tsv is printed later in the script, and having the block in output might be desirable there?

gufmar: right, this is a leftover from an older version, where we parsed blockNum instead of just using EKGs most recent blockHeight

gufmar: removed `data.block`, and added `data.peer.remote.addr|port` to also parse the header announcing peer

scripts/cnode-helper-scripts/blockPerf.sh
gufmar: right, agreed. will become part of next commit.

PR #1598, opened 2023-01-29
CNCLI Prometheus metrics error fix

Scitz0:
## Description
Prometheus doesn't deal with strings, so instead convert metric to a number to indicate status (ok / nok). And if run in non SERVED mode, print error and exit.

## Which issue it fixes?
String error output 

## How has this been tested?
No, not yet


Straightpool:
Looks good to me!

Straightpool (approved):

rdlrt (approved):

PR #1607, opened 2023-02-17
Clear header addr and header port every round

karknu:
## Description

Clear blockTimeTbhAddr and blockTimeTbhPort at the end of each round. Without it the first peer that provided a header when to scrip started will be reported as the first forever.

This is a one-line patch.
## Motivation and context

See description.
## How has this been tested?

Apply patch and observer the output of blockPerf.sh


rdlrt (approved):

rdlrt:
was included in #1609

PR #1609, opened 2023-02-21
Mgu blockperf updates

gufmar:
## Description
bugfix: added two variable resets


rdlrt (approved):

rdlrt:
approved, but same as #1607 (w/version bump)?

gufmar:
> approved, but same as #1607 (w/version bump)?

Ah sorry I missed @karknu already provided a fix. In order to have this fix detected from cntools env on node side as an update we need to increase the version number.

PR #1615, opened 2023-03-10
getmetrics: output arcused as integer

reqlez:
## Description
Bug fix for previous PR https://github.com/cardano-community/guild-operators/pull/1614

Issue with scientific notation output. Changed to integer.


Scitz0 (approved):
👍

PR #1617, opened 2023-03-12
CNTools 10.2.0

Scitz0:
## Description
Fixes HW signing and Koios status check.

## How has this been tested?
Yes, to some extent. But good to test more.

**TODO**:
- [x] ~~Add multi address query for Koios balance check to reduce calls and improve speed.~~ 
Disable Koios balance check. The way CNTools is constructed it doesn't allow for multi address balance query. It would require a complete re-write. In most cases, local node address lookup is much faster for multiple lookups due to this. Especially for testnet networks with smaller ledger state. 
- [x] Update changelog


scripts/cnode-helper-scripts/cntools.library
```diff
@@ -959,7 +959,7 @@ getBalance() {
   asset_name_maxlen=5; asset_amount_maxlen=12
   tx_in=""
   
-  if [[ -z ${KOIOS_API} ]]; then
+  # if [[ -z ${KOIOS_API} ]]; then
```
rdlrt, line 962: if we're disabling it, might as well remove the section?
Can always look back on previous codebase if desired, IMO - would be good to keep current version clean

rdlrt (approved):
LGTM, except the comment

PR #1619, opened 2023-03-20
trezord in container

TrevorBenson:
Adds trezord into the container. Will do a documentation PR for usage.


Scitz0:
I usually dont handle docker PRs as I dont use it my self. 
@redoracle added as a reviewer.

redoracle:
unless specific cases we don't specify the package version, so we should use always the latest version (talking about trezor-bridge_2.0.27_amd64.deb). Anyway imo it suites better guild-deploy.sh rather than docker, as an option to download hw drivers I guess. @Scitz0 What do you think?

Scitz0:
> unless specific cases we don't specify the package version, so we should use always the latest version (talking about trezor-bridge_2.0.27_amd64.deb). Anyway imo it suites better guild-deploy.sh rather than docker, as an option to download hw drivers I guess. @Scitz0 What do you think?

Docker design is not really my area. So how to best deal with trezor bridge I'm not sure. But maybe guild-deploy makes sense as then it could possibly even be made to work for CLI install. We currently just let user know when running cntools to install trezor bridge and link to site that describes it. While checking this I noticed that our link is now out of date and needs to be updated.

TrevorBenson:
> unless specific cases we don't specify the package version, so we should use always the latest version (talking about trezor-bridge_2.0.27_amd64.deb). Anyway imo it suites better guild-deploy.sh rather than docker, as an option to download hw drivers I guess. @Scitz0 What do you think?

* VaccumLabs have produced versions greater than 2.0.27.
* The Trezor.io team [link to only 2.0.27 as the latest version](https://suite.trezor.io/web/bridge/) to install.

I wanted to use the bridge version recommended by Trezor.io, presuming it has the most testing by them. If they decide to update their site and suggest using the latest version I'm happy to follow suit. However, if they do not I would prefer we discuss updating, which version and why when deciding to ignore the manufacturers suggested version.

If the preference is to integrate Trezor Bridge into `guild-deploy.sh` I will make the necessary changes to add it as an optional flag. However, as the container does not have SystemD init and the post installation scripts expect SystemD, this leads to errors installing inside containers. 

Here is output of an standard `apt` installation inside a container showing the errors mentioned:
```
root@91c1642b11dc:/home/guild# apt-get install /tmp/trezor-bridge_2.0.27_amd64.deb 
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Note, selecting 'trezor-bridge' instead of '/tmp/trezor-bridge_2.0.27_amd64.deb'
The following NEW packages will be installed:
  trezor-bridge
0 upgraded, 1 newly installed, 0 to remove and 0 not upgraded.
Need to get 0 B/5,067 kB of archives.
After this operation, 10.5 MB of additional disk space will be used.
Get:1 /tmp/trezor-bridge_2.0.27_amd64.deb trezor-bridge amd64 2.0.27 [5,067 kB]
Selecting previously unselected package trezor-bridge.
(Reading database ... 22921 files and directories currently installed.)
Preparing to unpack .../trezor-bridge_2.0.27_amd64.deb ...
Unpacking trezor-bridge (2.0.27) ...
Setting up trezor-bridge (2.0.27) ...
Created symlink /etc/systemd/system/multi-user.target.wants/trezord.service → /usr/lib/systemd/system/trezord.service.
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down
dpkg: error processing package trezor-bridge (--configure):
 installed trezor-bridge package post-installation script subprocess returned error exit status 1
Errors were encountered while processing:
 trezor-bridge
E: Sub-process /usr/bin/dpkg returned an error code (1)
root@91c1642b11dc:/home/guild# echo $?
100
```

Past conversations I believe preferred to not add container building logic into the `prereqs.sh` (now `guild-deploy.sh`) scripts. Unless this has changed I suggest the container retain its own steps to install Trezor bridge preventing error messages during `docker build`'s.

TrevorBenson:
@redoracle 

> unless specific cases we don't specify the package version, so we should use always the latest version (talking about trezor-bridge_2.0.27_amd64.deb). Anyway imo it suites better guild-deploy.sh rather than docker, as an option to download hw drivers I guess. @Scitz0 What do you think?

* PR branch includes workflow `trezor-bridge-latest.yml`. Once an epoch fetches [trezor/data/master/bridge/latest.txt](https://raw.githubusercontent.com/trezor/data/master/bridge/latest.txt) from upstream to track latest suggested trezord version. 
* trezor bridge installation still part of Dockerfile to avoid errors mentioned before. 
* Uses `master` branch to pull `trezor-bridge-latest.txt` for consistency. 
  * Unfortunately means once PR merges it will temporarily break the **Docker Image** workflow until `files/docker/node/release-versions/trezor-bridge-latest.txt` exists in master branch. 
* Documentation added for HW wallet mgmt in container. 
  * elevated permissions
  * mounting /dev/usb 
  * starting trezord process inside container as root 
* If the current PR as well as the documentation is approved I will retest the documentation steps under Ubuntu/Debian since I do all work under podman.

redoracle (commented):
I am not going to add this to the docker workflow, because Trezor github seems not that maintained I am not willing to risk break the docker building process.

You can always make a script to be copied over in the build in this folder:
https://github.com/cardano-community/guild-operators/tree/alpha/files/docker/node/addons

So anyone that wishes to install it can always run the script afterwards within the container it self.

FYI to download the lastest version of the trezor debian pkg you can use something similar to this:
```
#!/bin/bash

# Get the list of versions
json=$(curl -s "https://api.github.com/repos/trezor/data/contents/bridge")

# Get the latest version
latest_version=$(curl -s "https://api.github.com/repos/trezor/data/contents/bridge" | jq -r '.[] | select(.type=="dir") | .name' | grep -v cert | sort -V | tail -n 1)

# Get the URL of the latest version's _amd64.deb package
package_url="https://github.com/trezor/data/blob/master/bridge/${latest_version}/trezor-bridge_${latest_version}_arm64.deb"

# Download the package
wget $package_url
```

PR #1620, opened 2023-03-23
Misc

rdlrt:
## Description

Miniscule fixes for few files:

- guild-deploy.sh : Formatting - replace occurances of `echo` with `echo -e` and specify `\n` where desired
- topology-mainnet.json: Fix port number for RDLRT
- blockperf.sh: Prevent debug file from being created on each run
- setup-grest.sh:
  - HAProxy systemd definition was missing `$MAINPID` since it wasn't escaped correctly.
  - Fix output getting messing up with spaces when calling pkill


Scitz0 (approved):
👍

PR #1625, opened 2023-04-06
Bump node version

rdlrt:
## Description

Bump node version, still leave defaults on non-P2P as 1 relay + 1 BP is still expected to be run without P2P enabled


hodlonaut (approved):
Looks good.

PR #1626, opened 2023-04-12
Update guild-deploy.sh for cncli 5.3.2 download compiled with musl

Straightpool:
Changed cncli download parser to linux-musl from linux-gnu. Andrew obviously changed the compiler from standard GNU glibc to MUSL with cncli 5.3.2

## Description
Changed cncli download parser to linux-musl from linux-gnu

Try to download cncli 5.2.3 with script before patch and after patch

## Motivation and context
Andrew obviously changed the compiler from standard GNU to MUSL with cncli 5.3.2 and with this the file name changed and thus the parser fails without fix.

## Which issue it fixes?
Download cncli 5.3.2+

## How has this been tested?
Locally with patch and downloading cncli 5.3.2

### DIFF
--- a/scripts/cnode-helper-scripts/guild-deploy.sh
+++ b/scripts/cnode-helper-scripts/guild-deploy.sh
@@ -349,7 +349,7 @@ download_cncli() {
   echo -e "\n  Downloading CNCLI..."
   rm -rf /tmp/cncli-bin && mkdir /tmp/cncli-bin
   pushd /tmp/cncli-bin >/dev/null || err_exit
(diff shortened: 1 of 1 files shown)


Scitz0 (approved):
LGTM

AndrewWestberg:
Thanks guys. The motivation for moving to musl is to have a common target system instead of gnu. There were lots of compatibility issues and segfaults with people running older ubuntu versions or debian for example. This allows more people to simply use the binary instead of having to compile from scratch on every platform.

PR #1645, opened 2023-05-17
Remove cddl-format switch from cntools.library-buildtx

getwildr:
## Description
Remove cddl-format switch from cntools.library-buildtx

anythinh that needs to build/witness/submit a TX

## Motivation and context
CLI 8.0.0 does not accept the switch anymore

## Which issue it fixes?
n/c

## How has this been tested?
modified a pool OK with 8.0.0


Scitz0 (approved):
👍

PR #1650, opened 2023-05-30
Update guild-deploy.sh - libsecp build

Straightpool:
Moved LD_LIBRARY_PATH check to the front of the libsecp build, so it is applied on first run of the script.

## Description
Just ran the script on the fresh host. Libsecp was build but put into /usr/lib not /usr/local/lib due to the too late placement of the LD_LIBRARY_PATH check

Libsecp build

## Motivation and context
Sets correct location for libsecp 

## Which issue it fixes?
No issue yet to my knowledge

## How has this been tested?
Master script ran as is, libsecp libs in wrong folder, moved manually. Looked at source code found error. No re-runs of the script as of yet.

### DIFF
--- a/scripts/cnode-helper-scripts/guild-deploy.sh
+++ b/scripts/cnode-helper-scripts/guild-deploy.sh
@@ -277,6 +277,10 @@ build_dependencies() {
   fi
   # Cannot verify the version and availability of libsecp256k1 package built previously, hence have to re-install each time
   echo -e "\n[Re]-Install libsecp256k1 ..."
+  if ! grep -q "/usr/local/lib:\$LD_LIBRARY_PATH" "${HOME}"/.bashrc; then
+    echo -e "\nexport LD_LIBRARY_PATH=/usr/local/lib:\$LD_LIBRARY_PATH" >> "${HOME}"/.bashrc
+    export LD_LIBRARY_PATH=/usr/local/lib:$LD_LIBRARY_PATH
+  fi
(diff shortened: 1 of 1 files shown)


Straightpool:
Unsuccessful checks seem to be unrelated to the patch, some other construction maybe in alpha?

rdlrt:
Pretty sure the proposed change will not really have any change in outcome, especially the desired outcome.
The checks for LD_LIBRARY_PATH (to have it appended into bashrc and/or export it to current shell) should not have an impact on where the libraries are copied to.

That reference is done via `--prefix` option (absence of this option will use `/usr/local` , which is being substituted with `/usr` in this script. Thus, we could remove that argument to let the default behaviour of configure sustained

Straightpool:
You are of course right, I have removed the --prefix option and added an echo out at the end of the installation. I still believe it can't hurt to move the LD_LIBRARY_PATH setting before the make install call.

rdlrt:
> I still believe it can't hurt to move the LD_LIBRARY_PATH setting before the make install call.

While it wouldnt in most cases, in general - the variables are set after setup to allow for any erroneous or incomplete deployment to not add entries to bashrc file, as those should only be touched upon successful execution of a statement

reqlez:
Over at BSD crew we stopped building that all together and now use the release version of secp since it now includes the needed components by cardano-node
 https://www.freshports.org/math/secp256k1/

not sure if linux distributions have that package or not.

Just another option ;-)

PR #1658, opened 2023-06-20
Update docs about P2P

rdlrt:
## Description

With node-8.1.1, users may want to run their node with P2P  enabled. Update docs to include brief notes alongwith sample files for the same


hodlonaut (approved):
Looks good

PR #1661, opened 2023-06-22
Update topology-mainnet.json

yoann54:
## Description
change key "addr" to "address"

## Motivation and context
that create an error on node launch

## Which issue it fixes?
that create an error on node launch

## How has this been tested?
tested on mainnet


rdlrt:
Thanks, was updated meanwhile

PR #1664, opened 2023-06-30
Postgres Appendix: Update tuning & Retire centOS/RHEL as outdated

rdlrt:
## Description

- RHEL/CentOS - not used by instance providers, introduces lagged config references (besides there is no 'CentOS' anymore)
- Include a more complete sample configuration for postgres


hodlonaut (approved):
Looks good to me.

PR #1667, opened 2023-07-25
Add consumed-by-utxo flag

rdlrt:
## Description

If DBSync version is >= 13.1.1.3, no harm in setting this flag by default, it only adds value and simplifies joins

## Motivation and context

Various performance improvements for used vs unused tx_out consumptions


Scitz0 (approved):
👍  only reservation is how bash expands DBSYNC_ARGS later if/when adding more flags to it. Could potentially be converted into an array instead. A problem for tomorrow though.

PR #1671, opened 2023-08-06
Container Shutdown with SIGINT

### LINKED ISSUE #1670: [docker] - Signal not reaching node process
**Describe the bug**
Container restarts still lead to long startup times as originally reported in #1648.

**To Reproduce**
1. Pull an older cardanocommunity/cardano-node from hub.docker.com, for instance (e.g. `docker pull docker.io/cardanocommunity/cardano-node:latest`)
2. Start the node container with the required configs, environment vars and volumes.
3. Stop the container `docker stop <CONTAINER_NAME_OR_ID>`.
4. Observe that the container stop takes a long time. 
5. Start the container `docker start <CONTAINER_NAME_OR_ID>`.
6. Tail the `node0.json` file and observe `TraceImmutableDBEvent.ValidatedChunk` events.

**Expected behavior**
Restarting container or removing and running a container from the same image it was already running should start quickly and not require block validation each time.

**Version:**
 - OS: debian:stable-slim
 - Product version: CNTools 10.4.0 
 - Cardano Node version: 8.1.1
 - Network you're connecting to: 2022-10-25T00:00:00Z

**Additional context**
I observe a message during a stop event which says it will send a SIGKILL after not shutting down in a timely manner ` WARN[0010] StopSignal SIGTERM failed to stop container guild-relay in 10 seconds, resorting to SIGKILL`, I presume this shows up under docker as well as podman. The current trap and exec implementation does not appear to work as intended.

**Root Cause Analysis**
* The container image uses the default stop signal of SIGTERM, not SIGINT.
* The container  entrypoint is the `entrypoint.sh` script:
  * Implements **trap** for signals
  * Uses bash **exec** command to start `cnode.sh`
* The `cnode.sh` script:
  * Does not use **trap**
  * Does not use **exec** 

The result is that `entrypoint.sh` is not running inside the container to catch SIGTERM and translate it to a SIGINT for cardano-node. Since it uses exec the PID1 inside the container is actually `cnode.sh`. This can be observed in the following `ps aux` output from the container:
```
guild@f5e6d3408b6c:~$ ps aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
guild          1  0.0  0.0   8128  3512 ?        Ss   Jul04   0:00 bash /opt/cardano/cnode/scripts/cnode.sh
guild        340  5.7  0.8 1073793712 553604 ?   Sl   Jul04 2699:35 /home/guild/.local/bin/cardano-node run --topology /opt/cardano/cnode/files/topology.json --config /opt/cardano/cnode/files/config.json -
guild     837890  0.0  0.0   7196  4056 pts/0    Ss   19:10   0:00 bash
guild     837895  0.0  0.0  11080  4548 pts/0    R+   19:10   0:00 ps aux
```

Attempting to kill this with either SIGINT (2) or SIGTERM (15) does not stop the cardano-node process:

```
guild@f5e6d3408b6c:~$ kill -2 1
guild@f5e6d3408b6c:~$ 
guild@f5e6d3408b6c:~$ ps aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
guild          1  0.0  0.0   8128  3512 ?        Ss   Jul04   0:00 bash /opt/cardano/cnode/scripts/cnode.sh
guild        340  5.7  0.8 1073793712 553600 ?   Sl   Jul04 2700:11 /home/guild/.local/bin/cardano-node run --topology /opt/cardano/cnode/files/topology.json --config /opt/cardano/cnode/files/config.json -
guild     837890  0.0  0.0   7196  4056 pts/0    Ss   19:10   0:00 bash
guild     838070  0.0  0.0  11080  4488 pts/0    R+   19:20   0:00 ps aux
guild@f5e6d3408b6c:~$ kill -15 1
guild@f5e6d3408b6c:~$ 
guild@f5e6d3408b6c:~$ ps aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
guild          1  0.0  0.0   8128  3512 ?        Ss   Jul04   0:00 bash /opt/cardano/cnode/scripts/cnode.sh
guild        340  5.7  0.8 1073793712 553600 ?   Sl   Jul04 2700:11 /home/guild/.local/bin/cardano-node run --topology /opt/cardano/cnode/files/topology.json --config /opt/cardano/cnode/files/config.json -
guild     837890  0.0  0.0   7196  4056 pts/0    Ss   19:10   0:00 bash
guild     838071  0.0  0.0  11080  4460 pts/0    R+   19:20   0:00 ps aux
```

This is what leads to the docker/podman stop command reaching its timeout (30 second default unless shortened) and finally sending SIGKILL's.

I've got a branch which resolves this issue. I believe it may also provide a better experience for SystemD users since the Main PID would more accurately represent the cardano-node process, not the `cnode.sh` bash script. I will submit a PR for review.

EDIT: #1671 opened for review.

TrevorBenson:
The change this branch makes is to:
1. Define `STOPSIGNAL` as **SIGINT** in the dockerfile_bin used to build the container.
2. Remove the trap from `entrypoint.sh` because it uses `exec` and is no longer running inside the container to catch any signals.
3. Add **exec** to `cnode.sh` for command `$CNODEBIN` so that cnode.sh is not running, but relinquishes control to cardano-node.

This results in the container process having cardano-node as PID 1 and the runtime engine will default to sending a SIGINT signal to PID1. 
```
guild@b2a142633c12:~$ ps aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
guild          1  4.4  3.0 1073794660 1907756 ?  Ssl  03:05  44:14 /home/guild/.local/bin/cardano-node run --topology /opt/cardano/cnode/files/topology.json --config /opt/cardano/cnode/files/config.json --
guild    1846631  0.0  0.0   7196  3872 pts/0    Ss   19:38   0:00 bash
guild    1846637  0.0  0.0  11084  4456 pts/0    R+   19:38   0:00 ps aux
```

This should also result in the systemd setup having the `Main PID` of the `cardano-node` process itself, instead of `cnode.sh` as the Main PID of the service. From inside the container the `cnode.sh -s` shutdown logic still worked as intended. 

I believe since the service unit has KillSignal defined as SIGINT, that if cnode.sh uses exec the service unit could have the **ExecStop** removed. When cardano-node is the Main PID it will receive the SIGINT instead of `cnode.sh`. I believe this was the reason for an explicit **ExecStop** after defining the signal.

Closes #1670

### DIFF
--- a/files/docker/node/addons/entrypoint.sh
+++ b/files/docker/node/addons/entrypoint.sh
@@ -1,9 +1,5 @@
 #!/usr/bin/env bash
 
-trap 'killall -s SIGINT cardano-node' SIGINT SIGTERM
-# "docker run --init" to enable the docker init proxy
-# To manually test: docker kill -s SIGTERM container
-
 head -n 8 ~/.scripts/banner.txt
 
 . ~/.bashrc > /dev/null 2>&1
--- a/files/docker/node/dockerfile_bin
+++ b/files/docker/node/dockerfile_bin
@@ -1,5 +1,8 @@
 FROM debian:stable-slim
 
+# DEFINE THE PREFERRED STOP SIGNAL
+STOPSIGNAL SIGINT
+
 LABEL desc="Cardano Node by Guild's Operators"
 ARG DEBIAN_FRONTEND=noninteractive
 ARG G_ACCOUNT
--- a/scripts/cnode-helper-scripts/cnode.sh
+++ b/scripts/cnode-helper-scripts/cnode.sh
@@ -133,7 +133,7 @@ pre_startup_sanity
 
 # Run Node
 if [[ -f "${POOL_DIR}/${POOL_OPCERT_FILENAME}" && -f "${POOL_DIR}/${POOL_VRF_SK_FILENAME}" && -f "${POOL_DIR}/${POOL_HOTKEY_SK_FILENAME}" ]]; then
-  "${CNODEBIN}" "${CPU_RUNTIME[@]}" run \
+  exec "${CNODEBIN}" "${CPU_RUNTIME[@]}" run \
     --topology "${TOPOLOGY}" \
     --config "${CONFIG}" \
     --database-path "${DB_DIR}" \
@@ -144,7 +144,7 @@ if [[ -f "${POOL_DIR}/${POOL_OPCERT_FILENAME}" && -f "${POOL_DIR}/${POOL_VRF_SK_
     --port ${CNODE_PORT} \
     ${MEMPOOL_OVERRIDE} "${host_addr[@]}"
 else
-  "${CNODEBIN}" "${CPU_RUNTIME[@]}" run \
+  exec "${CNODEBIN}" "${CPU_RUNTIME[@]}" run \
     --topology "${TOPOLOGY}" \
     --config "${CONFIG}" \
     --database-path "${DB_DIR}" \


rdlrt (approved):

rdlrt:
> I believe since the service unit has KillSignal defined as SIGINT, that if cnode.sh uses exec the service unit could have the ExecStop removed.

I somehow am still not a fan of relying on systemd (as a dependency as well as future+flavours), I would still like to rely on `cnode.sh -s` which is in operator's hands, but agree with the changes in PR 👍🏻

Scitz0 (approved):

PR #1672, opened 2023-08-08
Fix startupLogMonitor systemd and remove duplicate service definition

rdlrt:
## Description

In it's current state, startupLogMonitor was preventing node startup if not running - due to dependency created on cnode itself (there does not need to be a dependency with node - as the script itself is primarily only reading log files). Additionally, the service definition was duplicated in deploy-as-systemd.sh as well as the original script, modify the script to 'call' the existing script instead.


Scitz0 (approved):
👍

Issue #1632, opened 2023-04-24
Feature request: Separate backup files location


intertreeJK:
Suggestion: 

Script updates so that the previous versions of the files being updated can be copied into an explicit 'archive' subdirectory off of 'scripts'.

As the file lists can get a bit messy all in the same place as the current versions over time. 

Would be nice to have a single directory in which they can be stored/nuked from.

Scitz0:
Not a bad idea. A simple wildcard delete command will remove them all though if you want to clear all backups.
`rm *_bkp*`

intertreeJK:
Indeed. I am aware of this, but it still leaves the tidying up to the user,
instead of being nicely managed from the start to avoid root scripts
directory clutter.

rdlrt:
@intertreeJK - Please check linked PR

PR #1673, opened 2023-08-09
Move script backups to archive subfolder

rdlrt:
## Description

- When scripts go through updates, previous backups are saved with _bkp$(date +%s) suffix in same folder. Instead, we start saving **future** backups to a subfolder called archive. This is only true for scripts (not for configs, which will continue to be saved in same folder). To keep it non-intrusive from user actions point of view, added a check for missing archive folder to create it within update function, instead of having to re-run guild-deploy.sh
- Also, addressed slight inconsistency with setup-grest.sh and cabal-build-all.sh which was saving using `.bkp_$(date +%s)` to keep it consistent with format used at other places.

## Which issue it fixes?

#1632

### DIFF
--- a/scripts/cnode-helper-scripts/cabal-build-all.sh
+++ b/scripts/cnode-helper-scripts/cabal-build-all.sh
@@ -30,7 +30,7 @@ else
   CUSTOM_CABAL="N"
 fi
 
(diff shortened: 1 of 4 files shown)


Scitz0 (approved):
Looks good 👍

PR #1690, opened 2023-10-06
Koios v1.1.0rc

rdlrt:
## Description

- [x] Bump base URL HAProxy to v1 from v0, and accordingly update references in `env` file
- [x] Swap grestrpcs file to list exceptions and treat everything else as RPC (since list of RPC is many times greater than views/exceptions)
- [x] Update grest-poll.sh
  - [x] Use stricter structure check from spec (before it was only checking schema formation, not the data type for the formed schema)
  - [x] Change health check for asset_registry to include `asset_name` instead of `name`, and increase active stake cache population limit to 1/8th (instead of 1/10th of epoch)
- [x] Update Changelogs as per cardano-community/koios-artifacts#239
- [x] Update ogmios download link to point to specific github build (6.0.0-rc3)


Scitz0 (approved):
:+1:

rdlrt:
Not to be merged until gRest v1 is up on all networks

PR #1693, opened 2023-10-21
Support for Mithril Signer deployment.

TrevorBenson:
Adds support to download, install, and set up a mithril-signer for the appropriate network. Includes changes to `deploy-as-systemd.sh` to provide an option to run it as a Systemd service. 

### `mithril-client.sh` 
* Opt `-d` downloads the latest mithril snapshot
  * If the environment file does not exist it will create it.
* Opt `-u` creates or updates the environment file.
  * If the POOL_NAME is undefined or **CHANGE_ME** will create a minimal environment suitable for mithril client
  * If the POOL_NAME is defined will create a complete environment suitable for mithril signer
    * Presumes a naive deployment without mithril relay for automated environment setup on first start
    * Use either `deploy-as-systemd.sh`, or `mithril-signer.sh -u` directly, to define a complex mithril environment with a mithril relay (or the sidecar)

### `mithril-signer.sh` 
* follows the `ogmios.sh` format such that it can be used to start the mithril signer.
* Opt `-d` sets up the Systemd service unit.
  * If the environment file does not exist it will create it.
* Opt `-u` creates or updates the environment file.
  * uses read to determine if naive deployment or use a mithril relay (or the sidecar) IP and port to set RELAY_ENDPOINT

### `mithril-relay.sh`
* Opt `-d`
  * installs squid
  * uses read to gather variables for block producer IP and relay port.
  * Creates a squid.conf for a reverse proxy.
    * applies an ACL in the config for each block producer.
  * Starts the squid systemd service.
* Opt `-l`
  * installs nginx
  * uses read to gather variables for relay IP address(es), relay port, and IP to bind sidecar  (Relay port shared between all relays and sidecar)
  * Creates an nginx.conf load balancing over relays.
  * starts nginx systems service.
## Additional Details
* Increases storage requirements for binaries by ~42MB.
* Currently uses a static `mithril_release` set to **2337.0**.
  * Includes a commented-out option to obtain the latest release via tag_name, similar to cardano-signer.
  * Open to the preferred method suggested.

The mithril relay and squid configuration comes from Mithril docs. The HA setup via an nginx sidecar is a Proof of Concept at the moment. I have tested with a single relay, but not yet with the sidecar and multiple relays.

### DIFF
--- a/scripts/cnode-helper-scripts/env
+++ b/scripts/cnode-helper-scripts/env
@@ -80,7 +80,7 @@
 #ASSET_POLICY_ID_FILENAME="policy.id"
 #CIP0094_POLL_URL="https://raw.githubusercontent.com/cardano-foundation/CIP-0094-polls/main/networks/polls.json" # URL for polls to vote against
 
-#MITHRIL_DOWNLOAD="Y"                                  # (Y|N) Download latest Mithril snapshot
+#MITHRIL_DOWNLOAD="N"                                  # (Y|N) Download latest Mithril snapshot
 
 ######################################
 # Do NOT modify code below           #
@@ -800,7 +800,7 @@ set_default_vars() {
   [[ -z ${ASSET_POLICY_SCRIPT_FILENAME} ]] && ASSET_POLICY_SCRIPT_FILENAME="policy.script"
   [[ -z ${ASSET_POLICY_ID_FILENAME} ]] && ASSET_POLICY_ID_FILENAME="policy.id"
   [[ -z ${CIP0094_POLL_URL} ]] && CIP0094_POLL_URL="https://raw.githubusercontent.com/cardano-foundation/CIP-0094-polls/main/networks/polls.json"
-  [[ -z ${MITHRIL_DOWNLOAD} ]] && MITHRIL_DOWNLOAD="Y"
+  [[ -z ${MITHRIL_DOWNLOAD} ]] && MITHRIL_DOWNLOAD="N"
   FG_BLACK='\e[30m'
   FG_RED='\e[31m'
   FG_GREEN='\e[32m'
--- a/scripts/cnode-helper-scripts/env
+++ b/scripts/cnode-helper-scripts/env
@@ -80,6 +80,8 @@
 #ASSET_POLICY_ID_FILENAME="policy.id"
 #CIP0094_POLL_URL="https://raw.githubusercontent.com/cardano-foundation/CIP-0094-polls/main/networks/polls.json" # URL for polls to vote against
 
+#MITHRIL_DOWNLOAD="Y"                                  # (Y|N) Download latest Mithril snapshot
+
 ######################################
 # Do NOT modify code below           #
 ######################################
@@ -795,6 +797,7 @@ set_default_vars() {
   [[ -z ${ASSET_POLICY_SCRIPT_FILENAME} ]] && ASSET_POLICY_SCRIPT_FILENAME="policy.script"
   [[ -z ${ASSET_POLICY_ID_FILENAME} ]] && ASSET_POLICY_ID_FILENAME="policy.id"
   [[ -z ${CIP0094_POLL_URL} ]] && CIP0094_POLL_URL="https://raw.githubusercontent.com/cardano-foundation/CIP-0094-polls/main/networks/polls.json"
+  [[ -z ${MITHRIL_DOWNLOAD} ]] && MITHRIL_DOWNLOAD="Y"
   FG_BLACK='\e[30m'
   FG_RED='\e[31m'
   FG_GREEN='\e[32m'
--- a/scripts/cnode-helper-scripts/guild-deploy.sh
+++ b/scripts/cnode-helper-scripts/guild-deploy.sh
@@ -74,7 +74,7 @@ usage() {
 		  p   Install common pre-requisite OS-level Dependencies for most tools on this repo (Default: skip)
 		  b   Install OS level dependencies for tools required while building cardano-node/cardano-db-sync components (Default: skip)
 		  l   Build and Install libsodium fork from IO repositories (Default: skip)
-		  m   Build and install mithril from IO repositories (Default: skip)
+		  m   Download latest (released) binaries for mithril-signer, mithril-client (Default: skip)
 		  f   Force overwrite entire content of scripts and config files (backups of existing ones will be created) (Default: skip)
 		  d   Download latest (released) binaries for bech32, cardano-address, cardano-node, cardano-cli, cardano-db-sync and cardano-submit-api (Default: skip)
 		  c   Install/Upgrade CNCLI binary (Default: skip)
--- a/scripts/cnode-helper-scripts/guild-deploy.sh
+++ b/scripts/cnode-helper-scripts/guild-deploy.sh
@@ -99,7 +99,6 @@ set_defaults() {
   [[ -z ${INSTALL_OGMIOS} ]] && INSTALL_OGMIOS='N'
   [[ -z ${INSTALL_CSIGNER} ]] && INSTALL_CSIGNER='N'
   [[ -z ${CNODE_PATH} ]] && CNODE_PATH="/opt/cardano"
-  [[ -z ${MITHRIL_PATH} ]] && MITHRIL_PATH="/opt/mithril"
   [[ -z ${CNODE_NAME} ]] && CNODE_NAME='cnode'
   [[ -z ${CURL_TIMEOUT} ]] && CURL_TIMEOUT=60
   [[ -z ${UPDATE_CHECK} ]] && UPDATE_CHECK='Y'
--- a/scripts/cnode-helper-scripts/guild-deploy.sh
+++ b/scripts/cnode-helper-scripts/guild-deploy.sh
@@ -3,7 +3,6 @@
 # shellcheck source=/dev/null
 
 unset CNODE_HOME
-unset MITHRIL_HOME
 
 ##########################################
 # User Variables - Change as desired     #
@@ -111,7 +110,6 @@ set_defaults() {
   [[ "${SUDO}" = 'Y' && $(id -u) -eq 0 ]] && err_exit "Please run as non-root user."
   CNODE_HOME="${CNODE_PATH}/${CNODE_NAME}"
   CNODE_VNAME=$(echo "$CNODE_NAME" | awk '{print toupper($0)}')
-  MITHRIL_HOME="${MITHRIL_PATH}/mithril-signer"
   REPO="https://github.com/${G_ACCOUNT}/guild-operators"
   REPO_RAW="https://raw.githubusercontent.com/${G_ACCOUNT}/guild-operators"
   URL_RAW="${REPO_RAW}/${BRANCH}"
@@ -528,16 +526,9 @@ setup_folder() {
   $sudo chown -R "$U_ID":"$G_ID" "${CNODE_HOME}" 2>/dev/null
   
   if [[ ${INSTALL_MITHRIL} == 'Y' ]]; then
-
-    if grep -q "export MITHRIL_HOME=" "${HOME}"/.bashrc; then
-      echo -e "\nEnvironment Variable MITHRIL_HOME already set up!"
-    else
-      echo -e "\nSetting up MITHRIL_HOME Environment Variable"
-      echo -e "\nexport MITHRIL_HOME=${MITHRIL_HOME}" >> "${HOME}"/.bashrc
-    fi
   
-    $sudo mkdir -p "${MITHRIL_HOME}"/data-stores
-    $sudo chown -R "$U_ID":"$G_ID" "${MITHRIL_HOME}" 2>/dev/null
+    $sudo mkdir -p "${CNODE_HOME}"/mithril-signer/data-stores
+    $sudo chown -R "$U_ID":"$G_ID" "${CNODE_HOME}"/mithril-signer 2>/dev/null
   fi
 }
 
--- a/scripts/cnode-helper-scripts/guild-deploy.sh
+++ b/scripts/cnode-helper-scripts/guild-deploy.sh
@@ -3,6 +3,7 @@
 # shellcheck source=/dev/null
 
 unset CNODE_HOME
+unset MITHRIL_HOME
 
 ##########################################
 # User Variables - Change as desired     #
@@ -61,7 +62,7 @@ versionCheck() { printf '%s\n%s' "${1//v/}" "${2//v/}" | sort -C -V; } #$1=avail
 usage() {
   cat <<-EOF >&2
 		
-		Usage: $(basename "$0") [-n <mainnet|preprod|guild|preview>] [-p path] [-t <name>] [-b <branch>] [-u] [-s [p][b][l][f][d][c][o][w][x]]
+		Usage: $(basename "$0") [-n <mainnet|preprod|guild|preview>] [-p path] [-t <name>] [-b <branch>] [-u] [-s [p][b][l][m][f][d][c][o][w][x]]
 		Set up dependencies for building/using common tools across cardano ecosystem.
 		The script will always update dynamic content from existing scripts retaining existing user variables
 		
@@ -74,6 +75,7 @@ usage() {
 		  p   Install common pre-requisite OS-level Dependencies for most tools on this repo (Default: skip)
 		  b   Install OS level dependencies for tools required while building cardano-node/cardano-db-sync components (Default: skip)
 		  l   Build and Install libsodium fork from IO repositories (Default: skip)
+		  m   Build and install mithril from IO repositories (Default: skip)
 		  f   Force overwrite entire content of scripts and config files (backups of existing ones will be created) (Default: skip)
 		  d   Download latest (released) binaries for bech32, cardano-address, cardano-node, cardano-cli, cardano-db-sync and cardano-submit-api (Default: skip)
 		  c   Install/Upgrade CNCLI binary (Default: skip)
@@ -92,11 +94,13 @@ set_defaults() {
   [[ -z ${WANT_BUILD_DEPS} ]] && WANT_BUILD_DEPS='N'
   [[ -z ${FORCE_OVERWRITE} ]] && FORCE_OVERWRITE='N'
   [[ -z ${LIBSODIUM_FORK} ]] && LIBSODIUM_FORK='N'
+  [[ -z ${INSTALL_MITHRIL} ]] && INSTALL_MITHRIL='N'
   [[ -z ${INSTALL_CNCLI} ]] && INSTALL_CNCLI='N'
   [[ -z ${INSTALL_CWHCLI} ]] && INSTALL_CWHCLI='N'
   [[ -z ${INSTALL_OGMIOS} ]] && INSTALL_OGMIOS='N'
   [[ -z ${INSTALL_CSIGNER} ]] && INSTALL_CSIGNER='N'
   [[ -z ${CNODE_PATH} ]] && CNODE_PATH="/opt/cardano"
+  [[ -z ${MITHRIL_PATH} ]] && MITHRIL_PATH="/opt/mithril"
   [[ -z ${CNODE_NAME} ]] && CNODE_NAME='cnode'
   [[ -z ${CURL_TIMEOUT} ]] && CURL_TIMEOUT=60
   [[ -z ${UPDATE_CHECK} ]] && UPDATE_CHECK='Y'
@@ -107,6 +111,7 @@ set_defaults() {
   [[ "${SUDO}" = 'Y' && $(id -u) -eq 0 ]] && err_exit "Please run as non-root user."
   CNODE_HOME="${CNODE_PATH}/${CNODE_NAME}"
   CNODE_VNAME=$(echo "$CNODE_NAME" | awk '{print toupper($0)}')
+  MITHRIL_HOME="${MITHRIL_PATH}/mithril-signer"
   REPO="https://github.com/${G_ACCOUNT}/guild-operators"
   REPO_RAW="https://raw.githubusercontent.com/${G_ACCOUNT}/guild-operators"
   URL_RAW="${REPO_RAW}/${BRANCH}"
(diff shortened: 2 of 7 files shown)


scripts/cnode-helper-scripts/guild-deploy.sh
```diff
@@ -3,6 +3,7 @@
 # shellcheck source=/dev/null
 
 unset CNODE_HOME
+unset MITHRIL_HOME
```
rdlrt, line 6: I am not sure if mithril should need a seperate top level folder, and even if there was a strong reason to do so - it would have required contention across scripts to be able to set the path via flags, but really - I'd rather this path to be set as a subset of existing top level folder

scripts/cnode-helper-scripts/mithril-signer.sh
```diff
@@ -0,0 +1,160 @@
...
+  if [[ ! -f "${MITHRIL_HOME}"/service.env ]]; then
+    generate_environment_file && echo "Environment file created successfully!!"
+  fi
+
+  echo "Deploying ${CNODE_VNAME}-mithril-signer as systemd service.."
+  sudo bash -c "cat <<-'EOF' > /etc/systemd/system/${CNODE_VNAME}-mithril-signer.service
+    [Unit]
+    Description=Cardano Mithril signer service
+    StartLimitIntervalSec=0
+	Wants=network-online.target
```
rdlrt, line 96: ident

scripts/cnode-helper-scripts/mithril-signer.sh
```diff
@@ -0,0 +1,160 @@
...
+  sudo bash -c "cat <<-'EOF' > ${MITHRIL_HOME}/service.env
+  AGGREGATOR_ENDPOINT=https://aggregator.${RELEASE}-${NETWORK}.api.mithril.network/aggregator
+  KES_SECRET_KEY_PATH=${POOL_DIR}/${POOL_HOTKEY_SK_FILENAME}
+  OPERATIONAL_CERTIFICATE_PATH=${POOL_DIR}/${POOL_OPCERT_FILENAME}
+  PARTY_ID=$(cat ${POOL_DIR}/${POOL_ID_FILENAME})
+  DB_DIRECTORY=${CNODE_HOME}/db
+  CARDANO_CLI_PATH=${HOME}/.local/bin/cardano-cli
+  DATA_STORES_DIRECTORY=${MITHRIL_HOME}/data-stores
+  ERA_READER_ADAPTER_TYPE=cardano-chain
+  ERA_READER_ADDRESS=https://raw.githubusercontent.com/input-output-hk/mithril/main/mithril-infra/configuration/${RELEASE}-${NETWORK}/era.addr
```
rdlrt, line 78: will need exception if we're to stand something for guildnet

TrevorBenson: When I inquired last about guildnet I don't recall hearing anyone was already testing, and didnt get a feeling anyone was interested in doing so. Or at least everyone seemed too busy with other projects at that time.

* If one does exist today point me at it and I'll work out the changes required. If not then I suggest we leave this as is until we have an additional need and layouts to support.
* If anyone is interested in adding mithril to guildnet now that we have tools I would be happy to participate. 
  * If we choose an incompatible structure then **ERA_READER_ADDRESS**, **ERA_READER_VKEY**, and **GENESIS_VERIFICATION_KEY** can be migrated to the `set_defaults()` where choices are made based on the NETWORK.

scripts/cnode-helper-scripts/guild-deploy.sh
TrevorBenson: Their documentation starts with `/opt/mithril` as their source for storage. In theory Mithril is not limited to Cardano as it will support other blockchains and potentially cardano sidechains. 

If `guild-deploy.sh` will not be used in the future to deploy a sidechain of cardano then having storage in `/opt/cardano/cnode/mithril-signer` instead makes perfect sense, as well as nativaly allows a mithril signer per VNODE instance without extending VNODE into `/opt/mithril/mithril-signer` etc.

As long as this is being considered while choosing, I am open to whatever path the guild prefers to use.

TrevorBenson:
I opened #1694 to address the pre-merge check failures observed in this PR.

TrevorBenson:
With commit 14e43d8 the environment file contains each variable required to start. I was hoping to just source `/opt/cardano/cnode/scripts/env` however **EnvironmentFile** only takes variable assignments, so sourcing inside the `${CNODE_HOME}/mithril-signer/service.env` file does not work. I also tried `ExecStartPre=. /opt/cardano/cnode/scripts/env`, but that caused an error parsing the service unit. 

In the end it seemed simplest that since `mithril-signer.sh -u` is a forced update of the entire environment file, to simply embed NETWORK and CARDANO_NODE_SOCKET_PATH into the environment file as a workaround. Since these would not normally change after deployment it seems acceptable to have the operator run an update of the environment if they change paths.


```
$ sudo systemctl status cnode-mithril-signer.service 
● cnode-mithril-signer.service - Cardano Mithril signer service
     Loaded: loaded (/etc/systemd/system/cnode-mithril-signer.service; enabled; vendor preset: enabled)
     Active: active (running) since Mon 2023-10-23 00:25:47 UTC; 3s ago
   Main PID: 38606 (mithril-signer)
      Tasks: 14 (limit: 4555)
     Memory: 3.3M
        CPU: 143ms
     CGroup: /system.slice/cnode-mithril-signer.service
             └─38606 /home/tbenson/.local/bin/mithril-signer -vv

Oct 23 00:25:47 mithril cnode-mithril-signer[38606]: {"msg":"Storing/Getting immutables digests cache from: /opt/cardano/cnode/mithril-signer/data-stores/immutables_di>
Oct 23 00:25:48 mithril cnode-mithril-signer[38606]: {"msg":"New AggregatorHTTPClient created","v":0,"name":"slog-rs","level":20,"time":"2023-10-23T00:25:48.071677604Z>
Oct 23 00:25:48 mithril cnode-mithril-signer[38606]: {"msg":"Started","v":0,"name":"slog-rs","level":20,"time":"2023-10-23T00:25:48.071709418Z","hostname":"mithril","p>
Oct 23 00:25:48 mithril cnode-mithril-signer[38606]: {"msg":"STATE MACHINE: launching","v":0,"name":"slog-rs","level":30,"time":"2023-10-23T00:25:48.071733362Z","hostn>
Oct 23 00:25:48 mithril cnode-mithril-signer[38606]: {"msg":"================================================================================","v":0,"name":"slog-rs",">
Oct 23 00:25:48 mithril cnode-mithril-signer[38606]: {"msg":"STATE MACHINE: new cycle: Init","v":0,"name":"slog-rs","level":30,"time":"2023-10-23T00:25:48.071764699Z",>
Oct 23 00:25:48 mithril cnode-mithril-signer[38606]: {"msg":"RUNNER: get_current_epoch","v":0,"name":"slog-rs","level":20,"time":"2023-10-23T00:25:48.071779925Z","host>
Oct 23 00:25:48 mithril cnode-mithril-signer[38606]: {"msg":"RUNNER: update_era_checker","v":0,"name":"slog-rs","level":20,"time":"2023-10-23T00:25:48.104549893Z","hos>
Oct 23 00:25:48 mithril cnode-mithril-signer[38606]: {"msg":"Current Era is thales (Epoch 101).","v":0,"name":"slog-rs","level":20,"time":"2023-10-23T00:25:48.52686054>
Oct 23 00:25:48 mithril cnode-mithril-signer[38606]: {"msg":"… Cycle finished, Sleeping for 60000 ms","v":0,"name":"slog-rs","level":30,"time":"2023-10-23T00:25:48.526>
```

scripts/cnode-helper-scripts/mithril-signer.sh
```diff
@@ -97,6 +97,8 @@ deploy_systemd() {
 	StartLimitIntervalSec=0
 	Wants=network-online.target
 	After=network-online.target
+	BindsTo=${vname}.service
```
rdlrt, line 100: From what I understand it does not require binding to cnode , and does not have a strong dependency on local client (if cannot connect on startup, it would fail independently). Do you foresee a special case where this will be needed? Just thinking ahead, if not necessary - it's best to avoid adding a dependency.

PS: I know a small subset of other systemd services does have `BindsTo` and `After` , but fwir they have cases which lead to unhandled loop, it shouldnt be the case for this service [similar to submit-api and ogmios].

TrevorBenson: Mithril signer relies on the Cardano node socket if it starts before the node the service will result in the following failure.
```
Error: "Could not get Epoch: general error Error launching command Command { std: CARDANO_NODE_SOCKET_PATH=\"/opt/cardano/cnode/sockets/node0.socket\" \"/app/bin/cardano-cli\" \"query\" \"tip\" \"--testnet-magic\" \"1\", kill_on_drop: false }, error = '\ncardano-cli: Network.Socket.connect: <socket: 11>: does not exist (No such file or directory)'"
```

My using containers, and not the Systemd cnode.service unit, I do not recall if the service shows Active only after the socket is up or beore. Mithril signer having `After` is to eliminate, or at least reduce, the errors from Mithril signer in the system logs. 

I went with `BindsTo` as I did not observe the use of `Requires`, so figured it was a style or standardization choice. Also in every instance where a Systemd service unit in `deploy-as-systemd.sh` had an `After=cnode.service` statement it also included a `BindsTo=cnode.service` statement. Given this would ensure startup after cnode when the service unit is deployed, and if the node was stopped would also stop the Mithril signer so it wasn't spamming the logs every `RestartSec` until the node was back online, this felt like a good combination for a production Mithril signer + cnode Sytemd setup.

Scitz0: I think bindsTo in this case makes sense as explained so from my end 👍

scripts/cnode-helper-scripts/guild-deploy.sh
```diff
@@ -92,11 +93,13 @@ set_defaults() {
   [[ -z ${WANT_BUILD_DEPS} ]] && WANT_BUILD_DEPS='N'
   [[ -z ${FORCE_OVERWRITE} ]] && FORCE_OVERWRITE='N'
   [[ -z ${LIBSODIUM_FORK} ]] && LIBSODIUM_FORK='N'
+  [[ -z ${INSTALL_MITHRIL} ]] && INSTALL_MITHRIL='N'
   [[ -z ${INSTALL_CNCLI} ]] && INSTALL_CNCLI='N'
   [[ -z ${INSTALL_CWHCLI} ]] && INSTALL_CWHCLI='N'
   [[ -z ${INSTALL_OGMIOS} ]] && INSTALL_OGMIOS='N'
   [[ -z ${INSTALL_CSIGNER} ]] && INSTALL_CSIGNER='N'
   [[ -z ${CNODE_PATH} ]] && CNODE_PATH="/opt/cardano"
+  [[ -z ${MITHRIL_PATH} ]] && MITHRIL_PATH="/opt/mithril"
```
Scitz0, line 102: This can be removed now?

TrevorBenson:
```suggestion
```

TrevorBenson:
I believe the final check failures are due to opening the PR from my forked branch (origin), instead of my push being to cardanocommunity (upstream), leading to the premerge expecting the branch/commit to come from upstream but not existing.

TrevorBenson:
> I believe the final check failures are due to opening the PR from my forked branch (origin), instead of my push being to cardanocommunity (upstream), leading to the premerge expecting the branch/commit to come from upstream but not existing.

I pushed the branch from my fork upstream and ran a [manual premerge workflow](https://github.com/cardano-community/guild-operators/actions/runs/6615798687) which was completed successfully.

TrevorBenson:
@rdlrt @Scitz0 
I created a new branch named feature/mithril-build-with-relays. Would you like me to merge that commit into this PR for review or make that a new PR after this branch reaches approval and merges?

rdlrt:
> Would you like this added to this PR for review or make a new PR after this branch reaches approval and merges?

Up to you, happy to use this PR itself so they're merged together

TrevorBenson:
> > Would you like this added to this PR for review or make a new PR after this branch reaches approval and merges?
> 
> Up to you, happy to use this PR itself so they're merged together

Commit cherry-picked into this branch, I think the trunk (flake8) formatting adjusted the layout so it may not match indenting etc. for existing scripts. I'll fix that prior to merge and other general linter suggestions prior to merge.

Scitz0:
The one thing that is important to add to the scripts themselves is docs addition. Describe how the scripts should be used and how it all ties together. Doesn't have to explain Mithril itself except for one or two high-level sentences. Can then link to https://mithril.network/doc/ for more information.

As I have mentioned before I have yet to take the time to read up on Mithril so my knowledge isn't extensive yet. `cardano-signer.sh` is to be deployed on block producing node, and cardano-relay.sh on the relay node.  Then we have the client to actually download the blockchain snapshot. Are there plans to add something for this as well?

TrevorBenson:
> The one thing that is important to add to the scripts themselves is docs addition. Describe how the scripts should be used and how it all ties together. Doesn't have to explain Mithril itself except for one or two high-level sentences. Can then link to https://mithril.network/doc/ for more information.

Agreed. Once the logic and script functionality is approved I'll include a commit for Documentation before merge.

> As I have mentioned before I have yet to take the time to read up on Mithril so my knowledge isn't extensive yet. `cardano-signer.sh` is to be deployed on block producing node, and cardano-relay.sh on the relay node.  

mithril-signer and mithril-relay in this instance. 

For a naive deployment (testnets) the Mithril signer can be directly on a Cardano relay node, or dedicated system but let's presume relay for discussion sake, using `mithril-signer.sh -d` and no Mithril relay is required. 

For a production deployment the Mithril signer should be on the block producer node  using `mithril-signer.sh -d`. As well as a Mithril relay on a Cardano relay node by using `mithril-relay.sh -d`. If the operator wants relay redundancy they repeat the `mithril-relay.sh -d` set up on more than one relay. Then to allow the Mithril signer to load balance all relays, on the block producer node use `mithril-relay.sh -l`, provide all the Mithril relay IP's, an nginx loadbalancer is setup with all relays in its backend server pool.

>Then we have the client to actually download the blockchain snapshot. Are there plans to add something for this as well? 


Potentially, if that's something the guild would like. Automating the download of snapshots and extracting them could be done. I could see this as a `mithril-client.sh` which determines when to obtain snapshots and extract, or to skip. 

It could be directly at the cnode.sh level, using a flag like ENABLE_MITHRIL_CLIENT. When true cnode.sh calls mithril-client.sh which either gets the needed snapshots and exits or decides no snapshots are needed and exits, returning to cnode.sh

Otherwise to not touch cnode.sh it could be a Systemd only option which `mithril-client.sh -d` could apply a unit override to for cnode.service using a cnode.service.d directory. Adding an **ExecStartPre=bash -c '/opt/cardano/cnode/scripts/mithril-client.sh -XX'** (flags undecided ATM), causing it to check before the service main **ExecStart**.

TrevorBenson:
> I'm a bit split on the best deployment for this. But I'm leaning towards the second option to add an ExecStartPre override. This would be a cleaner option keeping it totally separate from normal deployment.

*Mithril client is responsible for restoring the Cardano blockchain on an empty node from a certified snapshot.* this was the logic I intended to have the script follow. 

@Scitz0 Unless you think the script should attempt to determine how far out of sync a node is, when it already has content in the DB folder, and whether to rely on block synchronization or get a new Mithril snapshot?

To be honest I've not spent much time thinking about how to determine this when `query tip` would not be available. I used the Mithril client when I had existing DB data. Except I knew how far out of sync my node was or that my filesystem had filled and corrupted the current DB anyway, so made an informed decision to take the action manually.

Scitz0:
> > I'm a bit split on the best deployment for this. But I'm leaning towards the second option to add an ExecStartPre override. This would be a cleaner option keeping it totally separate from normal deployment.
> 
> _Mithril client is responsible for restoring the Cardano blockchain on an empty node from a certified snapshot._ this was the logic I intended to have the script follow.
> 
> @Scitz0 Unless you think the script should attempt to determine how far out of sync a node is, when it already has content in the DB folder, and whether to rely on block synchronization or get a new Mithril snapshot?
> 
> To be honest I've not spent much time thinking about how to determine this when `query tip` would not be available. I used the Mithril client when I had existing DB data. Except I knew how far out of sync my node was or that my filesystem had filled and corrupted the current DB anyway, so made an informed decision to take the action manually.

My idea was only as an initial fresh deployment, not to keep track of sync state. I think we can keep it easy like this.

TrevorBenson:
> My idea was only as an initial fresh deployment, not to keep track of sync state. I think we can keep it easy like this. I'm also in theory fine with just adding it to cnode.sh if priyank feels this is ok as well.

@rdlrt What say ye? :smile:

Scitz0:
> > My idea was only as an initial fresh deployment, not to keep track of sync state. I think we can keep it easy like this. I'm also in theory fine with just adding it to cnode.sh if priyank feels this is ok as well.
> 
> @rdlrt What do you think?
> 
> * Using a Systemd override file to **ExecStartPre** `mithril-client.sh`
>   
>   * Automatic for Systemd use case.
>   * Manual for (cnode.sh non Systemd) and container use cases
> * Using `cnode.sh` to call `mithril-client.sh`
>   
>   * Automatic for all use cases.
> 
> The latter definitely gives the most coverage with identical behavior. The only question, if we go with the latter, is should the check for an empty `${CNODE_HOME}/db` be in `cnode.sh` before calling `mithril-client.sh` or be in `mithril-client.sh` to determine if the db directory is empty. Either works but I figured there may be a preference for separation of responsibilities.

I think everything related to mithril, ie if db is empty or not should be a job for `mithril-client.sh`. We should also add a boolean in user config setting to enable/disable mithril. I'm thinking that mithril could be default on. Though need to make sure it properly handles networks that might not have any snapshots like guild network currently. So it silently just starts syncing normally in this case. Or perhaps prints a line about it and continues. 

Other than this I think the cnode.sh route is preferred as you mention, else manual execution of cnode.sh wouldn't work the same way as systemd deployed version of cnode.sh.

TrevorBenson:
> I think everything related to mithril, ie if db is empty or not should be a job for `mithril-client.sh`. We should also add a boolean in user config setting to enable/disable mithril. I'm thinking that mithril could be default on. Though need to make sure it properly handles networks that might not have any snapshots like guild network currently. So it silently just starts syncing normally in this case. Or perhaps prints a line about it and continues.
> 
> Other than this I think the cnode.sh route is preferred as you mention, else manual execution of cnode.sh wouldn't work the same way as systemd deployed version of cnode.sh.

Great, thats what I did in my local branch. I'll do some more tests this evening and get it pushed.

TrevorBenson:
@Scitz0 @rdlrt I've pushed the commit with `mithril-client.sh`. This changes a few things around in naming conventions.

* After discussion with Arnaud about mithril-aggregator being a temporary artifact, using `/opt/cardano/cnode/mithril` as the directory for both signer and client.
* Both signer and client require a set of variables. Renamed `${CNODE_HOME}/mithril/service.env` to `${CNODE_HOME}/mithril/mithril.env`.
* Converted from using NETWORK to NETWORK_NAME, as the "source of truth" for network
  * NETWORK is common for container entrypoint.sh setting configs, but not for general cnode.sh or cnode.service Systemd usage.
  * Leads to a minor logic loop for containers
    1. Container sets **NETWORK** to determine the configs to download through `entrypoint.sh`
    2. ${CNODE_HOME}/scipts/env uses the configuration files to determine the **NWMAGIC** and sets the **NETWORK_NAME**
    3. When generating a mithril environment file `mithril-signer.sh` or `mithril-client.sh` use **NETWORK_NAME** to set the **NETWORK** variable in the `${CNODE_HOME}/mithril/mithril.env`.

While it is somewhat cyclical it should result in essentially an identical **NETWORK** variable as already used by the container `entrypoint.sh` being added to the mithril environment file and uses an identical logic to determine the variable regardless of use case for cnode (standalone binary execution, Systemd service, containerized).

scripts/cnode-helper-scripts/env
```diff
@@ -80,6 +80,8 @@
 #ASSET_POLICY_ID_FILENAME="policy.id"
 #CIP0094_POLL_URL="https://raw.githubusercontent.com/cardano-foundation/CIP-0094-polls/main/networks/polls.json" # URL for polls to vote against
 
+#MITHRIL_DOWNLOAD="Y"                                  # (Y|N) Download latest Mithril snapshot
```
TrevorBenson, line 83:
```suggestion
#MITHRIL_DOWNLOAD="N"                                  # (Y|N) Download latest Mithril snapshot
```

scripts/cnode-helper-scripts/env
```diff
@@ -798,6 +800,7 @@ set_default_vars() {
   [[ -z ${ASSET_POLICY_SCRIPT_FILENAME} ]] && ASSET_POLICY_SCRIPT_FILENAME="policy.script"
   [[ -z ${ASSET_POLICY_ID_FILENAME} ]] && ASSET_POLICY_ID_FILENAME="policy.id"
   [[ -z ${CIP0094_POLL_URL} ]] && CIP0094_POLL_URL="https://raw.githubusercontent.com/cardano-foundation/CIP-0094-polls/main/networks/polls.json"
+  [[ -z ${MITHRIL_DOWNLOAD} ]] && MITHRIL_DOWNLOAD="Y"
```
TrevorBenson, line 803:
```suggestion
  [[ -z ${MITHRIL_DOWNLOAD} ]] && MITHRIL_DOWNLOAD="N"
```

scripts/cnode-helper-scripts/mithril-client.sh
```diff
@@ -0,0 +1,193 @@
...
+		
+		EOF
+}
+
+
+generate_environment_file() {
+  export ERA_READER_ADDRESS=https://raw.githubusercontent.com/input-output-hk/mithril/main/mithril-infra/configuration/${RELEASE}-${NETWORK_NAME,,}/era.addr
+  export ERA_READER_VKEY=https://raw.githubusercontent.com/input-output-hk/mithril/main/mithril-infra/configuration/${RELEASE}-${NETWORK_NAME,,}/era.vkey
+  if [[ -n "${POOL_NAME}" ]] && [[ "${POOL_NAME}" != "CHANGE_ME" ]]; then
+    bash -c "cat <<-'EOF' > ${CNODE_HOME}/mithril/mithril.env
```
TrevorBenson, line 40: Assignment closer to usage.


```suggestion
generate_environment_file() {
  if [[ -n "${POOL_NAME}" ]] && [[ "${POOL_NAME}" != "CHANGE_ME" ]]; then
    export ERA_READER_ADDRESS=https://raw.githubusercontent.com/input-output-hk/mithril/main/mithril-infra/configuration/${RELEASE}-${NETWORK_NAME,,}/era.addr
    export ERA_READER_VKEY=https://raw.githubusercontent.com/input-output-hk/mithril/main/mithril-infra/configuration/${RELEASE}-${NETWORK_NAME,,}/era.vkey
    bash -c "cat <<-'EOF' > ${CNODE_HOME}/mithril/mithril.env
```

TrevorBenson:
The successfull pre-merge check (manually run) summary is available in this job https://github.com/cardano-community/guild-operators/actions/runs/6708797371

rdlrt (approved):

PR #1702, opened 2023-11-15
Mithril mkdir in scripts for existing deployments

TrevorBenson:
Both mithril-signer and mithril-client will check for directory existence before generating the mithril environment file. If the directory does not exist it will `mkdir -p` and `chown -R`.

### DIFF
--- a/scripts/cnode-helper-scripts/mithril-client.sh
+++ b/scripts/cnode-helper-scripts/mithril-client.sh
@@ -176,6 +176,10 @@ while getopts :duh opt; do
       usage
       exit 1
       ;;
+    *)
+      usage
+      exit 1
+      ;;
   esac
 done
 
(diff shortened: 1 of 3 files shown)


TrevorBenson:
Small adjustment to ensure client creates directories even when POOL_NAME is still CHANGEME (for relays or syncing before pool has been created). 

## Testing
I tested under a RHEL 9 instance. First pass w/ `guild-deploy.sh` NOT using `-s m` creates the directories. I then `rm -rf`'d the **${CNODE_HOME}/mithril** path, and used `${CNODE_HOME/scripts/mithril-client.sh -u` and it pre-created the directories and set permissions prior to generating the env file. The generate function inside `mithril-signer.sh` also has the check and mkdir at the very beginning, so should handle it identically.

PR #1726, opened 2024-01-25
Config updates

rdlrt:
## Description

- Update config templates to mirror naming convention as on [cardano-playground repo](https://github.com/input-output-hk/cardano-playground/tree/main/docs/environments)
- Clone book.play.dev.cardano.org version (with minor differences to avoid having to maintain a seperate submitapi config)
- Update corresponding doco
- Rename default logging (node0.json => node.json) and socket (node0.socket => node.socket) to be consistent with IO's naming.
- [Unintended for this PR - but might as well] setup-grest.sh - No longer need to trigger epoch-summary-corrections cron job for Koios, the functions still left as-is in case the bug (or a version of it) re-appears in future

## Motivation and context

- [book.world.cardano.org](https://book.world.dev.cardano.org) updated config links from their github, and old peers retired - and unless part of calls, this could have been missed. In order to not be dependent on upstream changes - and more importantly - add some version matching of current supported repo configs with the scripts , while maintaining history - the configs are being cloned to this repo, while adding addendum for logging to be JSON based + setting known tracers (eg: ones used by blockPerf) on guild-ops repo

## How has this been tested?

Applied forced update on guild node and restart node/dbsync after re-applying topology/env/cnode.sh user variable customisations.


Scitz0 (approved):

hodlonaut (approved):
looks okay to me.

PR #1730, opened 2024-01-29
Update dockerfile_bin to use ${GUILD_DEPLOY_BRANCH} instead of master

TrevorBenson:
## Description
Changes `files/docker/node/dockerfile_bin` to always rely on the GUILD_DEPLOY_BRANCH for the `guild-deploy.sh` file as well as the configuration files. 

## Motivation and context
Changes merged altering `config-mainnet.json` to `files/config/mainnet/config.json` leads to failures to build the image. Can be observed in [this job run](https://github.com/cardano-community/guild-operators/actions/runs/7690838700). 
Changing `dockerfile_bin` to use the **GUILD_DEPLOY_BRANCH** does work. It also ensures a container that is intended to test a specific set of changes obtains the **addons**, **scripts**, or **entrypoint.sh** for the container from the same branch intended for testing.

## Which issue it fixes?
Didn't open an issue, just created the branch and tested the fix. Can open one if we want the PR to track closing an issue.

## How has this been tested?
The working job from this branch can be observed in [this job run](https://github.com/cardano-community/guild-operators/actions/runs/7690854231).

### DIFF
--- a/files/docker/node/dockerfile_bin
+++ b/files/docker/node/dockerfile_bin
@@ -42,7 +42,7 @@ RUN adduser --disabled-password --gecos '' guild \
 RUN set -x && apt update \
     && apt-get update \
     && mkdir -p /root/.local/bin \
-    && wget https://raw.githubusercontent.com/${G_ACCOUNT}/guild-operators/master/scripts/cnode-helper-scripts/guild-deploy.sh \
+    && wget https://raw.githubusercontent.com/${G_ACCOUNT}/guild-operators/${GUILD_DEPLOY_BRANCH}/scripts/cnode-helper-scripts/guild-deploy.sh \
     && export SUDO='N' \
     && export UPDATE_CHECK='N' \
     && export SKIP_DBSYNC_DOWNLOAD='Y' \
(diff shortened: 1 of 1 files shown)


rdlrt:
While updating this particular instance might be OK - these should not get published/tagged as latest docker images , as anything not on master is not tested in field yet

rdlrt (approved):
While updating this particular instance might be OK - these should not get published/tagged as latest docker images , as anything not on master is not tested in field yet - which should anyways be achieved using github workflow's branch tag (being `master`). Hence, approving

TrevorBenson:
> While updating this particular instance might be OK - these should not get published/tagged as latest docker images , as anything not on master is not tested in field yet - which should anyways be achieved using github workflow's branch tag (being `master`). Hence, approving

Agreed. During an earlier change I put a "safety valve" in, so if the workflow runs, even if someone uses the workflow dispatch and does NOT set it to testing, [if the branch isn't master it won't push a new image up to hub.docker.com](https://github.com/cardano-community/guild-operators/blob/alpha/.github/workflows/docker_bin.yml#L57-L64).

PR #1735, opened 2024-02-09
Fix retiring pool modification ability

hodlonaut:
Separated 3 and 4 return codes for isPoolRegistered, since we want users to be able to modify scheduled-for-retirement pools but not retired ones.


Scitz0 (approved):
Looking good but please bump patch version at the head of library file before merge.

PR #1736, opened 2024-02-29
Add log level setting for ogmios

Scitz0:
## Description
Adds the ability to launch Ogmios with a set log level.

### DIFF
--- a/scripts/cnode-helper-scripts/ogmios.sh
+++ b/scripts/cnode-helper-scripts/ogmios.sh
@@ -44,7 +44,6 @@ set_defaults() {
     case ${LOG_LEVEL} in
       Debug)   : ;;
       Info)    : ;;
-      Notice)  : ;;
       Warning) : ;;
       Error)   : ;;
       Off)     : ;;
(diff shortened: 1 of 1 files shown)


rdlrt (approved):

scripts/cnode-helper-scripts/ogmios.sh
```diff
@@ -37,6 +38,19 @@ set_defaults() {
   [[ -z "${OGMIOSBIN}" ]] && OGMIOSBIN="${HOME}"/.local/bin/ogmios
   [[ -z "${HOSTADDR}" ]] && HOSTADDR=127.0.0.1
   [[ -z "${HOSTPORT}" ]] && HOSTPORT=1337
+  if [[ -z "${LOG_LEVEL}" ]]; then
+    LOG_LEVEL=Notice
+  else
+    case ${LOG_LEVEL} in
+      Debug)   : ;;
+      Info)    : ;;
+      Notice)  : ;;
```
rdlrt, line 47: can remove this line  (default already at notice)?

rdlrt (approved):

scripts/cnode-helper-scripts/ogmios.sh
Scitz0: The case statement is there to "accept" all valid entries and default to set log level to Notice if invalid log level is set in user config section. There needs to be a correct log level set else the execution argument will fail. The other option would be to put entire thing in variable, ie including `--log-level`, but not sure if that adds any value?

PR #1739, opened 2024-03-13
Add node-deps.json for node dependencies when running guild-deploy.sh

rdlrt:
## Description

IO teams have [updated documentation](https://github.com/input-output-hk/cardano-node-wiki/pull/40/files) to use IOHK Nix references for build dependencies. However:
1. these dependencies are not consistent between 8.7.3 and 8.9.0 (8.7.3 has references to `rev` while 8.9.0 has `ref` for some components).
2. Also, we've burnt ourselves a few times trying to rely on standardisation on that repo as it can change with time.

Hence, with this PR - we start maintaining our own node dependencies matrix for components and start filling/using versions from there instead, with fallback to hardcoded references should someone try to build using custom commits. This should help have control when bumping for patches so that same scripts working on x commit does not stop working after few days without any change in this repo.


Scitz0 (approved):
👍

PR #1741, opened 2024-03-17
Mithril-client-updates-and-docs

TrevorBenson:
Updates the mithril-client wrapper script to provide access to each of the mithril-client binaries commands and subcommands for ease of use (provides all the parameters and arguments for each command).

Includes documentation on using the mithril-client.sh to download a latest snapshot, list all snapshots, or show a specific one, as well as download or list available stake-distributions. List and show commands allow json output formats for integration with other tools/purposes.

### DIFF
--- a/docs/Mithril/mithril-overview.md
+++ b/docs/Mithril/mithril-overview.md
@@ -1,0 +1,231 @@
+# Mithril Overview
+
+[Mithril Networks](https://mithril.network/docs) provide the ability to download and bootstrap cardano nodes via snapshots of the the Cardano blockchain. This is a great way to speed up the process of syncing a new node, especially for stake pool operators. The tools provided by Guild Operators are designed to facilitate the ease of use
+in setting up and managing the:
+
+- [Mithril Client](https://mithril.network/doc/mithril/mithril-network/client) to
+download a snapshot for the given network the node is attached to via the
+[mithril-client.sh](../Scripts/mithril-client.md) script.
+- [Mithril Signer](https://mithril.network/doc/mithril/mithril-network/signer) to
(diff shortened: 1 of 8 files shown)


Scitz0:
A few notes looking at the PR.

- I can see that PR adds information for how the updated mithril-client.sh script operates, but I'm missing the big picture. In my head, I envision a Mithril subsection within the guild docs. This would have a couple of subsections for high-level what mithril is and why it's needed. Just a few sentences. Then installation, client script, and signer(+relay) script. Ie, to deploy mithril you should run guild-deploy.sh with -s m option with information about what this will download and do. Then a section about the client and how it operates, sort of what you added in docs to this PR. But maybe also the cnode.sh integration and env setting omitted what I could see. The last section would be for SPO's to deploy signer and possibly relay. For general information we can just refer to mithril site. In our docs it would be more high level and how the scripts would be used to deploy a "normal" sign setup.
- As client script now operates differently, cnode.sh needs to be updated as well?

TrevorBenson:
@rdlrt I've got work underway to resolve the comments by @Scitz0 to provide a global view of Mithril, reference the upsteam docs, and explain the various architectures that can be implemented with the mithril scripts.

Scitz0 (changes requested):
Looks really good. One thing I noticed was that sidebar.md file had path to mithril-overview as `Scripts/mithril-overview.md` while actual file seems to reside in `Mithril/`. I guess changing sidebar.md file would be the easier thing as overview isn't just about the scripts but also general information so fine to keep in separate dir outside of Scripts to me.

TrevorBenson:
@Scitz0 Should be ready for another review.

Scitz0 (approved):
:+1:

PR #1746, opened 2024-04-02
CNCli wrapper script fixes

Scitz0:
## Description
Node updates has changed format of stake-snapshot output so node mode failed.
There was also an issue with nonce check.


hodlonaut (approved):
looks okay

Issue #1744, opened 2024-03-27
[Mithril] - Add version check if update_check is set to 'Y'


rdlrt:
**Describe the bug**

Mithril versions need to be updated - especially post a release of node release, and can easily be missed given that not everyone would use guild-deploy scripts to perform updates. Hence, the scripts themselves should be able to check the latest version of mithril from github (if UPDATE_CHECK is set to 'Y') and prompt for update - similar to cncli.sh

TrevorBenson:
I did a cursory review after work today of `env`, `cntools.sh` and `cntools.library` it appears **UPDATE_CHECK** and **checkUpdate** are specific to the scripts and not any third party release binaries. I thought there might be similar updates for other binaries in the previous question.

At the moment nothing else comes to mind to prevent breaking changes from Mithril release binaries intended only for the latest node version. I'll put something together to update binaries from the scripts without `guild-deoloy.sh` and consider if other options exist besides setting **UPDATE_CHECK** to default to N.

TrevorBenson:
@rdlrt 

I communicated with the Mithril developers on the topic of avoiding breaking changes in [input-output-hk/mithril discussion #1606](https://github.com/input-output-hk/mithril/discussions/1606) . A method should be implemented in [input-output-hk/mithril issue #1615](https://github.com/input-output-hk/mithril/issues/1615) to provide a simple way to avoid updates with breaking changes, or allow the SPO to update anyway once they understand it will lead to breaking changes between the updated Mithril binaries and the installed Cardano binaries.

After that I'll implement the logic requested in this issue.

TrevorBenson:
> **Describe the bug**
> 
> Mithril versions need to be updated - especially post a release of node release, and can easily be missed given that not everyone would use guild-deploy scripts to perform updates. Hence, the scripts themselves should be able to check the latest version of mithril from github (if UPDATE_CHECK is set to 'Y') and prompt for update - similar to cncli.sh

@rdlrt

After further review I'm a bit confused by this issues request. I checked the referenced cncli.sh script, it's **cncliInit** and logic for when **UPDATE_CHECK** is set to 'Y'. I've also reviewed the **checkUpdate** function inherited by cncli.sh from `env`. I don't see any logic where cncli.sh updates its own binary. I only see it updating the script itself, similar to other scripts, but no method for cncli.sh to actualy update the binary **cncli** it wraps.

[...]

PR #1761, opened 2024-04-17
[mithril] Fix bug in checkUpdate / Fix export for signer / Fix PARTY_ID for verification / Add mithril client download to non container docs

### LINKED ISSUE #1756: [mithril] script update check uses hardcoded cncli.sh
**Describe the bug**
The checkUpdate function uses dirname and basename, but the scripts (mithril-client and mithril-signer) that are calling it is passing in `cncli.sh`. This was copy/pasted from cncli.sh version of the check without modification.

**Expected behavior**
The script should always be updating itself, not any other scripts.

**Additional context**
I'll have a branch and PR ready for this later today once I return home. I'll review all scripts to determine if a general approach/improvement can be made.

On initial review I saw that checkUpdate() was using dirname/basename, but didnt pay close enough attention that it was $1, not $0, and therefore the values were always passed along and not presumed to be the script that actually called the function.

### LINKED ISSUE #1757: syntactic bug in entrypoint.sh
**Describe the bug**
Redirect the output of the sed command back into the script file itself using >>. This can lead to unintended consequences like duplicating or corrupting the file because sed -i modifies the file in place, and the >> operator appends the output of the command to the end of the file. 

I did not find any functional issues. Resolving this would improve robustness though. 

**To Reproduce**

https://github.com/cardano-community/guild-operators/blob/5e17fc32705d84ae4529a6ea55b181f028ab9973/files/docker/node/addons/entrypoint.sh#L40

**Expected behavior**
silence the output with '/dev/null 2>&1'

**Screenshots**
Screenshot 2024-04-14 at 11 58 06 PM

**Version:**
 - OS: Ubuntu 22
 - Product version: latest release
 - Cardano Node version: 8.7.1
 - Network you're connecting to:mainnet

**Additional context**

### LINKED ISSUE #1759: [mithril] release 2412.0 changed the mithril-client command/subcommand structure
**Describe the bug**
The snapshot download used to be called via `mithril-client` using command subcommand structures. The download was  **snapshot download**, there was also **snapshot list** and **snapshot show**. All snapshot logic is now contained inside **cardano-db**, but has also been restructured so **download** is a direct subcommand of **cardano-db** (i.e. **cardano-db download**) while the **show** and **list** subcommands remain as part of **snapshot**.

The current scripts still use the old method of **snapshot download** which is being deprecated.

```
cardano-db
    download
    snapshot
        list
        show
```
Other commands like **mithril-stake-distribution** appear unchanged.

**Additional context**
The scripts should use the updated structure to future proof for when full deprecation removes **snapshot** as a command and **cardano-db download** is the only method to download the database (snapshot).

TrevorBenson:
1. Includes a syntax fix for #1757
2. Removes unecessary UPDATE_CHECK=N exports in the workflow now that it is a default value in the Dockerfile ENV / Image.
3. Exports G_ACCOUNT before calls to `./guild-deploy.sh` to ensure if images are built in a fork that the files come from the fork, not **cardano-community**.
4. Sets default G_ACCOUNT and GUILD_DEPLOY_BRANCH so manual `docker build` commands do not require passing `--build-arg`'s except to alter default values.
5. Fixes a bug leftover in the update logic and abstracts common functions from mithril scripts into a new mithril.library  sourced by the scripts, reduces maintenance overhead and code duplication.
6. Prepares for #1744 by getting the minimum versions for cardano node from mithrils new `networks.json` file.
7. Implements changes suggested by @Fuma419 since branch node-8.10.0 will merge into alpha, but not node-8.9.0.
8. Fixes a bug leftover in the update logic

### 1756 update bug
* This changes the call to use `$(basename "$0")` for any instance of a mithril script calling checkUpdate for itself, should future proof against script renames.
  * Also runs checkUpdate for `env` and `mithril.library`.

### 1759 changes to mithril-client
* The cardano-db is now the base command
  * snapshot is relocated inside cardano-db
  * download is relocated from snapshot directly into cardano-db
* Implemented cleanup on crashes as well as user interupts. A new node should either have a full copy of the most recent snapshot, or be empty so `cnode.sh` has no corruption or incomplete files preventing a good sync.

* This also changes `mithril-signer.sh`
  * Existing flags in `mithril-signer.sh` changed to have a better overlap in mithril scripts using similar OPTARGS (-u always skip updates, etc.).
  * New flags implemented that integrates the upstream mithril scripts to Verify signer registration and signer signatures.

Documentation has been updated to account for changes in the scripts usage, flags, commands and subcommands. Also the MITHRIL_DOWNLOAD variable is documented into the `node-cli.md`.

closes #1756
closes #1757
closes #1759

### DIFF
--- a/scripts/cnode-helper-scripts/guild-deploy.sh
+++ b/scripts/cnode-helper-scripts/guild-deploy.sh
@@ -638,6 +638,7 @@ populate_cnode() {
   updateWithCustomConfig "mithril-client.sh"
   updateWithCustomConfig "mithril-relay.sh"
   updateWithCustomConfig "mithril-signer.sh"
+  updateWithCustomConfig "mithril.library"
   
   find "${CNODE_HOME}/scripts" -name '*.sh' -exec chmod 755 {} \; 2>/dev/null
   chmod -R 700 "${CNODE_HOME}"/priv 2>/dev/null
--- a/docs/Build/node-cli.md
+++ b/docs/Build/node-cli.md
@@ -63,7 +63,7 @@ POOL_NAME="GUILD"
 
 #### Start the node
 
-To test starting the node in interactive mode, we will make use of pre-built script `cnode.sh`. This script automatically determines whether to start the node as a relay or block producer (if the required pool keys are present in the `$CNODE_HOME/priv/pool/<POOL_NAME>` as mentioned above). The script contains a user-defined variable `CPU_CORES` which determines the number of CPU cores the node will use upon start-up:
+To test starting the node in interactive mode, we will make use of pre-built script `cnode.sh`. This script automatically determines whether to start the node as a relay or block producer (if the required pool keys are present in the `$CNODE_HOME/priv/pool/<POOL_NAME>` as mentioned above). If the `<MITHRIL_DOWNLOAD>` variable is set to 'Y' it will download the latest snapshot from a Mithril aggregator to speed up the blockchain synchronization. The script contains a user-defined variable `CPU_CORES` which determines the number of CPU cores the node will use upon start-up:
 
 ```bash
 ######################################
--- a/docs/Scripts/mithril-client.md
+++ b/docs/Scripts/mithril-client.md
@@ -1,9 +1,36 @@
 `mithril-client.sh` is a script to manage the Mithril client, a tool used to set up the Mithril client environment and manage downloading Mithril snapshots and stake distributions. The main features include:
 
 - **environment** - Creates a new `mithril.env` file with all the necessary environment variables for the Mithril client.
-- **snapshot** - Download, list all or show a specific available Mithril snapshot.
+- **cardano-db** - Download, list all or show a specific available Mithril snapshot.
 - **stake-distribution** - Download or list available Mithril stake distributions.
 - **-u** - Skip script update check.
+
+## Usage
+
+```bash
+Usage: bash [-u] <command> <subcommand> [<sub arg>]
+A script to run Cardano Mithril Client
+
+-u          Skip script update check overriding UPDATE_CHECK value in env (must be first argument to script)
+    
+Commands:
+environment           Manage mithril environment file
+  setup               Setup mithril environment file
+  override            Override default variable in the mithril environment file
+  update              Update mithril environment file
+cardano-db            Interact with Cardano DB
+  download            Download Cardano DB from Mithril snapshot
+  snapshot            Interact with Mithril snapshots
+    list              List available Mithril snapshots
+      json            List availble Mithril snapshots in JSON format
+    show              Show details of a Mithril snapshot
+      json            Show details of a Mithril snapshot in JSON format
+stake-distribution    Interact with Mithril stake distributions
+  download            Download latest stake distribution
+  list                List available stake distributions
+    json              Output latest Mithril snapshot in JSON format
+
+```
 
 ## Preparing a Relay or Block Producer Node
 
@@ -18,7 +45,7 @@ To prepare a relay or block producer node, you should follow these steps:
 2. **Download the latest Mithril snapshot:** Once the environment file is set up, you can download the latest Mithril snapshot by running the script with the `snapshot download` command. This snapshot contains the latest state of the Cardano blockchain db from a Mithril Aggregator.
 
    ```bash
-   ./mithril-client.sh snapshot download
+   ./mithril-client.sh cardano-db download
    ```
 
 ## Investigating Available Snapshots
@@ -28,16 +55,16 @@ You can investigate the available snapshots by using the `snapshot list` and `sn
 - **List all available Mithril snapshots:** You can list all available Mithril snapshots by running the script with the `snapshot list` command. Add `json` at the end to get the output in JSON format.
 
   ```bash
-  ./mithril-client.sh snapshot list
-  ./mithril-client.sh snapshot list json
+  ./mithril-client.sh cardano-dbsnapshot list
+  ./mithril-client.sh cardano-dbsnapshot list json
   ```
 
 - **Show details of a specific Mithril snapshot:** You can show details of a specific Mithril snapshot by running the script with the `snapshot show <DIGEST>` command, where `<DIGEST>` is the digest of the snapshot. Add `json` at the end to get the output in JSON format.
 
   ```bash
-  ./mithril-client.sh snapshot show <DIGEST>
-  ./mithril-client.sh snapshot show <DIGEST> json
-  ./mithril-client.sh snapshot show json <DIGEST>
+  ./mithril-client.sh cardano-dbsnapshot show <DIGEST>
+  ./mithril-client.sh cardano-dbsnapshot show <DIGEST> json
+  ./mithril-client.sh cardano-dbsnapshot show json <DIGEST>
   ```
 
 ## Managing Stake Distributions
--- a/docs/Scripts/mithril-relay.md
+++ b/docs/Scripts/mithril-relay.md
@@ -7,12 +7,13 @@ loadbalancer. It provides functionalities such as:
 ## Usage
 
 ```bash
-Usage: mithril-relay.sh [-d] [-l]
+bash [-d] [-l] [-u] [-h]
+A script to setup Cardano Mithril relays
 
-Options:
-    -d  Install squid and configure as a relay
-    -l  Install nginx and configure as a load balancer
-    -h  Show this help text
+-d  Install squid and configure as a relay
+-l  Install nginx and configure as a load balancer
+-u  Skip update check
+-h  Show this help text
 ```
 
 # Description
--- a/docs/Scripts/mithril-signer.md
+++ b/docs/Scripts/mithril-signer.md
@@ -5,12 +5,17 @@ environment file to contain variables specific to the Mithril Signer.
 ## Usage
 
 ```bash
-Usage: mithril-signer.sh [-d] [-u]
+Usage: bash [-d] [-D] [-e] [-k] [-r] [-s] [-u] [-h]
+A script to setup, run and verify Cardano Mithril Signer
 
-Options:
-    -d    Deploy mithril-signer as a systemd service
-    -u    Update mithril environment file
-    -h    Show this help text
+-d    Deploy mithril-signer as a systemd service
+-D    Run mithril-signer as a daemon
+-e    Update mithril environment file
+-k    Stop signer using SIGINT
+-r    Verify signer registration
+-s    Verify signer signature
+-u    Skip update check
+-h    Show this help text
 ```
 
 # Description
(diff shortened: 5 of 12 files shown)


TrevorBenson:
Taking this back out of draft. After further testing the edge case I thought I may have found seems specific to the container setup, not the branch or the images it generates.

rdlrt:
The only bit am unsure is what does the title refer to (cncli.sh)? That file is not touched in the PR

rdlrt (approved):

TrevorBenson:
I found an edge case that go hidden by the final exit message being `>>` redirected into the log instead of `| tee`. The sourcing of the mithril.env doesn't export the vars for mithril signer, and it doesn't have flags to explicitely pass each value like mithril-client does, so it ends up missing CARDANO_CLI_PATH and other variables.

@rdlrt Since you approved already I wanted to bring commit eac10bf to your attention. This loops over each line in `mithril.env` and exports them for the signer. I also changed logging from `>>` to `| tee -a` so that its not hidden from the user in the log if an issue occurs. This should result also make the `journalctl` container all the signer log content for review

TrevorBenson:
@Scitz0 Support for Sanchonet added to the current PR as discussed/requested.
* `guild-deploy.sh` has been modified to accept **CARDANO_NODE_VERSION** from the environment when defined.
* Sanchonet / Preview often relies on the pre-release binaries for either mithril, cardano node, or both.
  * Sanchonet branch
    * Created from and rebased continuously on alpha.
    * Uses the latest pre-release of cardano-node binaries.
    * Uses the unstable release of mithril binaries (as suggested from mithril).
  * Preview
    * Created from and rebased continuously on alpha.
    * Uses the latest full release of cardano-node binaries.
    * Uses the pre-release of mithril binaries (as suggested from mithril).
      * Except when `"${release}-pre" == "${prerelease} "`, then it defaults to using the full release binaries.
* The Docker Image, `docker_bin.yml`, workflow has been modified to account for new testing branches.
  * Uses the CNVERSION from the `files/docker/node/release-versions/cardano-node-prerelease.txt` contained in the sanchonet branch.
  * Sanchonet uses configs from cardano-community, so may result in incompatible configs to pre-release node when breaking changes exist in the pre-release.
    * An alternate approach might be to get sanchonet configs from the [book.world.dev.cardano.org environments](https://book.world.dev.cardano.org/environments.html). However I've left this open for discussion since breaking changes in configs are not a guarantee from a pre-release, just a posibility.
  * Preview uses the CNVERSION from the `files/docker/node/release-versions/cardano-node-latest.txt` contained in the preview branch.
* A new **Autoupdate testing branches** `autoupdate-testing.yml` workflow has been added.
  * Fetches the cardano-node and mithril pre-release versions and updates them in the sanchonet and preview branches.
  * Fetches the mithril unstable version and updates it in the sanchonet branch.
    * Currently this is `unstable` and does not change between releases. Logic included in case mithril changes to a `XXXX.Y-unstable` naming convention similar to what is used in pre-release versions, in which case the workflow and CI builds should continue to function as expected.
  * Rebases both preview and sanchonet on the alpha branch.
  * If any changes occur throughout the workflow the branches are updated. If no change occurs throughout the entire worflow no update/push occurs.

@rdlrt This PR has additional content added around sanchonet, workflow changes, and modifications to `guild-deploy.sh`. As such I am clearing the current approval and requesting an additional review. Please let me know if you have any questions or suggestions.

Thanks

scripts/cnode-helper-scripts/guild-deploy.sh
```diff
@@ -371,7 +371,7 @@ download_cnodebins() {
   pushd "${HOME}"/tmp >/dev/null || err_exit
   echo -e "\n  Downloading Cardano Node archive created from GitHub.."
   rm -f cardano-node cardano-address
-  curl -m 200 -sfL https://github.com/intersectmbo/cardano-node/releases/download/${CARDANO_NODE_VERSION}/cardano-node-${CARDANO_NODE_VERSION}-linux.tar.gz -o cnode.tar.gz || err_exit " Could not download cardano-node release ${CARDANO_NODE_VERSION} from GitHub!"
+  curl -m 200 -sfL https://github.com/intersectmbo/cardano-node/releases/download/${CARDANO_NODE_VERSION}/cardano-node-${CARDANO_NODE_VERSION%-pre}-linux.tar.gz -o cnode.tar.gz || err_exit " Could not download cardano-node release ${CARDANO_NODE_VERSION} from GitHub!"
```
rdlrt, line 374: I assume this got added in as an error, but even if so :

I dont think we want to entertain use of `-pre` (or any non-stable releases) in any of the core scripts that aim to merge to alpha/master unless they're in a very specific `if` condition , with a TODO comment to remove in future.

rdlrt:
@TrevorBenson - comments inline:

> * Sanchonet / Preview often relies on the pre-release binaries for either mithril, cardano node, or both.
>   
>   * Sanchonet branch
>     * Created from and rebased continuously on alpha.
>     * Uses the latest pre-release of cardano-node binaries.
>     * Uses the unstable release of mithril binaries (as suggested from mithril).
>   * Uses the CNVERSION from the `files/docker/node/release-versions/cardano-node-prerelease.txt` contained in the sanchonet branch.
>   * Sanchonet uses configs from cardano-community, so may result in incompatible configs to pre-release node when breaking changes exist in the pre-release.

I dont think node for preview itself should rely on pre-release node binaries, it does not have fork dependency (where stable version doesnt work post a hard fork) atleast. Accordingly, I'd propose remove preview branch completely. I am OK with sanchonet branch and it's status above (as long as core scripts on alpha/master are not changed to add an error when there's a breaking change on test networks) 👍🏼 Accordingly, my comment to remove `-pre` reference in `guild-deploy.sh`

>     * An alternate approach might be to get sanchonet configs from the [book.world.dev.cardano.org environments](https://book.world.dev.cardano.org/environments.html). However I've left this open for discussion since breaking changes in configs are not a guarantee from a pre-release, just a posibility.

I am OK with using configs from book.play (it's quite confusing to have book.world vs book.play, but as per my understanding - more automated updates for pre-release go to book.play). The node binary itself is a bit of a playground atm, so I am not too worried about sanchonet nodes breaking between commits.

>   * Preview uses the CNVERSION from the `files/docker/node/release-versions/cardano-node-latest.txt` contained in the preview branch.

Preview should stick to stable node release (it is OK to use test/pre-release mithril versions), thus - should not require seperate branch.

> The ~`cnode.sh`~ `env` script appears to have strong dependencies on the koios script support. The preview branch works and only uses the version in alpha. However, the sanchonet container using the suggested node and mithril binary versions results in an error on cnode.sh startup:
> 
> When asked about adding support for sanchonet I had not noticed the versionCheck for versions Koios is compatible with. My first thought is to add a environment variable which allows bypassing this check on sanchonet network. However I figured I would open this up for discussion in case there was a preference on the way to handle this for sanchonet.
> 
> @rdlrt This PR has additional content added around sanchonet, workflow changes, and modifications to `guild-deploy.sh`. As such I am clearing the current approval and requesting an additional review. Please let me know if you have any questions or suggestions.


`versionCheck` can be overridden by environment variable `STRICT_VERSION_CHECK`. Could be something for the workflow.

rdlrt (changes requested):

rdlrt:
At a higher level, I'd recommend to keep the PR regarding sanchonet addition seperate - mithril script changes could've been easier to merge =]

scripts/cnode-helper-scripts/guild-deploy.sh
TrevorBenson: This is not a mistake. The binaries/tarball does not contain the `-pre` portion of the pre-release name.

For example a release of `8.9.2` results in a URL of:
```
https://github.com/intersectmbo/cardano-node/releases/download/8.9.2/cardano-node-8.9.2-linux.tar.gz
```
However, for the pre-release of `8.10.1-pre` the URL is:
```
https://github.com/intersectmbo/cardano-node/releases/download/8.10.1-pre/cardano-node-8.10.1-linux.tar.gz
```
Without the `%-pre`  an error occurs trying to find the `cardano-node-8.10.1-pre-linux.tar.gz` tarball to download, which does not exist. I believe this is in preparation to move pre-releases to the full/final release. When no other commits/changes are merged the existing binaries could persist as the full release instead of rebuilding them from scratch only to apply a different name to the file.

rdlrt: But as said earlier, we wouldn't want `-pre` as default in guild-deploy.sh, it is OK to check and add that only for sanchonet (against NETWORK), but other networks should use proper stable versions

TrevorBenson:
> `versionCheck` can be overridden by environment variable `STRICT_VERSION_CHECK`. Could be something for the workflow.

:+1: I had overlooked the `STRICT_VERSION_CHECK`, but I tested this out yesterday on Sanchonet and it works perfectly.

TrevorBenson:
> I am OK with using configs from book.play (it's quite confusing to have book.world vs book.play, but as per my understanding - more automated updates for pre-release go to book.play). The node binary itself is a bit of a playground atm, so I am not too worried about sanchonet nodes breaking between commits. The only note is that using config from book.play will require adding config file manipulation to substitute logging and change some Tracers so that it doesnt break tools that read logs

I'll retool this for book.play and look for changes that would have been missed when comparing book.world to configurations in the repo.

TrevorBenson:
> > * Preview uses the CNVERSION from the `files/docker/node/release-versions/cardano-node-latest.txt` contained in the preview branch.
> 
> Preview should stick to stable node release (it is OK to use test/pre-release mithril versions), thus - should not require seperate branch.

I'll split changes for handling pre-releases of cardano node and mithril binaries required in `guild-deploy.sh` out of alpha and into only sanchonet/preview. The workflow to automatically rebase these branches on changes to alpha should suffice to keep it functional until merge conflicts arise, in which case I will manually resolve them.

That said, the intent to run only a stable release on the Preview network doesn't quite match upstream suggestions. Both teams state that Preview is for testing pre-release (release candidates):
* Cardano states it on the [cardano-testnets environments page](https://docs.cardano.org/cardano-testnets/environments#preview).
* Mithril states it on the [Mithril guide Run a Mithril signer as an SPO](https://mithril.network/doc/manual/getting-started/run-signer-node/). 

Creating a Docker Image from the `preview` (or `sanchonet`) branch should create a **non public** image. Per GitHub docs contributors to cardano-community/guild-operators won't even have access to the images. Someone with full control of the account would need to grant individual contributors access on the [cardano community packages page](https://github.com/cardano-community?tab=packages). Or an SPO could build them in their own fork. 

I believe this covers both use case :
* Running a stable releases on Preview:
  * Using `-b master`,  with stable/merged guild operators tools.
  * Using `-b alpha`, to test out the latest changes to guild operators tools
* Running release candidates on Preview:
  * By using `-b preview`, which may result in non compatible combinations of cardano-node and cntools, cncli, gLiveView, but should be compatible for prerelease node and mithril client/signer.

We can discuss this further once a new PR is opened since the work is no longer inside this branch/PR.

rdlrt (approved):

rdlrt:
> That said, the intent to run only a stable release on the Preview network doesn't quite match upstream suggestions

While that maybe (I dont think I agree, some releases are specifically made available for sancho and later on said OK to run on preview) true for them, it does have a direct impact on scripts on this repository. The breaking changes added on releases need adjustments beyond download/run to be able to support. Having a higher version downloaded will automatically cause cascading effect of abundant requests to add support in scripts (which personally over the years I am not keen to - as the goal posts and comms can change quickly).

Another side-effect of having something ready is folks will reuse it preprod/mainnet too - we would not want to contribute to network issues. Sancho network is easy to reset if things go south, same cant be said for preview, as cardano does not have a comparable testnet with long enough history for any performance impact assessment on mainnet.

PR #1767, opened 2024-05-15
CNCLI leaderlog fixes

Scitz0:
Increase delay to 10 min for leaderlog calculation as it has been noticed that bad data has been calculated sometimes.

Prune db before adding new slots if forced leaderlog


rdlrt (approved):

rdlrt:
(to be merged after addressing shellcheck error)

PR #1771, opened 2024-05-22
Replace all references to yum with dnf.

TrevorBenson:
## Description
This PR replaces references to yum with dnf.

## Motivation and context
The alias from yum to dnf was originally described as stable to use until future deprecation. It appears it has disappeared from some builds or is no longer installed by default in cloud or container images. 

## Which issue it fixes?
Pre-merge failures in CI for RockyLinux images.

## How has this been tested?
Should be tested by kicking off a pre-merge test when the PR is opened.


TrevorBenson:
Seems something else is preventing the premerge from rocky, either in addition to the reported issue with yum or possibly the root cause. I'll check into this shortly.

rdlrt (approved):

PR #1774, opened 2024-05-31
Provisioning script updates

rdlrt:
### Guild Deploy changes:
- Only run build dependencies as part of os dependencies when included as flag
- Adapt libsodium flag to only use os_dependencies (and include minimal set of dependency there) to make it faster/lighter to deploy libsodium fork on block producing nodes
- Include os_dependencies prior to running cardano-hw-cli download (to ensure libusb dependencies are available)
- Allow CARDANO_NODE_VERSION to be set as environment variable for those who'd like to test guild-deploy against different node release
- Bump CARDANO_NODE_VERSION to 8.9.3

### files/docker:
- Remove docker stage files
- Remove cardano-wallet, as it is no longer relevant in the ecosystem now
- for dockerfile_bin, remove tcpping (was unused due to comment), and remove cleanup references for build packages, as they're no longer deployed with `-s p` flag. Also, remove gnupg git from initial run, as they're already included as part of `guild-deploy.sh -s -p` call

### setup-grest.sh:
- Dont use `command -v haproxy` (as the shell isnt always refreshed and /usr/sbin may not be part of $PATH), instead look for /usr/sbin/haproxy to verify if haproxy was already installed
- Add haproxy installer dependencies (if guild-deploy is only run with os_dependencies flag, haproxy wouldnt compile in it's current state)


TrevorBenson (approved):
LGTM. 

I suspect Systemd may be able to be removed from the container. I'll review that separately and if there is no real usage or dependency inside the container open a new PR for testing to further slim down the container, I'll include others in review to make sure I'm not overlooking some special use cases.

TrevorBenson:
> LGTM. 
> 
> I suspect Systemd may be able to be removed from the container. I'll review that separately and if there is no real usage or dependency inside the container open a new PR for testing to further slim down the container, I'll include others in review to make sure I'm not overlooking some special use cases.

As discussed, haproxy or great might be related to Systemd in container. I'll play with those in looking into possible additional slimming options.

Scitz0 (approved):

PR #1782, opened 2024-07-24
[Feature] - cncli.sh - integration of full stakepool history load into epochdata table

asnakep:
Hello CNTools/Koios Team,

I would like to propose the integration in cncli.sh of a process I wrote to perform a full load of epochdata table in blocklog db.

It checks all previous epochs where your pool got at least one block based on the query "SELECT DISTINCT epoch FROM blocklog."

It also takes into account the current epoch and the next epoch if it's found in the blocklog table.

Depending of the pool history it can take some minutes or a lot of hours.
It took 2 hours and ~30 minutes to process 138 epochs.

It is shown in cncli.sh menu as:

> epochdata   Manually re-calculate leaderlog to load stakepool history into epochdata table of blocklog db. Needs completion of cncli.sh sync process.
&nbsp;&nbsp;&nbsp;all       One-time re-calculation of all epochs (avg execution duration: 1hr / 50 epochs)
&nbsp;&nbsp;&nbsp;epoch     One-time re-calculation for the specified epoch
  
Following the logic of validate all and validate epoch, I added the feature to process a single epoch, with overwrite if epoch already exists.

Altough the most important is epochdata all which requires to run only once, or more if is needed for some reason.

Since it should run only once and optionally I didn't create a systemd service for this function in script, let me know if you want me to add it.

Thank you,
kind regards,
Manuel

## Description
Added new function cncliEpochData in cncli.sh

Run: "cncli.sh epochdata all" or "cncli.sh epochdata epoch"

## Motivation and context
The scenario is this: An operator runs a stakepool for months or years without using cntools/cncli and makes blocks during this period. At some point, the operator starts using cntools/cncli, and when cncli initializes, all block schedules from the first pool block until the last are entered into the blocklog table in blocklog.db. Meanwhile, the epochdata table is updated at every slot leader check, so you only have data in epochdata from when you started using CNTools, not from before.

## How has this been tested?
New function cncliEpochData has been extensively tested during mainnet epochs 497 and 498.

Needed variables in cncli.sh.

>POOL_ID
POOL_ID_BECH32


scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -799,8 +804,271 @@ cncliPTsendslots() {
...
+
+chooseConsensus() {
+  if (( EPOCH <= 363 )); then
+      consensus="tpraos"
+  elif (( EPOCH >= 364 )); then
+      consensus="praos"
+  else
+      consensus="cpraos"
+  fi
+}
```
Scitz0, line 919: hardcoded for mainnet, fetch protocol version from Koios for epoch in question instead and use existing getConsensus function.
I propose the following changes:
- change cncli.sh getConsensus function and add a check if epoch is provided as argument to function.
- If so, call koios epoch_params function instead to get major and minor version for that epoch. Else, call normal getProtocolParams function.

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -799,8 +804,271 @@ cncliPTsendslots() {
...
+
+}
+
+loadEpochData() {
+  [[ -z "${POOL_VRF_SKEY}" ]] && POOL_VRF_SKEY="${POOL_DIR}/${POOL_VRF_SK_FILENAME}"
+  [[ -z "${CNCLI_DIR}" ]] && CNCLI_DIR="${CNODE_HOME}/guild-db/cncli"
+  [[ -z "${BLOCKLOG_DIR}" ]] && BLOCKLOG_DB="${CNODE_HOME}/guild-db/blocklog"
+
+  CNCLI_DB="${CNCLI_DIR}/cncli.db"
+  BLOCKLOG_DB="${BLOCKLOG_DIR}/blocklog.db"
```
Scitz0, line 831: These are already defined by cncliInit, better to call that instead like all the other commands

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -799,8 +804,271 @@ cncliPTsendslots() {
...
+
+  return 0
+}
+
+checkEpochSlot() {
+  getNodeMetrics
+  local func=$1
+  local epochSlot=${slot_in_epoch}
+
+  if [[ $epochSlot -ge 0 && $epochSlot -le 21600 ]]; then
```
Scitz0, line 865: hardcoded mainnet value, lets make this dynamic based on values from genesis to be compatible with other networks. `read_genesis` function in env file populates various variables with data that could be used for this.

Scitz0 (commented):
First of, thanks for the contribution. I didn't have time to review it fully as there are many changes and code edits. One thing that looks strange to me in general is the fact that loadEpochData seems to be called regardless of subfunction used. Not sure I like that. 

I will reply again once I have properly looked it over. In the meantime I left a couple of comments as I saw it in the code.

scripts/cnode-helper-scripts/cncli.sh
asnakep: Thank you, in the following days I'll review the above requirements
cheers,
Manuel

Scitz0:
A general comment after looking some more at it is that its very hard to read the code with a large number of small functions and often nested with many/most of them only called once, so you have to jump back and forth in code to see whats actually going on. Spacing / indentation is also in need of a lookover to make it clearer for what belongs where. 

How I would prefer to have it rewritten would be something like this.
1. Case switcher for subcommand at the bottom is good as is. This is the thing that kicks of the subcommand. All else should be functions and no code should run unless started from here.
2. Try to limit the number of functions if they dont serve a purpose, ie only called once for a single job. Then just inline the code in the flow to not have to jump around to know whats going on. Feel free to add comments instead to explain why stuff is done. 
3. please dont drop epochdata table, just truncate it instead.
4. Create a structure similar to this:
```
# Put various helper functions needed at root level
exempleHelpFunction() {
  ...
}
cncliEpochData() {
  # some common variables and code not set by cncliInit and unique to epochdata
  if [[ ${subarg} = all ]]; then
    # code to process all epochs
    # inline as much as possible without function calls unless called more than once.
  elif isNumber "${subarg}"; then
    # code to process single epoch
  else
    # error
  fi
}
```

PS! I hope I dont sound too negative :)
I do like when people contribute, I just want to get it in line with the rest of the code to make it manageable in the future.

rdlrt:
Pushed a small update to address indentation part and get rid of duplicate variable initialisations

asnakep:
Good Morning, I proceed to close this PR in order to restart the work from cloned fork in a server instead of working with web ui.
All comments have been taken into account.
Cheers,
Manuel

rdlrt:
Thanks a lot @asnakep - apologies for the rework, happy to run through cherry-picking commits offline to assist in future

PR #1786, opened 2024-07-26
[Feature] - gLiveView - Mithril Signer Section

asnakep:
Hello CNTools/Koios Team,

I would like to propose the mithril-signer section I added in gLiveView, it can be useful signer monitoring.

Condition to activate the section is:

[[ "$nodemode" = "Core" ]] && [[ "$SERVICE_EXISTS" -eq 0 ]] && [[ "$SIGNER_SERVICE_STATUS" == "running" ]]

gLiveView with new feature has been tested on a relay too.

Thank you,
kind regards,
Manuel

## Description
mithril-signer section for gLiveView

## Updated files:
gLiveView.sh
mithril.library

- on a Block Producer: run ./mithril-signer.sh -e to recreate mithril.env with ENABLE_METRICS_SERVER=true
- restart mithril-signer service
- run gLiveView

## Motivation and context
Section has been added to have a clear view of mithril-signer status.

Compact view
image

Verbose view
image


scripts/cnode-helper-scripts/gLiveView.sh
```diff
@@ -1429,6 +1431,65 @@ while true; do
...
+          printf "${VL} Registered : ${style_values_1}%-${three_col_1_value_width}s${NC}" "$signer_reg_success_VAL"
+          printf "           : Registered Total     : ${style_values_1}%-${three_col_1_value_width}s${NC}" "$signer_reg_total_VAL"
+          closeRow
+          printf "${VL} Cycles     : ${style_values_1}%-${three_col_1_value_width}s${NC}" "$cycle_total_VAL"
+          printf "           : Signing in Epoch     : ${style_values_2}%-${three_col_1_value_width}s${NC}" "$signatures_epoch_VAL"
+          closeRow
+          printf "${VL} Signatures : ${style_values_2}%-${three_col_1_value_width}s${NC}" "$signatures_reg_success_VAL"
+          printf "           : Total Signatures     : ${style_values_1}%-${three_col_1_value_width}s${NC}" "$signatures_reg_total_VAL"
+          closeRow
+        else
```
TrevorBenson, line 1479: Moving the Registed / Registered Total might make for a better UX experience. The user will have the same field order in Verbose non Verbose modes, instead of inserting them between the Status and Cycles rows.
```suggestion
        if [[ ${VERBOSE} = "Y" ]]; then
          printf "${VL} Status     : ${style_values_2}%-${three_col_1_value_width}s${NC}" "$SIGNER_SERVICE_STATUS"
          printf "           : Registered Epoch     : ${style_values_1}%-${three_col_1_value_width}s${NC}" "$signer_reg_epoch_VAL"
          closeRow
          printf "${VL} Cycles     : ${style_values_1}%-${three_col_1_value_width}s${NC}" "$cycle_total_VAL"
          printf "           : Signing in Epoch     : ${style_values_2}%-${three_col_1_value_width}s${NC}" "$signatures_epoch_VAL"
          closeRow
          printf "${VL} Signatures : ${style_values_2}%-${three_col_1_value_width}s${NC}" "$signatures_reg_success_VAL"
          printf "           : Total Signatures     : ${style_values_1}%-${three_col_1_value_width}s${NC}" "$signatures_reg_total_VAL"
          closeRow
          printf "${VL} Registered : ${style_values_1}%-${three_col_1_value_width}s${NC}" "$signer_reg_success_VAL"
          printf "           : Registered Total     : ${style_values_1}%-${three_col_1_value_width}s${NC}" "$signer_reg_total_VAL"
          closeRow
        else
```

TrevorBenson:
It appears some editor/formatter may have done a tab to space conversion on mithril.library, which is one reason why much of the file appears to have been changed.

I talked in DM with @asnakep about the branch a bit.  He's going to do a rebase on upstream/alpha and a new push. This might clean up the review a bit and we can discuss further after that. I'll try to review the mithril changes in a local branch tomorrow, restore tabs and determine what was changed/added.

scripts/cnode-helper-scripts/mithril.library
```diff
@@ -1,196 +1,196 @@
...
+                CARDANO_NODE_SOCKET_PATH=${CARDANO_NODE_SOCKET_PATH}
+                CARDANO_CLI_PATH=${HOME}/.local/bin/cardano-cli
+                DATA_STORES_DIRECTORY=${CNODE_HOME}/mithril/data-stores
+                STORE_RETENTION_LIMITS=5
+                ERA_READER_ADAPTER_TYPE=cardano-chain
+                ERA_READER_ADAPTER_PARAMS=$(jq -nc --arg address "$(wget -q -O - "${ERA_READER_ADDRESS}")" --arg verification_key "$(wget -q -O - "${ERA_READER_VKEY}")" '{"address": $address, "verification_key": $verification_key}')
+                GENESIS_VERIFICATION_KEY=$(wget -q -O - https://raw.githubusercontent.com/input-output-hk/mithril/main/mithril-infra/configuration/${RELEASE}-${NETWORK_NAME,,}/genesis.vkey)
+                PARTY_ID=$(cat ${POOL_DIR}/${POOL_ID_FILENAME}-bech32)
+                SNAPSHOT_DIGEST=latest
+                ENABLE_METRICS_SERVER=true
```
TrevorBenson, line 162: 1. Should make this a choice of user input, like ENABLE_RELAY_ENDPOINT from lines [135-138](https://github.com/cardano-community/guild-operators/pull/1786/files#diff-222307d8e0610ac57be518231d9ac3fa96991b9da929d4f314937c0fd7e19eaaR135-R138) and [lines 165-167](https://github.com/cardano-community/guild-operators/pull/1786/files#diff-222307d8e0610ac57be518231d9ac3fa96991b9da929d4f314937c0fd7e19eaaR165-R167)

2. The `METRICS_SERVER_PORT` should also be a choice of user input. Otherwise if the signer runs on the same system as the monitoring is configured it would conflict, [setup_mon.sh Line 13](https://github.com/cardano-community/guild-operators/blob/alpha/scripts/cnode-helper-scripts/setup_mon.sh#L13) defaults to the same port.

3. Also `METRICS_SERVER_IP` from user input. This way the signer metric server does not automatically bind to all addresses if the user would prefer it not to. Also provides an option to keep the signer metric server port 9090, but but a different interface/IP than the `PROM_PORT`.

TrevorBenson (changes requested):
Turned out to be a single line change in mithril.library. A few suggested changes that could avoid monitoring conflicts, and provide the user with more control, especially if the signer and monitoring are colocated on the same system.

scripts/cnode-helper-scripts/gLiveView.sh
```diff
@@ -1429,6 +1431,65 @@ while true; do
...
+      SIGNER_SERVICE_STATUS=$(systemctl status cnode-mithril-signer.service 2>/dev/null | grep 'Active:' | awk '{print $3}' | sed 's/[()]//g')
+      MITHRIL_SIGNER_METRICS=$(curl -s http://127.0.0.1:9090/metrics 2>/dev/null | grep -v -E "HELP|TYPE" | sed 's/mithril_signer_//g')
+
+      if [[ "$nodemode" = "Core" ]] && [[ "$SERVICE_EXISTS" -eq 0 ]] && [[ "$SIGNER_SERVICE_STATUS" == "running" ]]; then
+
+        printf "${mithrildivider}\n" && ((line++))
+
+        get_metric_value() {
+            local metric_name="$1"
+            echo "$MITHRIL_SIGNER_METRICS" | awk -v metric="$metric_name" '{for(i=1;i<=NF;i++){if($i==metric){print $(i+1)}}}'
```
Scitz0, line 1445: Though awk is fairly quick, using bash internal REMATCH is even faster. For example on how this is done, check getNodeMetrics() in env file under else section. It has example for parsing prometheus output from node. 

Reason I mention this is because due to the frequent refreshes, any millisecond saved in processing, is good. 

Something else to consider, as these metrics are fairly static, would be to not re-fetch metrics on every 2s refresh, but wait for x number of refreshes between updates.

scripts/cnode-helper-scripts/gLiveView.sh
```diff
@@ -1429,6 +1431,65 @@ while true; do
     fi
   fi
 
+      # Mithril Signer Section
+      SERVICE_EXISTS=$(systemctl list-units --type=service --all | grep -q 'cnode-mithril-signer.service'; echo $?)
```
Scitz0, line 1435: Could be moved outside of loop to not have to check this on every refresh. Its also hardcoded for `cnode` name, please use `${CNODE_VNAME}` instead. 

Not a must but can check how cncli service is checked in same script to not have to pipe and grep. Though I guess fine like this as well. 
```
if command -v cncli >/dev/null && command -v systemctl >/dev/null && systemctl is-active --quiet ${CNODE_VNAME}-cncli-sync.service 2>/dev/null; then
```

rdlrt:
> It appears some editor/formatter may have done a tab to space conversion on mithril.library

I'd welcome that change tbh (except in cases where we actually need tab char, eg: <<-EOF). Keeping consistency with other scripts, 2-space indent are preferable to tab character - as otherwise the tab character rendering is inconsistent between editor's default spacing preferences

asnakep:
Good Morning, I proceed to close this PR in order to restart the work from cloned fork in a server instead of working with web ui.
All comments have been taken into account.
Cheers,
Manuel

scripts/cnode-helper-scripts/gLiveView.sh
asnakep: Hello,

I used ${CNODE_VNAME}, about "systemctl is-active" this only return if is active but not the sub-statuses of active: running, waiting, exited

PR #1789, opened 2024-07-27
[Feature] - gLiveView - Mithril Signer Section

asnakep:
Hello CNTools/Koios Team,

I would like to propose the mithril-signer section I added in gLiveView, it can be useful signer monitoring.

All suggestions from previous PR #1786 have been applied.

1. Moving the Registered / Registered Total row to the bottom might make for a better UX experience...
2. Should make this a choice of user input, like ENABLE_RELAY_ENDPOINT...
   The METRICS_SERVER_PORT should also be a choice of user input.
   Also METRICS_SERVER_IP from user input... 
3. Though awk is fairly quick, using bash internal REMATCH is even faster....
4. Could be moved outside of loop to not have to check this on every refresh... 
5. Its also hardcoded for cnode name, please use ${CNODE_VNAME} instead...

gLiveView with new feature has been tested on a relay too.

Thank you,
kind regards,
Manuel

## Description
mithril-signer section for gLiveView

## Updated files:
gLiveView.sh
mithril-signer.sh
mithril.library

on a Block Producer:
- run ./mithril-signer.sh -e to recreate mithril.env with ENABLE_METRICS_SERVER=true and chosen IP:PORT for endpoint
- restart mithril-signer service
- run gLiveView.sh

## Motivation and context
Section has been added to have a clear view of mithril-signer status.

Compact view
image

Verbose view
image

### DIFF
--- a/scripts/cnode-helper-scripts/gLiveView.sh
+++ b/scripts/cnode-helper-scripts/gLiveView.sh
@@ -261,6 +261,7 @@ if [[ ${LEGACY_MODE} = "true" ]]; then
   propdivider=$(printf "${NC}|- ${style_info}BLOCK PROPAGATION${NC} " && printf "%0.s-" $(seq $((width-21))) && printf "|")
   resourcesdivider=$(printf "${NC}|- ${style_info}NODE RESOURCE USAGE${NC} " && printf "%0.s-" $(seq $((width-23))) && printf "|")
   blockdivider=$(printf "${NC}|- ${style_info}BLOCK PRODUCTION${NC} " && printf "%0.s-" $(seq $((width-20))) && printf "|")
+  mithrildivider=$(printf "${NC}|- ${style_info}MITHRIL SIGNER${NC} " && printf "%0.s-" $(seq $((width-18))) && printf "|")
   blank_line=$(printf "${NC}|%$((width-1))s|" "")
 else
   VL=$(printf "${NC}\\u2502")
@@ -283,6 +284,7 @@ else
   propdivider=$(printf "${NC}\\u2502- ${style_info}BLOCK PROPAGATION${NC} " && printf "%0.s-" $(seq $((width-21))) && printf "\\u2502")
   resourcesdivider=$(printf "${NC}\\u2502- ${style_info}NODE RESOURCE USAGE${NC} " && printf "%0.s-" $(seq $((width-23))) && printf "\\u2502")
   blockdivider=$(printf "${NC}\\u2502- ${style_info}BLOCK PRODUCTION${NC} " && printf "%0.s-" $(seq $((width-20))) && printf "\\u2502")
+  mithrildivider=$(printf "${NC}\\u2502- ${style_info}MITHRIL SIGNER${NC} " && printf "%0.s-" $(seq $((width-18))) && printf "\\u2502")
   blank_line=$(printf "${NC}\\u2502%$((width-1))s\\u2502" "")
 fi
 
@@ -401,7 +403,7 @@ clrLine () {
 # Command    : clrScreen
 # Description: clear the screen, move to (0,0), and reset screen update counter
 clrScreen () {
-  clear
+  printf "\033[2J"
   screen_upd_cnt=0
 }
 
@@ -767,6 +769,21 @@ tcols=$(tput cols)   # set initial terminal columns
 printf "${NC}"       # reset and set default color
 
 unset cpu_now cpu_last
+
+####################################
+# Mithril Signer Section Variables #
+####################################
+
+# mithril.env sourcing needed to have values in ${METRICS_SERVER_IP} and ${METRICS_SERVER_PORT}
+. ${CNODE_HOME}/mithril/mithril.env
+
+serviceExists=$(systemctl list-units --type=service --all | grep -q ${CNODE_VNAME}-mithril-signer.service; echo $?)
+
+mithrilSignerVars() {
+signerServiceStatus=$(systemctl status ${CNODE_VNAME}-mithril-signer.service 2>/dev/null | grep 'Active:' | awk '{print $3}' | sed 's/[()]//g')
+metricsEnabled=$(grep -q "ENABLE_METRICS_SERVER=true" ${CNODE_HOME}/mithril/mithril.env && echo "true" || echo "false")
+mithrilSignerMetrics=$(curl -s "http://${METRICS_SERVER_IP}:${METRICS_SERVER_PORT}/metrics" 2>/dev/null | grep -v -E "HELP|TYPE" | sed 's/mithril_signer_//g')
+}
 
 #####################################
 # MAIN LOOP                         #
@@ -1428,6 +1445,70 @@ while true; do
       fi
     fi
   fi
+
+      # Mithril Signer Section
+      mithrilSignerVars
+
+      if [[ "$nodemode" = "Core" ]] && [[ "$serviceExists" -eq 0 ]] && [[ "$signerServiceStatus" == "running" ]] && [[ "$metricsEnabled" == "true" ]]; then
+
+        printf "${mithrildivider}\n" && ((line++))
+
+        get_metric_value() {
+            local metric_name="$1"
+            local metric_value
+            while IFS= read -r line; do
+                if [[ $line =~ $metric_name[[:space:]]+([0-9]+) ]]; then
+                    metric_value="${BASH_REMATCH[1]}"
+                    echo "$metric_value"
+                    return
+                fi
+            done <<< "$mithrilSignerMetrics"
+        }
+
+        metrics=(
+            "runtime_cycle_total_since_startup"
+            "signer_registration_success_last_epoch"
+            "signer_registration_success_since_startup"
+            "signer_registration_total_since_startup"
+            "signature_registration_success_last_epoch"
+            "signature_registration_success_since_startup"
+            "signature_registration_total_since_startup"
+        )
+
+        cycle_total_VAL=$(get_metric_value "runtime_cycle_total_since_startup")
+        signer_reg_epoch_VAL=$(get_metric_value "signer_registration_success_last_epoch")
+        signer_reg_success_VAL=$(get_metric_value "signer_registration_success_since_startup")
+        signer_reg_total_VAL=$(get_metric_value "signer_registration_total_since_startup")
+        signatures_epoch_VAL=$(get_metric_value "signature_registration_success_last_epoch")
+        signatures_reg_success_VAL=$(get_metric_value "signature_registration_success_since_startup")
+        signatures_reg_total_VAL=$(get_metric_value "signature_registration_total_since_startup")
+
+        if [[ ${VERBOSE} = "Y" ]]; then
+          printf "${VL} Status     : ${style_values_2}%-${three_col_1_value_width}s${NC}" "$signerServiceStatus"
+          printf "           : Registered Epoch     : ${style_values_1}%-${three_col_1_value_width}s${NC}" "$signer_reg_epoch_VAL"
+          closeRow
+          printf "${VL} Cycles     : ${style_values_1}%-${three_col_1_value_width}s${NC}" "$cycle_total_VAL"
+          printf "           : Signing in Epoch     : ${style_values_2}%-${three_col_1_value_width}s${NC}" "$signatures_epoch_VAL"
+          closeRow
+          printf "${VL} Signatures : ${style_values_2}%-${three_col_1_value_width}s${NC}" "$signatures_reg_success_VAL"
+          printf "           : Total Signatures     : ${style_values_1}%-${three_col_1_value_width}s${NC}" "$signatures_reg_total_VAL"
+          closeRow
+          printf "${VL} Registered : ${style_values_1}%-${three_col_1_value_width}s${NC}" "$signer_reg_success_VAL"
+          printf "           : Registered Total     : ${style_values_1}%-${three_col_1_value_width}s${NC}" "$signer_reg_total_VAL"
+          closeRow
+        else
+          printf "${VL} Status     : ${style_values_2}%-${three_col_1_value_width}s${NC}" "$signerServiceStatus"
+          printf "           : Registered Epoch     : ${style_values_1}%-${three_col_1_value_width}s${NC}" "$signer_reg_epoch_VAL"
+          closeRow
+          printf "${VL} Cycles     : ${style_values_1}%-${three_col_1_value_width}s${NC}" "$cycle_total_VAL"
+          printf "           : Signing in Epoch     : ${style_values_2}%-${three_col_1_value_width}s${NC}" "$signatures_epoch_VAL"
+          closeRow
+          printf "${VL} Signatures : ${style_values_2}%-${three_col_1_value_width}s${NC}" "$signatures_reg_success_VAL"
+          printf "           : Total Signatures     : ${style_values_1}%-${three_col_1_value_width}s${NC}" "$signatures_reg_total_VAL"
+          closeRow
+        fi
+      fi
+
 
   [[ ${check_peers} = "true" ]] && check_peers=false && show_peers=true && clrScreen && continue
 
--- a/scripts/cnode-helper-scripts/gLiveView.sh
+++ b/scripts/cnode-helper-scripts/gLiveView.sh
@@ -1461,7 +1461,7 @@ while true; do
       # Mithril Signer Section
       mithrilSignerVars
 
-      if [[ "$nodemode" = "Core" ]] && [[ "$serviceExists" -eq 0 ]] && [[ "$signerServiceStatus" == "running" ]] && [[ "$metricsEnabled" == "true" ]]; then
+      if [[ "${MITHRIL_SIGNER_ENABLED}" == "Y" ]] ; then
 
         printf "${mithrildivider}\n" && ((line++))
 
--- a/scripts/cnode-helper-scripts/gLiveView.sh
+++ b/scripts/cnode-helper-scripts/gLiveView.sh
@@ -777,13 +777,25 @@ unset cpu_now cpu_last
 # mithril.env sourcing needed to have values in ${METRICS_SERVER_IP} and ${METRICS_SERVER_PORT}
 . ${CNODE_HOME}/mithril/mithril.env
 
-serviceExists=$(systemctl list-units --type=service --all | grep -q ${CNODE_VNAME}-mithril-signer.service; echo $?)
-
 mithrilSignerVars() {
-signerServiceStatus=$(systemctl status ${CNODE_VNAME}-mithril-signer.service 2>/dev/null | grep 'Active:' | awk '{print $3}' | sed 's/[()]//g')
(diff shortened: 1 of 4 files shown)


scripts/cnode-helper-scripts/gLiveView.sh
```diff
@@ -60,7 +60,7 @@ setTheme() {
 # Do NOT modify code below           #
 ######################################
 
-GLV_VERSION=v1.30.0
+GLV_VERSION=v1.29.1
```
TrevorBenson, line 63: Restore the existing version. Although maybe increment to 1.30.1. I'll defer to @rdlrt or @Scitz0 on correct semver for gLiveView changes.

scripts/cnode-helper-scripts/gLiveView.sh
```diff
@@ -401,7 +403,7 @@ clrLine () {
 # Command    : clrScreen
 # Description: clear the screen, move to (0,0), and reset screen update counter
 clrScreen () {
-  clear
+  printf "\033[2J"
```
TrevorBenson, line 406: @asnakep I have not run into instances where `clear` might not be available as both debian and enterprise linux based distros will have the libncurses or ncurses-libs packages installed from guild-deploy.sh which should ensure `clear` is always available.

That said I'm not against relying on bash `printf` for screen clears.  Although, it might be nice to add a note to the description mentioning the change and reasoning for others to consider.

scripts/cnode-helper-scripts/gLiveView.sh
```diff
@@ -714,21 +716,6 @@ checkPeers() {
...
-  if [[ ${node_version} != "${running_node_version}" || ${node_rev} != "${running_node_rev}" ]]; then
-    clrScreen
-    printf "\n ${style_status_3}Node version mismatch${NC} - running version doesn't match found binary!"
-    printf "\n\n Forgot to restart node after upgrade?"
-    printf "\n\n Deployed version : ${node_version} (${node_rev}) => ${CNODEBIN}"
-    printf "\n Running version  : ${running_node_version} (${running_node_rev})\n"
-    waitToProceed && clrScreen
-  fi
-}
-
```
TrevorBenson, line 731: I think this is an artifact from the source branch starting from alpha before recent commits were merged. I would have thought your rebase off upstream/alpha would have dropped this change, bit it appears to be stored in the commit instead of cleared up by the rebase operation.

scripts/cnode-helper-scripts/gLiveView.sh
```diff
@@ -746,8 +733,9 @@ if [[ ${SHELLEY_TRANS_EPOCH} -eq -1 ]]; then
   printf "\n After successful node boot or when sync to shelley era has been reached, calculations will be correct\n"
   waitToProceed && clrScreen
 fi
-checkNodeVersion
-
```
TrevorBenson, line 750: Ditto, I believe this is unintended.

scripts/cnode-helper-scripts/gLiveView.sh
```diff
@@ -814,7 +817,9 @@ while true; do
     logln "ERROR" "${error_msg}"
     waitForInput && continue
   elif [[ ${fail_count} -ne 0 ]]; then # was failed but now ok, re-check
-    checkNodeVersion
```
TrevorBenson, line 817: Ditto, I believe this is unintended.

scripts/cnode-helper-scripts/gLiveView.sh
```diff
@@ -850,7 +855,7 @@ while true; do
     else
       cpu_util="0.0"
     fi
-    if [[ ${forging_enabled} -eq 1 ]]; then
+    if [[ ${about_to_lead} -gt 0 ]]; then
```
TrevorBenson, line 858: Ditto, I believe this is unintended.

scripts/cnode-helper-scripts/gLiveView.sh
asnakep: I didn't modify clrScreen function

TrevorBenson: hmmm, strange. 
```suggestion
  clear
```

TrevorBenson:
@asnakep 
Much cleaner. I'll give it a test here shortly and retest the mithril changes as well. :+1: Thanks for all the additions!

asnakep:
Thanks for all the help with git, cheers

scripts/cnode-helper-scripts/mithril.library
```diff
@@ -129,6 +129,14 @@ get_relay_endpoint() {
   echo "Using RELAY_ENDPOINT=${RELAY_ENDPOINT_IP}:${RELAY_PORT} for the Mithril signer relay endpoint."
```
TrevorBenson, line 129:
```suggestion
  echo "Using RELAY_ENDPOINT=http://${RELAY_ENDPOINT_IP}:${RELAY_PORT} for the Mithril signer relay endpoint."
```

scripts/cnode-helper-scripts/gLiveView.sh
```diff
@@ -768,6 +770,21 @@ printf "${NC}"       # reset and set default color
...
+# mithril.env sourcing needed to have values in ${METRICS_SERVER_IP} and ${METRICS_SERVER_PORT}
+. ${CNODE_HOME}/mithril/mithril.env
+
+serviceExists=$(systemctl list-units --type=service --all | grep -q ${CNODE_VNAME}-mithril-signer.service; echo $?)
+
+mithrilSignerVars() {
+signerServiceStatus=$(systemctl status ${CNODE_VNAME}-mithril-signer.service 2>/dev/null | grep 'Active:' | awk '{print $3}' | sed 's/[()]//g')
+metricsEnabled=$(grep -q "ENABLE_METRICS_SERVER=true" ${CNODE_HOME}/mithril/mithril.env && echo "true" || echo "false")
+mithrilSignerMetrics=$(curl -s "http://${METRICS_SERVER_IP}:${METRICS_SERVER_PORT}/metrics" 2>/dev/null | grep -v -E "HELP|TYPE" | sed 's/mithril_signer_//g')
+}
```
TrevorBenson, line 786: I've tested and it looks good and mithril script changes work as intended. I've been contemplating the way to provide support for non Systemd environments. Two thoughts come to mind. 
1. **MITHRIL_SIGNER_TYPE** `(SYSTEMD|OTHER) Type of Mithril Signer Environment`
   * This keeps the current logic where systemd service unit and service status checks are used.
   * When OTHER (anything besides SYSTEMD) is set, only the metrics port responsiveness is used for service status.
   * Unset by default.
2. **MITHRIL_SIGNER_ENABLED** `(Y|N) If the Mithril Signer is enabled`
   * Drops the specific logic for Systemd, and only relies on the metrics server port and responsiveness.
   * Unset by default, but logic defaults to `MITHRIL_SIGNER_ENABLED=N` when variable is not set.

```
mithrilSignerVars() {
  signerMetricsEnabled=$(grep -q "ENABLE_METRICS_SERVER=true" ${CNODE_HOME}/mithril/mithril.env && echo "true" || echo "false")
  echo "singerMetricsEnabled: ${signerMetricsEnabled}" >> /tmp/mithril-gliveview.log
  if [[ "${signerMetricsEnabled}" == "true" ]] ; then
    mithrilSignerMetrics=$(curl -s "http://${METRICS_SERVER_IP}:${METRICS_SERVER_PORT}/metrics" 2>/dev/null | grep -v -E "HELP|TYPE" | sed 's/mithril_signer_//g')
    echo "mithrilSignerMetrics: ${mithrilSignerMetrics}" >> /tmp/mithril-gliveview.log
    if [[ "${MITHRIL_SIGNER_TYPE}" == "SYSTEMD" ]] ; then 
      echo "MITHRIL_SIGNER_TYPE: ${MITHRIL_SIGNER_TYPE}" >> /tmp/mithril-gliveview.log

      signerServiceStatus=$(systemctl status ${CNODE_VNAME}-mithril-signer.service 2>/dev/null | grep 'Active:' | awk '{print $3}' | sed 's/[()]//g')
      echo "signerServiceStatus: ${signerServiceStatus}" >> /tmp/mithril-gliveview.log
    elif [[ -n "${MITHRIL_SIGNER_TYPE}" ]] && [[ ! "${MITHRIL_SIGNER_TYPE}" == "SYSTEMD" ]] ; then
      SIGNER_METRICS_HTTP_RESPONSE=$(curl --write-out "%{http_code}" --silent --output /dev/null --connect-timeout 2 http://${METRICS_SERVER_IP}:${METRICS_SERVER_PORT}/metrics)
      echo "SIGNER_METRICS_HTTP_RESPONSE: ${SIGNER_METRICS_HTTP_RESPONSE}" >> /tmp/mithril-gliveview.log
      if [[ "$SIGNER_METRICS_HTTP_RESPONSE" -eq 200 ]] ; then
        signerServiceStatus='running'
      else
        signerServiceStatus='inactive (dead)'
      fi
      unset SIGNER_METRICS_HTTP_RESPONSE
    fi
  fi
}
```

This change skips trying to curl the metrics server when the metrics are not enabled via the variable. Later the check would drop `serviceExists` an use instead `[[ -n "${MITHRIL_SIGNER_TYPE}" ]]`.

```
      if [[ "$nodemode" = "Core" ]] && [[ -n "${MITHRIL_SIGNER_TYPE}" ]] && [[ "$signerMetricsEnabled" == "true" ]] && [[ "$signerServiceStatus" == "running" ]] ; then
```

Given Systemd and non systemd (manual, containerized, etc.) all rely on the fact that the metrics server has to be used I'm a bit partial to the second solution. The number of branches required in **mithrilSignerVars** would be reduced quite a bit as when the metrics server is not responding the Mithril Signer status could be set to offline, no need to check the systemd service. However, I only provided code suggestion for the first option to keep it aligned with the current PR logic even though it takes more lines/logic to achieve.

@asnakep What are your thoughts?

@rdlrt @Scitz0 Do you have any preference between the two options mentioned? Does a third option come to mind?

TrevorBenson: This is untested, but I quickly tossed together what I think the function would look like if using **MITHRIL_SIGNER_ENABLED** and not relying on systemd, but simply the metrics server response.

```
mithrilSignerVars() {
  signerMetricsEnabled=$(grep -q "ENABLE_METRICS_SERVER=true" ${CNODE_HOME}/mithril/mithril.env && echo "true" || echo "false")
  if [[ "${MITHRIL_SIGNER_ENABLED}" == "Y" ]] && [[ "${signerMetricsEnabled}" == "true" ]] ; then
    mithrilSignerMetrics=$(curl -s "http://${METRICS_SERVER_IP}:${METRICS_SERVER_PORT}/metrics" 2>/dev/null | grep -v -E "HELP|TYPE" | sed 's/mithril_signer_//g')
    SIGNER_METRICS_HTTP_RESPONSE=$(curl --write-out "%{http_code}" --silent --output /dev/null --connect-timeout 2 http://${METRICS_SERVER_IP}:${METRICS_SERVER_PORT}/metrics)
    if [[ "$SIGNER_METRICS_HTTP_RESPONSE" -eq 200 ]] ; then
      signerServiceStatus='running'
    else
      signerServiceStatus='inactive (dead)'
    fi
    unset SIGNER_METRICS_HTTP_RESPONSE
  fi
}
```

scripts/cnode-helper-scripts/gLiveView.sh
```diff
@@ -1429,6 +1446,70 @@ while true; do
     fi
   fi
 
+      # Mithril Signer Section
+      mithrilSignerVars
+
+      if [[ "$nodemode" = "Core" ]] && [[ "$serviceExists" -eq 0 ]] && [[ "$signerServiceStatus" == "running" ]] && [[ "$metricsEnabled" == "true" ]]; then
```
TrevorBenson, line 1452: This line prevents gLiveview from showing the Mithril Signer is offline since it requires the state to be running.

scripts/cnode-helper-scripts/gLiveView.sh
rdlrt: Rather than extensive checks every time, regardless of whether mithril signer is being run - I'd prefer to have a variable at the top about enabling Mithril section (defaulting to 'N' ) and then only when the value for that option is not 'Y' , to start looking at options. Thus, your suggestion wrt `MITHRIL_SIGNER_ENABLED` makes sense - just need to ensure this part.

As regards with systemd v/s without, I dont want to put too many combinations in, if enabled and running on a particular port, best to simply use IP:port as a test, exclude systemd completely from code where possible - as it wont be present on docker systems anyways

asnakep: Hi Trevor and CNTools/Koios Team,

I strongly agree with your additions on check metrics response, mithril signer enabled var and signer type, just please let me know if some action is required from my side, or if this changes will be applied by you.

Thank you
Cheers,
Manuel

TrevorBenson: I'll make some adjustments and retest for MITHRIL_ENABLED. Submit a new commit later today after some additional testing.

scripts/cnode-helper-scripts/gLiveView.sh
TrevorBenson: An SPO may decide to copy the required KES hot secret key onto a relay to run a signer decoupled from their Core/Block Producer. They also may have a misconfigured, or previously setup environment missing the variable to enable the metrics server. Since there are multiple variations of how/why the SPO may adjust things compared to our expectations I'm changing this to simply verify that **MITHRIL_SIGNER_ENABLED** is Y. Then the SPO will know if its online, or if offline that something needs to be further investigated.

```suggestion
      if [[ "${MITHRIL_SIGNER_ENABLED}" == "Y" ]] ; then
```

scripts/cnode-helper-scripts/gLiveView.sh
TrevorBenson: Change now gates all steps based on **MITHRIL_SIGNER_ENABLED** being Y, as per @rdlrt suggestion we won't parse files or collect variables unless this is set, even when the metrics server is enabled by the SPO.

This also allows the SPO to run a mithril signer, but decide they want the original gLiveView layout by changing this value to N, without impacting the signer itself or their ability to query its port separately. Would require an update somewhere in documentation to describe it.

```suggestion
mithrilSignerVars() {
  # Require MITHRIL_SIGNER_ENABLED or do not proceed to source files or check ports.
  if [[ "${MITHRIL_SIGNER_ENABLED}" == "Y" ]] ; then
    # mithril.env sourcing needed to have values in ${METRICS_SERVER_IP} and ${METRICS_SERVER_PORT}
    . ${CNODE_HOME}/mithril/mithril.env
    signerMetricsEnabled=$(grep -q "ENABLE_METRICS_SERVER=true" ${CNODE_HOME}/mithril/mithril.env && echo "true" || echo "false")
    if [[ "${signerMetricsEnabled}" == "true" ]] ; then
      mithrilSignerMetrics=$(curl -s "http://${METRICS_SERVER_IP}:${METRICS_SERVER_PORT}/metrics" 2>/dev/null | grep -v -E "HELP|TYPE" | sed 's/mithril_signer_//g')
      SIGNER_METRICS_HTTP_RESPONSE=$(curl --write-out "%{http_code}" --silent --output /dev/null --connect-timeout 2 http://${METRICS_SERVER_IP}:${METRICS_SERVER_PORT}/metrics)
      if [[ "$SIGNER_METRICS_HTTP_RESPONSE" -eq 200 ]] ; then
        signerServiceStatus='online'
      else
        signerServiceStatus='offline'
      fi
      unset SIGNER_METRICS_HTTP_RESPONSE
    fi
  fi
}

```

scripts/cnode-helper-scripts/mithril-signer.sh
```diff
@@ -225,9 +224,18 @@ else
     # Run Mithril Signer Server
     echo "Starting Mithril Signer Server.."
     trap 'user_interrupt_received' INT
-    if ! "${MITHRILBIN}" -vv | tee -a "${LOG_DIR}/$(basename "${0::-3}")".log 2>&1 ; then
-      echo "Failed to start Mithril Signer Server" | tee -a "${LOG_DIR}/$(basename "${0::-3}")".log 2>&1
-      exit 1
+
+    if grep -q "ENABLE_METRICS_SERVER=true" ${CNODE_HOME}/mithril/mithril.env; then
+      METRICS_SERVER_PARAMS="--enable-metrics-server --metrics-server-ip ${METRICS_SERVER_IP} --metrics-server-port ${METRICS_SERVER_PORT}"
+      if ! "${MITHRILBIN}" ${METRICS_SERVER_PARAMS} -vv | tee -a "${LOG_DIR}/$(basename "${0::-3}")".log 2>&1 ; then
```
TrevorBenson, line 230: Enables gLiveView automatically. We could also inquire with an additional question if they want it enabled. I think when reconfiguring the environment file automatically haveing gLiveView show the signer might be useful to show if a misconfiguration exists and the signer does not start, the SPO can always disable it again in the env file.

```suggestion
    if grep -q "ENABLE_METRICS_SERVER=true" ${CNODE_HOME}/mithril/mithril.env; then
      METRICS_SERVER_PARAMS="--enable-metrics-server --metrics-server-ip ${METRICS_SERVER_IP} --metrics-server-port ${METRICS_SERVER_PORT}"
      # If ENABLE_METRICS_SERVER is true, then an environment update will enable gLiveView automatically.
      sudo sed -i "s/#MITHRIL_SIGNER_ENABLED=[YN]/MITHRIL_SIGNER_ENABLED=Y/" ${CNODE_HOME}/scripts/env
      if ! "${MITHRILBIN}" ${METRICS_SERVER_PARAMS} -vv | tee -a "${LOG_DIR}/$(basename "${0::-3}")".log 2>&1 ; then
```

TrevorBenson (changes requested):
@asnakep I've added comments with code blocks that can be applied in the UI (just remember if you apply in the UI to do a git pull if you want them in your local branch before applying further changes).

Separately the env file requires 1 additional change around [line 113](https://github.com/asnakep/guild-operators/blob/gliveview_mithril-signer/scripts/cnode-helper-scripts/env#L112-L113):

```
#MITHRIL_DOWNLOAD="N"                                   # (Y|N) Download latest Mithril snapshot
#MITHRIL_SIGNER_ENABLED="N"                             # (Y|N) Enable the gLiveView Mithril Signer Section
```

scripts/cnode-helper-scripts/gLiveView.sh
```diff
@@ -1429,6 +1446,70 @@ while true; do
...
+
+        printf "${mithrildivider}\n" && ((line++))
+
+        get_metric_value() {
+            local metric_name="$1"
+            local metric_value
+            while IFS= read -r line; do
+                if [[ $line =~ $metric_name[[:space:]]+([0-9]+) ]]; then
```
TrevorBenson, line 1460: This will resolve the linter/checks in the PR showing red, removing the shellcheck error:

```suggestion
                if [[ $line =~ ${metric_name}[[:space:]]+([0-9]+) ]]; then
```

asnakep:
Thanks so much Trevor for your changes!
Cheers,
Manuel

TrevorBenson (approved):
Approved. Thanks @asnakep for the great addition.

@rdlrt Since this touches gLiveView.sh I'd appreciate your final review/blessing prior to merge.

PR #1793, opened 2024-07-30
[Feature] - cncli.sh - integration of full stakepool history load into epochdata table

asnakep:
Hello CNTools/Koios Team,

I would like to propose the integration in cncli.sh of a process I wrote to perform a full load of epochdata table in blocklog db.

It checks all previous epochs where your pool got at least one block based on the query "SELECT DISTINCT epoch FROM blocklog."

It also takes into account the current epoch and the next epoch if it's found in the blocklog table.

Depending of the pool history it can take some minutes or a lot of hours.
It took 2 hours and ~30 minutes to process 138 epochs.

It is shown in cncli.sh menu as:

> epochdata   Manually re-calculate leaderlog to load stakepool history into epochdata table of blocklog db. Needs completion of cncli.sh sync process.
&nbsp;&nbsp;&nbsp;all       One-time re-calculation of all epochs (avg execution duration: 1hr / 50 epochs)
&nbsp;&nbsp;&nbsp;epoch     One-time re-calculation for the specified epoch
  
Following the logic of validate all and validate epoch, I added the feature to process a single epoch, with overwrite if epoch already exists.

Altough the most important is epochdata all which requires to run only once, or more if is needed for some reason.

@Scitz0 thanks a lot for your suggestions in PR 1782, it looks much better now. 
Anyway I wait for your feedback in case something needs to be reviewed.

Thank you,
kind regards,
Manuel

## Description
Added new function cncliEpochData in cncli.sh

Run: "cncli.sh epochdata all" or "cncli.sh epochdata epoch"

## Motivation and context
The scenario is this: An operator runs a stakepool for months or years without using cntools/cncli and makes blocks during this period. At some point, the operator starts using cntools/cncli, and when cncli initializes, all block schedules from the first pool block until the last are entered into the blocklog table in blocklog.db. Meanwhile, the epochdata table is updated at every slot leader check, so you only have data in epochdata from when you started using CNTools, not from before.

## How has this been tested?
New function cncliEpochData has been extensively tested during mainnet epochs 500.

A new function getProtocolParamsHist has been added in env file and added in getConsensus func of cncli.sh to handle consensus choice in previous epochs.

Needed variables in cncli.sh :

>POOL_ID
POOL_ID_BECH32


scripts/cnode-helper-scripts/env
```diff
@@ -844,6 +844,27 @@ getProtocolParams() {
   [[ ${PROT_MAJOR} -eq 0 ]] && return 1 || return 0
 }
 
+getProtocolParamsHist() {
```
Scitz0, line 847: 1. subarg is a special cncli variable, it looks like for how its called that $1 should be used here instead.
2. cli doesn't allow for historical parameter lookup, so if KOIOS API is not available function should return a failcode. Currently it will just continue and process PROT_PARAMS variable anyway. 
3. dont print error, just return with error code and let the calling code print error if needed.

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -127,11 +127,18 @@ getLedgerData() { # getNodeMetrics expected to have been already run
...
 
 getConsensus() {
-  getProtocolParams
+  if [[ -n ${subarg} ]] && [[ ${subarg} != "all" ]]; then
+    getProtocolParamsHist "${subarg}"
+  elif [[ ${subarg} == "all" ]]; then
+    getProtocolParamsHist "${EPOCH}"
+  else
+    getProtocolParams
+  fi
```
Scitz0, line 136: 1. subarg can contain any number of values, its general purpose for all subcommands to cncli. Ie it can be 'force' for leaderlog etc that also call this getConsensus function. 
Please change getConsensus function to look at input params to function and pass epoch number to it if a historical epoch is requested. 
2. This was a problem since before, but if getProtocolParamsHist or getProtocolParams returns an error code, this function should also return with an error no further processing being done.

Suggested change (including code change for all places calling this function):
```
if isNumber "$1"; then
  getProtocolParamsHist "$1" || return 1
else
  getProtocolParams || return 1
fi
```

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -797,10 +804,209 @@ cncliPTsendslots() {
...
+  sed 's/"//g' >> "$tmpcsv"
+
+  return 0
+
+}
+
+runLeaderlog() {
+ for EPOCH in "${epochs_array[@]}"; do
+   subarg="${EPOCH}"
+   getConsensus "${EPOCH}"
```
Scitz0, line 867: check for function error: 
```
if ! getConsensus "${EPOCH}"; then echo "ERROR: Failed to fetch protocol parameters for epoch ${EPOCH}.; return 1"; fi
```

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -797,10 +804,209 @@ cncliPTsendslots() {
...
+
+  ${CNCLI} leaderlog ${cncliParams} --consensus "${consensus}" --epoch="${EPOCH}" --ledger-set prev ${stake_param_prev} |
+  jq -r '[.epoch, .epochNonce, .poolId, .sigma, .d, .epochSlotsIdeal, .maxPerformance, .activeStake, .totalActiveStake] | @csv' |
+  sed 's/"//g' >> "$tmpcsv"
+
+  return 0
+
+}
+
+runLeaderlog() {
```
Scitz0, line 864: runLeaderlog never called?
Can maybe remove this function then.

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -797,10 +804,209 @@ cncliPTsendslots() {
...
+ done
+}
+
+processAllEpochs() {
+  getCurrNextEpoch
+  IFS=' ' read -r -a epochs_array <<< "$EPOCHS"
+
+  for EPOCH in "${epochs_array[@]}"; do
+    subarg="${EPOCH}"
+    getConsensus "${EPOCH}"
```
Scitz0, line 884: See other comment for same code

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -797,10 +804,209 @@ cncliPTsendslots() {
...
+   IFS=' ' read -r -a epochs_array <<< "$EPOCHS"
+
+   local EPOCH=${subarg}
+   if [[ ! " ${epochs_array[@]} " =~ " ${EPOCH} " ]]; then
+     echo "No slots found in blocklog table for epoch ${EPOCH}."
+     echo
+     exit 1
+   fi
+
+   getConsensus "${EPOCH}"
```
Scitz0, line 927: See other comment for same code

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -797,10 +804,209 @@ cncliPTsendslots() {
...
+     fi
+ done
+}
+
+processAllEpochs() {
+  getCurrNextEpoch
+  IFS=' ' read -r -a epochs_array <<< "$EPOCHS"
+
+  for EPOCH in "${epochs_array[@]}"; do
+    subarg="${EPOCH}"
```
Scitz0, line 883: Why set subarg to value of EPOCH?

Scitz0 (changes requested):
Thanks for all the code changes, its now much easier to read 👍 
Though I did find stuff to fix so please check comments added.

scripts/cnode-helper-scripts/cncli.sh
asnakep: Thank you, I'll review these points during next days

asnakep: removed 
subarg="${EPOCH}", sorry I left it from previous tests

scripts/cnode-helper-scripts/cncli.sh
asnakep: removed function runLeaderlog, it was previously used, I forgot to remove it

scripts/cnode-helper-scripts/env
asnakep: fixed and testing

scripts/cnode-helper-scripts/cncli.sh
asnakep: Hello, 

getConsensus is also called twice in cncliLeaderlog, do you want me to add this check there too?

if ! getConsensus ; then echo "ERROR: Failed to fetch protocol parameters; return 1"; fi

In epochdata process has been added as:

subarg all
if ! getConsensus "${EPOCH}"; then echo "ERROR: Failed to fetch protocol parameters for epoch ${EPOCH}.; return 1"; fi

subarg epoch
if ! getConsensus "${1}"; then echo "ERROR: Failed to fetch protocol parameters for epoch ${1}.; return 1"; fi

scripts/cnode-helper-scripts/cncli.sh
asnakep: removed too

scripts/cnode-helper-scripts/env
asnakep: getProtocolParamsHist in env file has been updated as requested.

scripts/cnode-helper-scripts/cncli.sh
asnakep: getConsensus has been updated as:

>getConsensus() {
  if isNumber "$1" ; then
    getProtocolParamsHist "$1" || return 1
  elif [[ ${subarg} == "all" ]]; then
    getProtocolParamsHist "${EPOCH}" || return 1
  else
    getProtocolParams || return 1
  fi

asnakep:
Hello @Scitz0 

I applied/fixed all requested points, I'll keep testing tomorrow for next epoch, and then I'll push the updates

Cheers,
Manuel

scripts/cnode-helper-scripts/cncli.sh
Scitz0: You can leave cncliLeaderlog as is for now, I will have a look at it later after this.

Scitz0:
@asnakep did you push changes?
I cant see edits mentioned when I look at code.

Scitz0:
> Hello @Scitz0
> 
> all requested points have been applied/fixed, I'll keep testing tomorrow for next epoch, and then I'll push the updates
> 
> Cheers, Manuel

👍 I will have a look again once pushed. Thanks for changes.

asnakep:
Hello @Scitz0 

I've just pushed the changes 

Cheers
Manuel

asnakep:
Hello @Scitz0 

I've to review an issue i have with getconsensus and next epoch, I'll do a new push when solved.

Cheers,
Manuel

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -797,10 +807,203 @@ cncliPTsendslots() {
...
+
+  return 0
+}
+
+processAllEpochs() {
+  getCurrNextEpoch
+  IFS=' ' read -r -a epochs_array <<< "$EPOCHS"
+
+  for EPOCH in "${epochs_array[@]}"; do
+    if ! getConsensus "${EPOCH}"; then echo "ERROR: Failed to fetch protocol parameters for epoch ${EPOCH}.; return 1"; fi
```
Scitz0, line 888: End quote misplaced, should be after EPOCH param and not after return.

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -797,10 +807,203 @@ cncliPTsendslots() {
...
+
+   rm $onerow_csv $tmpcsv
+ else
+   echo "No data found in $tmpcsv."
+   exit 1
+ fi
+}
+
+cncliEpochData() {
+  cncliInit
```
Scitz0, line 964: remove, cncliInit already called before cncliEpochData is called.

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -797,10 +807,203 @@ cncliPTsendslots() {
...
+  getNodeMetrics
+
+  if ! cncliDBinSync; then
+    echo ${proc_msg}
+    echo "CNCLI DB out of sync :( [$(printf "%2.4f %%" ${cncli_sync_prog})] ... check cncli sync service!"
+    exit 1
+  else
+    echo ${proc_msg}
+    local epochSlot=${slot_in_epoch}
+    local waitFromEpochstart=21600 # if within 6 hours of new epoch don't run epochdata
```
Scitz0, line 986: Need to make this network agnostic and not mainnet specific based on genesis params, ie percentage of slots in epoch or something.

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -797,10 +807,203 @@ cncliPTsendslots() {
...
+
+  row_count=$(sqlite3 "$BLOCKLOG_DB" "SELECT COUNT(*) FROM epochdata;")
+  echo "$row_count rows have been loaded into epochdata table in blocklog db"
+  echo "~ CNCLI epochdata table load completed ~"
+
+  rm $csvfile $tmpcsv
+}
+
+processSingleEpoch() {
+ if isNumber "$1"; then
```
Scitz0, line 920: This check is already done in cncliEpochData, no need to re-check that input is a number.

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -797,10 +807,203 @@ cncliPTsendslots() {
...
+ else
+   echo "No data found in $tmpcsv."
+   exit 1
+ fi
+}
+
+cncliEpochData() {
+  cncliInit
+  cncliParams="--db ${CNCLI_DB} --byron-genesis ${BYRON_GENESIS_JSON} --shelley-genesis ${GENESIS_JSON} --pool-id ${POOL_ID} --pool-vrf-skey ${POOL_VRF_SKEY} --tz UTC"
+  EPOCHS=$(sqlite3 "$BLOCKLOG_DB" "SELECT replace(group_concat(DISTINCT epoch ORDER BY epoch ASC), ',', ' ') AS EPOCHS FROM blocklog;")
```
Scitz0, line 966: Invalid SQL, change to:
```
EPOCHS=$(sqlite3 "$BLOCKLOG_DB" "SELECT replace(group_concat(DISTINCT epoch), ',', ' ') AS EPOCHS FROM blocklog ORDER BY epoch ASC;")
```

asnakep:
updated getConsensus function to use getProtocolParams on next_epoch instead of getProtocolParamsHist

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -797,10 +807,203 @@ cncliPTsendslots() {
...
+
+  rm $csvfile $tmpcsv
+}
+
+processSingleEpoch() {
+ if isNumber "$1"; then
+     getCurrNextEpoch
+     IFS=' ' read -r -a epochs_array <<< "$EPOCHS"
+
+   if [[ ! " ${epochs_array[@]} " =~ "$1" ]]; then
```
Scitz0, line 924: This logic will/can cause issues. First of, when matching against array this will be read as a long string with all epochs separated with space. Matching this against a specific number can cause issues. Like 12 will match both 12 and 122. 

Safer to just loop array and break on match. something like this:
```
unset matched
for epoch in "${epochs_array[@]}"; do
  [[ ${epoch} = "$1" ]] && matched=true && break
done
if [[ -n ${matched} ]]; then
  echo -e "No slots found in blocklog table for epoch ${EPOCH}.\n"; return 1
fi
```

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -797,10 +807,203 @@ cncliPTsendslots() {
...
+     echo "No slots found in blocklog table for epoch ${EPOCH}."
+     echo
+     exit 1
+   fi
+
+   if ! getConsensus "${1}"; then echo "ERROR: Failed to fetch protocol parameters for epoch ${1}.; return 1"; fi
```
Scitz0, line 930: quote in wrong place

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -797,10 +807,203 @@ cncliPTsendslots() {
...
+
+runPreviousEpochs() {
+  [[ -z ${KOIOS_API} ]] && return 1
+
+  local one_or_many="${EPOCH}"
+  if [[ -n "$1" && "$1" != "all" ]]; then
+    one_or_many="$1"
+  elif [[ "${subarg}" != "all" && -n "${subarg}" ]]; then
+    one_or_many="${subarg}"
+  fi
```
Scitz0, line 858: I dont see the purpose of this one. 
In processSingleEpoch() when runPreviousEpochs is called, just pass the epoch to run for as a param to runPreviousEpochs.
Do the same in processAllEpochs(), runPreviousEpochs is called for each epoch in array so can just as well pass the current epoch in loop along to runPreviousEpochs.

Scitz0 (changes requested):
Added several requested code changes now... lets revisit again once all are addressed.

scripts/cnode-helper-scripts/cncli.sh
asnakep: I tried, but epochs comes partially ordered ASC 

>464 465 466 467 468 469 470 471 472 473 474 475 476 230 252 274 282 299 311 312 316 317 318 319 320 321 322 323 324 325 326 327 328 329 330 331 332 333 334 335 339 380 387 391 392 393 394 395 396 397 398 399 400 401 402 403 404 406 407 408 409 410 411 412 413 414 415 416 417 418 419 420 421 422 423 424 425 426 427 428 429 430 431 432 433 434 435 436 437 438 439 440 441 442 443 444 445 446 447 448 449 450 451 452 453 454 455 456 457 458 459 460 461 462 463 477 478 479 480 481 482 483 484 485 486 487 488 489 490 491 492 493 494 495 496 497 498 499 500 501

do you agree with a solution like this?

>EPOCHS=$(sqlite3 "$BLOCKLOG_DB" "SELECT group_concat(epoch, ' ') AS EPOCHS FROM (SELECT DISTINCT epoch FROM blocklog ORDER BY CAST(epoch AS INTEGER) ASC);")

scripts/cnode-helper-scripts/cncli.sh
asnakep: removed

scripts/cnode-helper-scripts/cncli.sh
asnakep: fixed, sorry for the misview

scripts/cnode-helper-scripts/cncli.sh
asnakep: fixed, sorry for the misview

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -127,11 +130,18 @@ getLedgerData() { # getNodeMetrics expected to have been already run
 }
 
 getConsensus() {
-  getProtocolParams
+  if isNumber "$1" ; then
+    getProtocolParamsHist "$1" || return 1
+  elif [[ ${subarg} == "all" ]]; then
+    getProtocolParamsHist "${EPOCH}" || return 1
```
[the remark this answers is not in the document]
asnakep: removed

scripts/cnode-helper-scripts/cncli.sh
asnakep: removed cncliInit from cncliEpochData

scripts/cnode-helper-scripts/cncli.sh
asnakep: applied and works as expected just changed -n for -z in 
if [[ -n ${matched} ]]; then...

scripts/cnode-helper-scripts/cncli.sh
asnakep: applied, thank you

scripts/cnode-helper-scripts/cncli.sh
Scitz0: ahh yeah, -z is indeed the correct switch :)

scripts/cnode-helper-scripts/cncli.sh
Scitz0: Also, like you changed code to, remove replace and just add ' ' as separator to group_concat. But you should be able to order by epoch right away instead of casting to integer.

Scitz0: Another option to distinct is to use GROUP BY, it would then be something similar to this.
```
EPOCHS=$(sqlite3 "$BLOCKLOG_DB" "SELECT group_concat(epoch, ' ') AS EPOCHS FROM blocklog GROUP BY epoch ORDER BY epoch ASC;")
```

scripts/cnode-helper-scripts/cncli.sh
asnakep: Hello @Scitz0 

you mentioned this on the first pr but i forgot it, I changed as

>    read_genesis
    local epochSlot=${slot_in_epoch}
    local epochLength=${EPOCH_LENGTH}
    local remainingSlots=$(( epochLength - epochSlot))
    local waitFromEpochstart=21600 # if within 6 hours of new epoch don't run epochdata

   > if [[ $(( epochLength - remainingSlots)) -le $waitFromEpochstart ]]; then

scripts/cnode-helper-scripts/cncli.sh
Scitz0: Ok, many comments but had to test it my self as sqlite isn't my goto database. I updated example in previous comment with what works best for me. 

Ie call group_concat on subquery as it will otherwise not work correctly with distinct.

asnakep: while "select epoch FROM blocklog GROUP BY epoch ORDER BY epoch ASC" returns distinct epochs

the combnation of group_concat func and GROUP BY makes return all the epochs from blocklog

I propose this:

EPOCHS=$(sqlite3 "$BLOCKLOG_DB" "SELECT group_concat(epoch, ' ') AS EPOCHS FROM (SELECT DISTINCT epoch FROM blocklog ORDER BY epoch ASC);")

asnakep: about  PRAGMA table_info(blocklog);
yes, epoch is integer

asnakep:
Hello @Scitz0 

I just pushed last changes

cheers,
Manuele

scripts/cnode-helper-scripts/cncli.sh
Scitz0: You mention group by but example you posted use use distinct. Group by or distinct should both produce the same result so happy with whatever you like best.

scripts/cnode-helper-scripts/cncli.sh
Scitz0: It's waitFromEpochstart that's the problem. Guild network for example has 1h epochs, ie 3600 slots in total per epoch. That is what I meant by this needing to be dynamic and a percentage of total slots and not a static value of 6h.

scripts/cnode-helper-scripts/cncli.sh
Scitz0: If you are referring to me using group by, I deleted that comment as it was just wrong. Please see my first example SQL as this is now updated.

asnakep: Hello @Scitz0 

thank you, I replaced with

>EPOCHS=$(sqlite3 "$BLOCKLOG_DB" "SELECT group_concat(epoch,' ') FROM (SELECT DISTINCT epoch FROM blocklog ORDER BY epoch);")

and just pushed the change

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -127,11 +130,20 @@ getLedgerData() { # getNodeMetrics expected to have been already run
 }
 
 getConsensus() {
-  getProtocolParams
+  if isNumber "$1" ; then
+    if [[ "$1" == "$next_epoch" ]]; then
+       getProtocolParams || return 1
+    else
+       getProtocolParamsHist "$1" || return 1
+    fi
```
Scitz0, line 138: I dont like to rely on some arbitrary variable that might or might not be set outside of function. Actually, leaderlogs for an epoch should always use parameters from the previous epoch to determine the consensus used. So we might as well always use -1. This was confirmed by Sam, that nonce calculation is done at the end of previous epoch, this its that epochs parameters that determine the consensus. 

New logic:
```
if isNumber "$1" ; then
  getProtocolParamsHist "$(( $1 - 1 ))" || return 1
else
  getProtocolParams || return 1
fi
```

scripts/cnode-helper-scripts/cncli.sh
Scitz0: What is the reason for the 6h wait after epoch start by the way?

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -797,10 +809,187 @@ cncliPTsendslots() {
...
+     echo -e "No slots found in blocklog table for epoch ${EPOCH}.\n"; return 1
+   fi
+   if ! getConsensus "${1}"; then echo "ERROR: Failed to fetch protocol parameters for epoch ${1}."; return 1; fi
+   if [[ "$1" == "$curr_epoch" ]]; then
+      runCurrentEpoch
+   elif [[ "$1" == "$next_epoch" ]]; then
+      runNextEpoch
+   else
+      runPreviousEpochs
+   fi
```
Scitz0, line 924: Please pass $1 to all three functions runCurrentEpoch, runNextEpoch and runPreviousEpochs as param. 
Then as the first thing in each of these functions, do `local epoch="$1"`. Currently `${epoch}` is used in these functions but I dont think its properly set. 

Also do the same in `processAllEpochs()`, pass `${epoch}` there to all three functions. It might work anyway in some cases but this will make it much clearer when reading code.

Scitz0 (changes requested):

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -797,10 +809,187 @@ cncliPTsendslots() {
...
+processSingleEpoch() {
+   getCurrNextEpoch
+   IFS=' ' read -r -a epochs_array <<< "$EPOCHS"
+
+   unset matched
+   for epoch in "${epochs_array[@]}"; do
+     [[ ${epoch} = "$1" ]] && matched=true && break
+   done
+   if [[ -z ${matched} ]]; then
+     echo -e "No slots found in blocklog table for epoch ${EPOCH}.\n"; return 1
```
Scitz0, line 915: Should be $1 I think.

scripts/cnode-helper-scripts/cncli.sh
asnakep: cool insight, thank you

asnakep: applied

scripts/cnode-helper-scripts/cncli.sh
asnakep: Hello @Scitz0 

thanks for all your feedbacks, regarding:

>Please pass $1 to all three functions runCurrentEpoch, runNextEpoch and runPreviousEpochs as param.....

I added your solution for runPreviousEpochs only.

runCurrentEpoch uses "${curr_epoch}" from getKoiosData and runNextEpoch doesn't need the epoch parameter.

Please let me know if it is really needed also for runCurrentEpoch and runNextEpoch.

scripts/cnode-helper-scripts/cncli.sh
asnakep: I tried both "$1" and $1 and being the parameter only a single string of 3 digits number, I don't see the difference in behaviour, I can add the condition that epoch number must be 3 digits, still can be passed a larger number and get the message "No slots found in blocklog table for epoch "

I left it as $1 with no quotes anyway

asnakep: About

>waitFromEpochstart=21600 # if within 6 hours of new epoch don't run epochdata

maybe 6 hours are too much, but I noticed several times that pool_stake_snapshot point is intermittently available at least during the first three hours of a new epoch, altough I'm not sure about pool_history point

asnakep:
Hello @Scitz0 

I just pushed last changes, I'm ready to review them as per last comment

cheers,
Manuel

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -797,10 +805,187 @@ cncliPTsendslots() {
...
+  next_epoch=$((curr_epoch+1))
+}
+
+runCurrentEpoch() {
+  getKoiosData
+  getCurrNextEpoch
+  echo "Processing current epoch: ${curr_epoch}"
+  stake_param_curr="--active-stake ${active_stake_set} --pool-stake ${pool_stake_set}"
+
+  ${CNCLI} leaderlog ${cncliParams} --consensus "${consensus}" --epoch="${curr_epoch}" --ledger-set current ${stake_param_curr} |
```
Scitz0, line 824: According to cncli documentation:
> --epoch <epoch>
>             Provide a specific epoch number to calculate for and ignore --ledger-set option
So either skip `--epoch` parameter or `--ledger-set`. I dont think it matter that --ledger-set is present as --epoch will override it but makes it clearer if one or the other is skipped.

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -797,10 +805,187 @@ cncliPTsendslots() {
...
+  fi
+
+  local epoch="$1"
+  pool_stake_hist=$(jq -r '.[].active_stake' <<< "${pool_hist}")
+  active_stake_hist=$(jq -r '.[].active_stake' <<< "${epoch_hist}")
+
+  echo "Processing previous epoch: ${1}"
+  stake_param_prev="--active-stake ${active_stake_hist} --pool-stake ${pool_stake_hist}"
+
+  ${CNCLI} leaderlog ${cncliParams} --consensus "${consensus}" --epoch="${1}" --ledger-set prev ${stake_param_prev} |
```
Scitz0, line 859: Remove --ledger-set param, see other comment.

scripts/cnode-helper-scripts/cncli.sh
Scitz0: See new comment regarding --ledger-set.

scripts/cnode-helper-scripts/cncli.sh
Scitz0: > I tried both "$1" and $1 and being the parameter only a single string of 3 digits number, I don't see the difference in behaviour, I can add the condition that epoch number must be 3 digits, still can be passed a larger number and get the message "No slots found in blocklog table for epoch "
> 
> I left it as $1 with no quotes anyway

My comment here wasn't for the comparison but EPOCH variable in echo string.

scripts/cnode-helper-scripts/cncli.sh
Scitz0: I will mark this conversation as resolved, lets continue discussion in new comment.

scripts/cnode-helper-scripts/cncli.sh
Scitz0: > About
> 
> > waitFromEpochstart=21600 # if within 6 hours of new epoch don't run epochdata
> 
> maybe 6 hours are too much, but I noticed several times that pool_stake_snapshot point is intermittently available at least during the first three hours of a new epoch, altough I'm not sure about pool_history point

I will check this and get back to you.

asnakep: i missed this, I'm going to check it

asnakep: Hello @Scitz0 

I just pushed the updates

>1. updated EPOCH for $1 in
echo -e "No slots found in blocklog table for epoch ${1}.\n"

>2. requoted $1 in condition  [[ ${epoch} = "$1" ]] && matched=true && break

Cheers
Manuel

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -797,10 +805,187 @@ cncliPTsendslots() {
...
+    echo -e "No slots found in blocklog table for epoch ${1}.\n"; return 1
+  fi
+  if ! getConsensus "${1}"; then echo "ERROR: Failed to fetch protocol parameters for epoch ${1}."; return 1; fi
+  if [[ "$1" == "$curr_epoch" ]]; then
+     runCurrentEpoch ${epoch}
+  elif [[ "$1" == "$next_epoch" ]]; then
+     runNextEpoch ${epoch}
+  else
+     runPreviousEpochs ${epoch}
+  fi
```
Scitz0, line 920: Should be `$1` instead of `${epoch}` here.

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -797,10 +805,187 @@ cncliPTsendslots() {
...
+
+  local epoch="$1"
+  pool_stake_hist=$(jq -r '.[].active_stake' <<< "${pool_hist}")
+  active_stake_hist=$(jq -r '.[].active_stake' <<< "${epoch_hist}")
+
+  echo "Processing previous epoch: ${1}"
+  stake_param_prev="--active-stake ${active_stake_hist} --pool-stake ${pool_stake_hist}"
+
+  ${CNCLI} leaderlog ${cncliParams} --consensus "${consensus}" --epoch="${epoch}" ${stake_param_prev} |
```
Scitz0, line 859: As `${1}` is used everywhere else in function might as well use it here as well and remove `local epoch="$1"`

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -797,10 +805,187 @@ cncliPTsendslots() {
...
+  sed 's/"//g' >> "$tmpcsv"
+}
+
+runNextEpoch() {
+  getKoiosData
+  local epoch="$1"
+  echo "Processing next epoch: ${1}"
+  stake_param_next="--active-stake ${active_stake_mark} --pool-stake ${pool_stake_mark}"
+
+  ${CNCLI} leaderlog ${cncliParams} --consensus "${consensus}" --epoch="${epoch}" ${stake_param_next} |
```
Scitz0, line 835: Same as other comment, either set `local epoch="$1"` at the top and use it everywhere instead of `${1}` or skip setting local variable and use `${1}`.

Scitz0 (changes requested):

Scitz0:
> Hello @Scitz0
> 
> I just pushed last changes. I removed --ledger-set from the three leaderlog functions, added --epoch param in runCurrentEpoch and runNextEpoch accordingly to the input var $1
> 
> local epoch="$1" in now in the three leaderlog funcs and the calls from processAllEpochs and processSingleEpoch, have been updated as:
> 
> > runCurrentEpoch ${epoch}
> > runNextEpoch ${epoch}
> > runPreviousEpochs ${epoch}
> 
> Cheers, Manuel

Ok great, though added some more comments as its currently a bit inconsistent with usage of local `${epoch}` variable and `${1}`. So either go `${1}` everywhere or move local variable to the top and use everywhere instead of `$1`

scripts/cnode-helper-scripts/cncli.sh
asnakep: I chose to widely use ${1} - (with no local epoch="$1"). It looks cleaner and less verbose and the reference to epoch number is clear by the context.

I have updated processAllEpochs as

>processAllEpochs() {
  getCurrNextEpoch
  IFS=' ' read -r -a epochs_array <<< "$EPOCHS"

  >for epoch in "${epochs_array[@]}"; do
    **set -- "$epoch"**
    if ! getConsensus "${1}"; then echo "ERROR: Failed to fetch protocol parameters for epoch ${1}."; return 1; fi
    if [[ "$1" == "$curr_epoch" ]]; then
      runCurrentEpoch ${1}
    elif [[ "$1" == "$next_epoch" ]]; then
      runNextEpoch ${1}
    else
      runPreviousEpochs ${1}
    fi
  done

asnakep:
Hello @Scitz0 

last updates have been pushed

cheers,
Manuel

scripts/cnode-helper-scripts/cncli.sh
Scitz0: ok great, but a bit too aggressively :)
In processAllEpochs you need to use epoch as this is the variable used in the for loop. There is no input argument to this function.

asnakep: lets fix it, first time on this last change i did it as you said, then i got the ${1} fever, lol

asnakep: just fixed and pushed

>updated processAllEpochs function to use epoch var from epoch list loop, instead of input "$1"

Scitz0:
I now marked all as resolved and pushed a small fix to solve merge conflict. 

The one thing I want to fix before we merge is the wait time at epoch start. You dont have to do anything about that right now, just leave PR as is and I'll think about it for a bit.

asnakep:
Hello @Scitz0 

Thanks so much for your precious advices on how to proceed with this work. This is the first time I have worked with CNTools scripts, and I hope I haven’t bothered you too much with all the reviews.

Cheers,
Manuel

Scitz0:
> Hello @Scitz0 
> 
> Thanks so much for your precious advices on how to proceed with this work. This is the first time I have worked with CNTools scripts, and I hope I haven’t bothered you too much with all the reviews.
> 
> Cheers,
> Manuel
> 

You are welcome and not at all, always glad to see when people take the time try to improve and add features to the scripts.

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -127,11 +130,15 @@ getLedgerData() { # getNodeMetrics expected to have been already run
...
+  if isNumber "$1"; then
+     getProtocolParamsHist "$(( $1 - 1 ))" || return 1
+  else
+     getProtocolParams || return 1
+  fi
   if versionCheck "9.0" "${PROT_VERSION}"; then
     consensus="cpraos"
     stability_window_factor=3
-  elif versionCheck "8.0" "${PROT_VERSION}"; then
+  elif versionCheck "7.0" "${PROT_VERSION}"; then
```
rdlrt, line 141: was this intentional?

scripts/cnode-helper-scripts/cncli.sh
rdlrt: A note around this, I would definitely not suggest adding any hard-coded entries shen it comes to slots.
The reason when it currently requires waiting from Koios endpoints pov when transitioning across into new epoch is that dbsync 13.2 only completed populating epoch stake for current epoch few hours into new one (wasnt as high as 6 hours, but was certainly a bit).
With dbsync 13.3 (that will be used as part of Koios 1.2.0), this data is populated about 50-70% prior to completion of epoch instead of post, thus - this should be a non-issue post the upgrade.
Given that current behaviour is not changing, I'd prefer not to add such limits for now, we'd be trying to push for koios 1.2.0 sooner (in coming weeks)

scripts/cnode-helper-scripts/cncli.sh
asnakep: Hello @rdlrt 

the condition I found in getConsensus function of cncli.sh when I started the work was:

>getConsensus() {
  getProtocolParams
  if versionCheck "10.0" "${PROT_VERSION}"; then
    consensus="cpraos"
    stability_window_factor=3
  elif versionCheck "7.0" "${PROT_VERSION}"; then
    consensus="praos"
    stability_window_factor=2
  else
    consensus="tpraos"
    stability_window_factor=2
  fi
}

I've just completed the upgrade to node 9.1 using latest guild-deploy / CNTools 

and condition now is 

>getConsensus() {
  getProtocolParams
  if versionCheck "9.0" "${PROT_VERSION}"; then
    consensus="cpraos"
    stability_window_factor=3
  elif versionCheck "8.0" "${PROT_VERSION}"; then
    consensus="praos"
    stability_window_factor=2
  else
    consensus="tpraos"
    stability_window_factor=2
  fi
}

I also just checked in the backup files in archive dir from the upgrade.
And noticed since days that in alpha was as the latest consensus condition.

Cheers,
Manuel

Scitz0: Yeah this error snuck in there for a short time by mistake. Just bad luck that you managed to base your code with that error.  

8.0 I'm fairly certain is the correct major version for this.

asnakep: Hello @Scitz0 and @rdlrt 

I updated getConsensus as

>getConsensus() {
  if isNumber "$1" ; then
     getProtocolParamsHist "$(( $1 - 1 ))" || return 1
  else
     getProtocolParams || return 1
  fi

  >if versionCheck "9.0" "${PROT_VERSION}"; then
    consensus="cpraos"
    stability_window_factor=3
  elif versionCheck "8.0" "${PROT_VERSION}"; then
    consensus="praos"
    stability_window_factor=2
  else
    consensus="tpraos"
    stability_window_factor=2
  fi
}

and started to test the epochs I have with set -x, in this way consensus praos is used from epoch 395, I had a schedule on 394, last epoch using tpraos. The change from 7.0 to 8.0 makes use tpraos for 30 epochs more.

I remember that praos consensus starts in epoch 364, I can be wrong, I'll run leaderlog in both ways for some previous epoch to check for different calculation.

asnakep: Hello @Scitz0

I tested about consensus and protocol version:

this returns correct slots calculation

elif versionCheck "7.0" "${PROT_VERSION}"; then
consensus="praos"

>epoch,epoch_nonce,sigma,epoch_slots_ideal,max_performance
380,3950677ef48bb212ad335f253ad1998be4e9fe4ae2a26d402607850461398125,0.0000015520185588821839,0.03,3333.33
387,6d49fffa700bb7d3399e2dcad064feb9d881900e80e3148471bd5029382ac4da,0.0000016038306633996149,0.03,6666.67

using 8.0 calculation is wrong

elif versionCheck "8.0" "${PROT_VERSION}"; then
consensus="praos"

>epoch,epoch_nonce,sigma,epoch_slots_ideal,max_performance
>380,3950677ef48bb212ad335f253ad1998be4e9fe4ae2a26d402607850461398125,0.0000015520185588821839,0.03,0.0
387,6d49fffa700bb7d3399e2dcad064feb9d881900e80e3148471bd5029382ac4da,0.0000016038306633996149,0.03,0.0

a manual check at epoch/version change

>curl -s "https://api.koios.rest/api/v1/epoch_params?_epoch_no=364" | jq -r '.[].protocol_major,.[].protocol_minor'
6
0

>curl -s "https://api.koios.rest/api/v1/epoch_params?_epoch_no=365" | jq -r '.[].protocol_major,.[].protocol_minor'
7
0

>curl -s "https://api.koios.rest/api/v1/epoch_params?_epoch_no=393" | jq -r '.[].protocol_major,.[].protocol_minor'
7
0

>curl -s "https://api.koios.rest/api/v1/epoch_params?_epoch_no=394" | jq -r '.[].protocol_major,.[].protocol_minor'
8
0

cheers
Manuel

Scitz0: Ok thanks for the tests. Then let's switch to 7.0 👍🏻

asnakep: Hello @Scitz0 

thanks for confirming

cheers,
Manuel

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -45,10 +45,13 @@ usage() {
 		  force     Manually force leaderlog calculation and overwrite even if already done, exits after leaderlog is calculated
 		validate    Continously monitor and confirm that the blocks made actually was accepted and adopted by chain (deployed as service)
 		  all       One-time re-validation of all blocks in blocklog db
-		  epoch     One-time re-validation of blocks in blocklog db for the specified epoch 
+		  epoch     One-time re-validation of blocks in blocklog db for the specified epoch
+		epochdata   Manually re-calculate leaderlog to load stakepool history into epochdata table of blocklog db. Needs completion of cncli.sh sync and cncli.sh validate processes.
+		  all       One-time re-calculation of all epochs (avg execution duration: 1hr / 50 epochs)
+		  epoch     One-time re-calculation for the specified epoch
```
TrevorBenson, line 51: I know I'm late to the review party but have a few comments. Please don't take this as a request for any changes, it's only intended to open a discussion between reviewers/maintainers.

I also have some results from testing I will post separately.
----
The **epochdata**  command seems very tightly related to the **leaderlog** command.  It seems like it could be easy to overlook for the average user while they review **leaderlog** options. 

From solely a user perspective (not considering the current code or complexity to change things) it seems like  **all** and **epoch** subcommands would make more sense being an extension of the **leaderlog** command since it appears to be specific to only adjusting the table where leaderlog may not have been calculated, or was calculated incorrectly.

asnakep: The initial idea of mine was to have an option to load full pool history once, epochdata table will then get informed at every leaderlog check, i don't disagree the idea of make it part of leaderlog function, personally I would not run it as service for first automatic load after the cncli db full sync and of historical load of blocklog table, since it could take several hours to load all the information, 2hr/140epochs approx.

For sure it would be easier for users, epochdata gets loaded by default, then you have the option to do a manual reload, or a single reload (quite useless feature imo, but may be can be useful in some cases).

On last point of single epoch choice I'm going to add: print available epochs from blocklog, otherwise some user could input an epoch for which there were no assigned slots, and get message: slots not found for epoch...

asnakep: it could be a service that run only once and dependent from cncli-sync and cncli-validate or cncli-leaderlog, the one that fill blocklog table.

The only question I have on this, it is fine to run automatic process that could take hours on a BP that has several blocks every epochs? The process runs "cncli leaderlog" at every iteration, I noticed from htop that is quite loady, but may be not to the point to affect negatively the BP node.

asnakep: Hello Team, I added "show all available epochs" on wrong epoch number input. So instead of print epochs at every "cncli.sh epochdata num" that looks ugly to me, is printed on wrong (epoch not in blocklog) input and should be fine to get the list once (or at every wrong input) to know pool's epochs.

image

TrevorBenson: > The initial idea of mine was to have an option to load full pool history once, epochdata table will then get informed at every leaderlog check, i don't disagree the idea of make it part of leaderlog function, personally I would not run it as service for first automatic load after the cncli db full sync and of historical load of blocklog table, since it could take several hours to load all the information, 2hr/140epochs approx.

When first contemplating the feature, I wasn't considering the **init**  command. It makes sense to not combine with the **leaderlog** command given **init** exists and the requirement of the **validate** command has been run.

rdlrt: Just to double-check - Is this conversation OK to resolve?

rdlrt:
The PR will not be merged until koios-1.2.0a is on mainnet (since we'd be able to query stats much earlier with upgraded dbysnc) - removing the wait time post epoch transition

scripts/cnode-helper-scripts/cncli.sh
asnakep: Hello @rdlrt 

I think yes

rdlrt:
With release of koios-1.2.0, this PR should be unblocked - need to ensure that stake snapshot stats are available immediately post epoch transition, and remove 6 hour hard coded timeout (replace it perhaps with check tip and ensure epoch has new block post transition)

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -794,10 +801,173 @@ cncliPTsendslots() {
...
+  getKoiosData
+  echo "Processing current epoch: ${1}"
+  stake_param_curr="--active-stake ${active_stake_set} --pool-stake ${pool_stake_set}"
+
+  ${CNCLI} leaderlog ${cncliParams} --consensus "${consensus}" --epoch="${1}" ${stake_param_curr} |
+  jq -r '[.epoch, .epochNonce, .poolId, .sigma, .d, .epochSlotsIdeal, .maxPerformance, .activeStake, .totalActiveStake] | @csv' |
+  sed 's/"//g' >> "$tmpcsv"
+}
+
+runNextEpoch() {
```
Scitz0, line 824: This function should check that we have crossed into stability window like the normal leaderlog function. Else calculating values for next epoch wont be possible, or add invalid values.

scripts/cnode-helper-scripts/cncli.sh
```diff
@@ -794,10 +801,173 @@ cncliPTsendslots() {
...
+  IFS=' ' read -r -a epochs_array <<< "$EPOCHS"
+
+  for epoch in "${epochs_array[@]}"; do
+    if ! getConsensus "${epoch}"; then echo "ERROR: Failed to fetch protocol parameters for epoch ${epoch}."; return 1; fi
+    if [[ "$epoch" == "$curr_epoch" ]]; then
+      runCurrentEpoch ${epoch}
+    elif [[ "$epoch" == "$next_epoch" ]]; then
+      runNextEpoch ${epoch}
+    else
+      runPreviousEpochs ${epoch}
```
Scitz0, line 870: For previous epoch, ie current - 1, a function similar to runCurrentEpoch should be added that use `go` values instead of set from getKoiosData. Koios doesn't have data for previous epoch in pool_history endpoint. This might be a bug, looking into it. Nonetheless, this change makes sense. runPreviousEpochs can be used for all older epochs. 

Needs to be added in processSingleEpoch as well.

Scitz0 (changes requested):
Found a few more things while testing this PR.

scripts/cnode-helper-scripts/cncli.sh
asnakep: Thank you, I'll review it during the day

scripts/cnode-helper-scripts/cncli.sh
asnakep: Hello @Scitz0 
I added stability window check as requested, to be pushed yet.
Just as question, couldn't the condition simply be: "if blocklog table has no entries for next epoch, then don't run runNextEpoch else run it?

Scitz0: I guess this subcommand is mostly for backfill and that normal leaderlog command handle current/next. So in my view, maybe current and next shouldn't be touched at all?

scripts/cnode-helper-scripts/cncli.sh
asnakep: Hello @Scitz0,

about go field I noticed that it was removed after new Koios release. I just queried pool_history and it returns data for previous epochs.

I'll wait for you input on this point. 

Meanwhile I push the changes on runNextEpoch (added check on stability window).

Cheers,
Manuel

Scitz0: pool_stake_snapshot endpoint is affected by the same bug as pool_history. That pool_stats table in dbsync is a bit messed up in current release. 

pool_stake_snapshot should return go snapshot still, if there are data to return. I will think about this for a moment before we finally merge PR.

Scitz0: I will just go ahead and merge this as is now. We can possibly switch to go snapshot later.

Scitz0 (approved):

PR #1800, opened 2024-08-06
[mithril-client.sh] Use the DB_DIR instead of CNODE_HOME/db

TrevorBenson:
Updates mithril.library to use env DB_DIR value instead of CNODE_HOME/db when creating the mithril.env, and mithril-client.sh to use the DB_DIRECTORY value from the mithril.env file per upstream mithril documentation.

### DIFF
--- a/scripts/cnode-helper-scripts/mithril-client.sh
+++ b/scripts/cnode-helper-scripts/mithril-client.sh
@@ -104,7 +104,7 @@ download_snapshot() {
   if [[ "${DOWNLOAD_SNAPSHOT}" == "Y" ]]; then
     echo "INFO: Downloading latest mithril snapshot.."
     trap 'cleanup_db_directory' INT
-    if ! "${MITHRILBIN}" -v --aggregator-endpoint ${AGGREGATOR_ENDPOINT} cardano-db download --download-dir ${CNODE_HOME} --genesis-verification-key ${GENESIS_VERIFICATION_KEY} ${SNAPSHOT_DIGEST} ; then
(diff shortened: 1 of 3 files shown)


scripts/cnode-helper-scripts/mithril-client.sh
```diff
@@ -104,7 +104,7 @@ download_snapshot() {
   if [[ "${DOWNLOAD_SNAPSHOT}" == "Y" ]]; then
     echo "INFO: Downloading latest mithril snapshot.."
     trap 'cleanup_db_directory' INT
-    if ! "${MITHRILBIN}" -v --aggregator-endpoint ${AGGREGATOR_ENDPOINT} cardano-db download --download-dir ${CNODE_HOME} --genesis-verification-key ${GENESIS_VERIFICATION_KEY} ${SNAPSHOT_DIGEST} ; then
+    if ! "${MITHRILBIN}" -v --aggregator-endpoint ${AGGREGATOR_ENDPOINT} cardano-db download --download-dir ${DB_DIRECTORY%/*} --genesis-verification-key ${GENESIS_VERIFICATION_KEY} ${SNAPSHOT_DIGEST} ; then
```
Scitz0, line 107: Safer to use `$(dirname "${DB_DIRECTORY}")` instead of `${DB_DIRECTORY%/*}` if for some reason a trailing slash is added to DB_DIR. Not sure if thats valid syntax but dirname should make sure to treat both equal and get the parent directory path.

TrevorBenson:
```suggestion
    if ! "${MITHRILBIN}" -v --aggregator-endpoint ${AGGREGATOR_ENDPOINT} cardano-db download --download-dir $(dirname ${DB_DIRECTORY}) --genesis-verification-key ${GENESIS_VERIFICATION_KEY} ${SNAPSHOT_DIGEST} ; then
```

Scitz0 (approved):

PR #1808, opened 2024-08-18
getProtocolParams use CNCLI if ENABLE_KOIOS=N

### LINKED ISSUE #1807: KOIOS API can still be queried when ENABLE_KOIS and USE_KOIOS_API are set to N
**Describe the bug**
When disabling KOIOS entirely via **ENABLE_KOIOS=N** and **USE_KOIOS_API=N** `cncli.sh leaderlog` can still end up calling **getProtocolParams** function from the `env`. The function only considers if **KOIOS_API** is defined, but not if **ENABLE_KOIOS** is **N**.

Strangely enough even though `cntools.sh` seems to call **updateProtocolParams** function during its **main** function loop, and the `cntools.library` defining this function appears to use **getProtocolParams** I cannot seem to capture a call to koios when using cntools.sh.

**To Reproduce**
1. run `export ENABLE_KOIS=N`
2. run `export USE_KOIOS_API=N`
3. Run `cncli.sh leaderlog`
4. Either
   a. Use `bash -x` and examine the trace to find calls to koios.rest endpoint route of cli_protocl_params
   b. Use tcpdump and capture the network traffic showing the call occurs.
6. See either:
   a. curl command: `curl -sSL -f -X GET https://guild.koios.rest/api/v1/cli_protocol_params`
   b. tcp traffic to the IP of testnet (or mainnet) koios`08:25:11.767094 IP X.X.X.X.54316 > 95.216.173.194.https`

**Expected behavior**
Setting **ENABLE_KOIOS=N** along with **USE_KOIOS_API=N** should not query KOIOS for anything and all functions should rely on local methods instead of querying the KOIOS API.

**Version:**
 - OS: Debian GNU/Linux 12 (bookworm) version (cardano-node from guild-operators containers)
 - Product version: CNTools 13.1.0 (don't see a unique version for cncli.sh)
 - Cardano Node version: N/A
 - Network you're connecting to: 2021-12-09T22:55:22Z

**Additional context**
It's a combo in `env` of when NWMAGIC is set, then KOIOS_API will always be defined, and in getProtocolParams function if KOIOS_API is defined it will get used instead of cardano-cli, even when both ENABLE_KOIOS=N and USE_KOIOS_API=N.

It's not impacting the total amount of KOIOS queries I was sending previously when trying to track down HTTP 429 rate limit exceeded issues. I just stumbled on it after disabling KOIOS while trying to verify I didn't have **any** access of KOIOS API's happening other than when I enable leaderlog to use KOIOS to accurately count the total qty of queries being sent from a single IPv4 public address shared by my testnet pools.

TrevorBenson:
## Description
Disables KOIOS access from getProtocolParams if ENABLE_KOIOS is set to N.

## Motivation and context
Found unexpected KOIOS API access when trying to completely disable KOIOS. I had set both ENABLE_KOIOS and USE_KOIOS_API to N, but still found cases where traffic was hitting KOIOS endpoints.

## Which issue it fixes?
Closes #1807

## How has this been tested?
From inside a container using bash -x and capturing stderr to stdout and confirming the curl to cli_protocol_params API route no longer occurs when cncli.sh leaderlog is executed. Also using tcpdump to just examine traffic destined for the IP address of guild.koios.rest.

### DIFF
--- a/scripts/cnode-helper-scripts/cncli.sh
+++ b/scripts/cnode-helper-scripts/cncli.sh
@@ -210,7 +210,9 @@ cncliInit() {
     echo "sleeping for 10s and testing again..."
     sleep 10
   done
-  
+
+  test_koios
+
   TMP_DIR="${TMP_DIR}/cncli"
   if ! mkdir -p "${TMP_DIR}" 2>/dev/null; then echo "ERROR: Failed to create directory for temporary files: ${TMP_DIR}"; exit 1; fi
   
--- a/scripts/cnode-helper-scripts/cncli.sh
+++ b/scripts/cnode-helper-scripts/cncli.sh
@@ -25,7 +25,6 @@
 #CONFIRM_SLOT_CNT=600                     # CNCLI validate: require at least these many slots to have passed before validating
 #CONFIRM_BLOCK_CNT=15                     # CNCLI validate: require at least these many blocks on top of minted before validating
 #BATCH_AUTO_UPDATE=N                      # Set to Y to automatically update the script if a new version is available without user interaction
-#USE_KOIOS_API=Y                          # Use Koios API in cncli leaderlog instead of local stake-snapshot query to reduce system resources. (default true)
 #CNCLI_PROM_PORT=12799                    # Set Prometheus port for cncli block metrics available through metrics operation (default: 12799)
 
 ######################################
@@ -141,7 +140,6 @@ getConsensus() {
 }
 
 getKoiosData() {
-  [[ -z ${KOIOS_API} ]] && return 1
   if ! stake_snapshot=$(curl -sSL -f -d _pool_bech32=${POOL_ID_BECH32} "${KOIOS_API}/pool_stake_snapshot" 2>&1); then
     echo "ERROR: Koios pool_stake_snapshot query failed: curl -sSL -f -d _pool_bech32=${POOL_ID_BECH32} ${KOIOS_API}/pool_stake_snapshot"
     return 1
@@ -223,7 +221,6 @@ cncliInit() {
   [[ -z "${CNCLI_DIR}" ]] && CNCLI_DIR="${CNODE_HOME}/guild-db/cncli"
   if ! mkdir -p "${CNCLI_DIR}" 2>/dev/null; then echo "ERROR: Failed to create CNCLI DB directory: ${CNCLI_DIR}"; exit 1; fi
   CNCLI_DB="${CNCLI_DIR}/cncli.db"
-  [[ -z "${USE_KOIOS_API}" ]] && USE_KOIOS_API=Y
   [[ -z "${CNODE_HOST}" ]] && CNODE_HOST="127.0.0.1"
   [[ -z "${SLEEP_RATE}" ]] && SLEEP_RATE=60
   [[ -z "${CONFIRM_SLOT_CNT}" ]] && CONFIRM_SLOT_CNT=600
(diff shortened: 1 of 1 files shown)


TrevorBenson:
@rdlrt @Scitz0
Should this PR include conditional logic to also check if **USE_KOIOS_API** is set to N and also fallback to cardano-cli for protocol params?

* `cncli.sh` ignores **ENABLE_KOIOS** in favor of checking **USE_KOIOS_API** environment variables.
* `cntools.sh does not consider either environment variable.
  * I could not get `cntools.sh` to cause this behavior of calling koios API's via **updateProtocolParams** function after a couple of tries of jumping between menus in cntools, even though it appeared to be part of the main loop.

scripts/cnode-helper-scripts/env
```diff
@@ -817,7 +817,7 @@ slotInterval() {
 # 2 : CLI - node response not valid json
 # 3 : Koios - general error
 getProtocolParams() {
-  if [[ -n ${KOIOS_API} ]]; then
+  if [[ -n ${KOIOS_API} ]] && [[ ${ENABLE_KOIOS} != N ]]; then
```
rdlrt, line 820: For this one, it is preferable to use test_koios in cncli.sh like CNTools does [here](https://github.com/cardano-community/guild-operators/blob/1bc41b7487bcddc7c4e1e7dd268d9c0b63bbb205/scripts/cnode-helper-scripts/cntools.sh#L141) , so that it will test connectivity alongwith just variable definition

rdlrt (changes requested):

scripts/cnode-helper-scripts/env
TrevorBenson: test_koios only does this when **ENABLE_KOIOS = Y**, giving the SPO the control to disable KOIOS:
```
test_koios() {
  # make sure KOIOS_API is reachable, else fall back to cli
  [[ ${ENABLE_KOIOS} = 'Y' && -n ${KOIOS_API} && $(curl -sfk -o /dev/null -w "%{http_code}" -m 5 ${KOIOS_API}/tip | awk '{print $1}') = "200" ]] || unset KOIOS_API
}
```
I don't recall why I didn't make it an identical check of `= Y` instead of `!= N`. I remember I had something in mind when I did it, but maybe it was just an edge case of SPO typo/using `YES` if re-enabling it KOIOS without commenting out their setting.

I'm good with either check but prefer a consistent behavior when the SPO explicitly disables KOIOS via **ENABLE_KOIOS=N**. This way tools honor the SPO configuration, and no traffic would be observed via tcpdump or result in alerts for EDR systems.

TrevorBenson: I may have misunderstood your statement as "don't do this". Did you actually mean:
```
if [[ -n ${KOIOS_API} ]] && [[ ! test_koios ]]; then
```

Scitz0: No, as part of cncli.sh cncliInit function, we should call test_koios function just as CNTools does. This checks both ENABLE_KOIOS and that koios api is actually reachable and network supported. Else it will unset KOIOS_API variable. 

So if this is test function is called, `-n ${KOIOS_API}` should be enough to use everywhere after this. 

I can also see that we have USE_KOIOS_API in cncli.sh. We should remove this and replace with same logic.

TrevorBenson: Gotcha, I'll modify the PR to do it during the init.

Scitz0: Preferred solution:

1. revert env change
2. global search on `USE_KOIOS_API` in cncli.sh and remove where needed and replace with `-n ${KOIOS_API}` where check is done.
3. add a test_koios call at the appropriate place in `cncliInit` function, ie after loading/updating env but before needing to use KOIOS_API variable.

Scitz0 (approved):
👍

rdlrt (approved):

PR #1842, opened 2025-01-07
Include healthcheck logic for helper scripts running as sidecars

### LINKED ISSUE #1841: Docker container healthcheck for CNCLI usage
Hello,

I wonder if it makes sense to request an improvement to the builtin healthcheck of the Docker container? ... 
When enabling CNCLI services by configuring it to run as a docker sidecar container, the sidecar container's built-in healthcheck fails.

Would it be possible to update healthcheck.sh to detect when the container has been started with 'ENTRYPOINT_PROCESS: cncli.sh', and 'command: sync, validate, etc...' and adjust the heathcheck that is performed?

Below is an example docker compose file to illustrate a sidecar config... 
(The CNCLI sidecar container's entrypoint is overridden by setting env variable: ```ENTRYPOINT_PROCESS: cncli.sh```  and ```command: sync```)

```
services:

#  built-in healthcheck works:
  cardano-node:
    image: cardanocommunity/cardano-node:latest
    volumes:
      - cn-sockets:/opt/cardano/cnode/sockets
      - cn-logs:/opt/cardano/cnode/logs
      - cn-guild-db:/opt/cardano/cnode/guild-db

#  built-in healthcheck fails:
  cardano-cncli-sync:
    image: cardanocommunity/cardano-node:latest
    network_mode: service:cardano-node
    volumes:
      - cn-sockets:/opt/cardano/cnode/sockets
      - cn-logs:/opt/cardano/cnode/logs
      - cn-guild-db:/opt/cardano/cnode/guild-db
    environment:
      ENTRYPOINT_PROCESS: cncli.sh                            # Override entrypoint
    command: sync                                             # Entrypoint command
```

Thankyou!
-Adam

TrevorBenson:
## Description
Enhances the healthcheck.sh script to work for checking permissions on sidecar containers (helper scripts) via the ENTRYPOINT_PROCESS. 

1. Copy the updated healthcheck.sh script into the sidecar container in `/home/guild/.scripts/`.
2. Wait for the 5 minute check interval to occur and confirm if the sidecar is now shown as healthy.

### Testing different CPU Usage values
* Define a CPU_THRESHOLD environment variable (defaults to `80` %) at a value you want to mark a container unhealthy when it is exceeded.

### Testing different amount of retries (internal to healthcheck.sh script).
* Define a RETRIES environment variable (defaults to `20`) at a number of retries you want to perform if the CPU usage is above the CPU_THRESHOLD value before exiting non zero

Currently it is a 3 second delay between checks, so 20 retries results in up to 60 seconds before the healthcheck will exit as unhealthy due to CPU load.

### Testing different healthcheck values (external to healthcheck.sh script).
The current HEALTHCHECK of the container image is:
* 5 minutes start period
* 5 minutes interval
* 100 seconds timeout
* 3 retries (default value from being undefined)

Reducing the start period and intervals to something more appropriate for the sidecar script will result in a much shorter period to determine the sidecar containers health.

**Make sure to keep the environment variable RETRIES * 3 < container healthcheck timeout to avoid marking the container unhealthy before the script will return during periods of high cpu load.**

## Motivation and context
Issue #1841 

## Which issue it fixes?
Closes #1841

## How has this been tested?
1. `docker cp` the script into preview network cncli sync, validate and leaderlog containers and waiting until the interval runs the script
2. Execute the script with `docker exec` to confirm it reports healthy
3. Monitor the containers until the healthcheck interval occurs and that they are marked healthy.

## Additional Details

There is a SLEEPING_SCRIPTS array which is used for validate and leaderlog to still check for the cncli binary, but not consider a sleep period for validate and leaderlog to be unhealthy. Not 100% sure this is the best approach, but with sleep periods being variable I felt it was likely an acceptable middle ground.

Please do not hesitate to suggest an alternative approach to handling sleeping sidecars healthchecks if you think you have an improvement.

@adamsthws if you could please copy this into your sidecar containers (and your pool) and report back any results. I am marking this as a draft PR for the time being until testing is completed, after which if things look good I will mark it for review and get feedback from others.

Thanks


TrevorBenson:
FWIW Here is my preview network pool, and cncli containers showing healthy once the script was copied in and healthcheck interval was reached:
```
# podman ps --filter 'name=preview-cncli-[slv]' --filter 'name=preview-ccio-pool --format '{{ .Names }}\t{{ .Status }}''
preview-ccio-pool	Up 4 weeks (healthy)
preview-cncli-sync	Up 3 days (healthy)
preview-cncli-validate	Up 3 days (healthy)
preview-cncli-leaderlog	Up 3 days (healthy)
```

adamsthws:
Looks good!
How I have tested...

    cp the script into cncli sync, validate, leaderlog, pt-send-slots, pt-send-tip containers 
    
    Execute the script with docker exec. 
      - Result: exit 0. 'We're healthy - cncli'
    
    Monitor the containers until the healthcheck interval occurs and that they are marked healthy.
      - Result: Docker ps shows containers are healthy.
    
    Adjusted RETRIES... 
       Note - Setting RETRIES=0 results in exit 127, 'Max retries reached for cncli'. 
      
    Adjusted CPU_THRESHOLD...
       Note - The cncli process uses very little cpu so even setting as low as 1% i was unable to intentionally get healthcheck to fail.
    
    
```
# sudo docker ps --filter 'name=cardano' --format '{{ .Names }}\t{{ .Status }}' | column -t
cardano-mainnet-bp_cncli-pt-send-tip    Up  33  minutes  (healthy)
cardano-mainnet-bp_cncli-pt-send-slots  Up  33  minutes  (healthy)
cardano-mainnet-bp_cncli-sync           Up  33  minutes  (healthy)
cardano-mainnet-bp_cncli-validate       Up  33  minutes  (healthy)
cardano-mainnet-bp_cncli-leaderlog      Up  33  minutes  (healthy)
cardano-mainnet-bp                      Up  33  minutes  (healthy)
cardano-mainnet-relay0                  Up  33  minutes  (healthy)
```

adamsthws:
Further testing...

After deleting the cncli db and re-syncing I was able to test it with higher cpu load. 

Result
```
# ./healthcheck.sh
Checking health for process: cncli 
./healthcheck.sh: line 44: ((: 67.9: syntax error: invalid arithmetic operator (error token is ".9")
We're healthy - cncli
# exit 0
```

TrevorBenson:
> Looks good! How I have tested...
> 
> cp the script into cncli sync, validate, leaderlog, pt-send-slots, pt-send-tip containers
> 
> Execute the script with docker exec.
> 
>     * Result: exit 0. 'We're healthy - cncli'
> 
> 
> Monitor the containers until the healthcheck interval occurs and that they are marked healthy.
> 
>     * Result: Docker ps shows containers are healthy.
> 
> 
> Adjusted RETRIES
> 
>     * Note
>       
>       * Setting RETRIES=0 results in exit 127, 'Max retries reached for cncli'.
>       * Setting RETRIES=1 results in exit 0. 'We're healthy - cncli'
> 
> 
> Adjusted CPU_THRESHOLD
> 
>     * Note
>       
>       * The cncli process uses very little cpu so even setting threshold as low as 1% i was unable to intentionally get healthcheck to fail.
> 
> 
> ```
> # docker ps --filter 'name=cncli' --format '{{ .Names }}\t{{ .Status }}' | column -t
> cncli-pt-send-tip    Up  33  minutes  (healthy)
> cncli-pt-send-slots  Up  33  minutes  (healthy)
> cncli-sync           Up  33  minutes  (healthy)
> cncli-validate       Up  33  minutes  (healthy)
> cncli-leaderlog      Up  33  minutes  (healthy)
> ```

Yeah, there are rare instances where cncli percentage can be high, but this tends to be when there are issues with the node process itself, like if it gets stuck chainsync/blockfetch and never completes, I have seen cncli get a high percentage. Otherwise its quite rare to see it increase, but I figured with mithril-signer or db-sync, it might be more useful.

files/docker/node/addons/healthcheck.sh
```diff
@@ -1,43 +1,132 @@
...
-        echo "there is a problem"
-        exit 1
+# Function to check if a process is running and its CPU usage
+check_process() {
+    local process_name="$1"
+    local cpu_threshold="$2"
+
+    for (( CHECK=1; CHECK<=RETRIES; CHECK++ )); do
+        # Check CPU usage of the process
+        CPU_USAGE=$(ps -C "$process_name" -o %cpu= | awk '{s+=$1} END {print s}')
```
TrevorBenson, line 41:
```suggestion
        CPU_USAGE=$(ps -C "$process_name" -o %cpu= | awk '{s+=$1} END {print int(s + 0.5)}')
```

TrevorBenson:
@adamsthws Feel free to submit suggestions to adjust the `SCRIPT_TO_BINARY_MAP` entries. Otherwise this week when I look at some other issues I'll go through each helper script and update the map and set this PR to ready to review.

Thanks for the testing.

adamsthws:
I previously mentioned that setting RETRIES=0 results in script exit 1 without running the loop... it would be preferable to run the loop once when RETRIES=0.

Suggestion - Modify the loop condition to handle RETRIES=0 - Changing line 39 to the following:
```
    for (( CHECK=1; CHECK<=RETRIES || (RETRIES==0 && CHECK==1); CHECK++ )); do
```

adamsthws:
I started thinkinng about a cncli specific check. The following function is an idea to check cncli status...

```
# Function to check cncli status
check_cncli_status() {
    CNCLI=$(which cncli)

    for (( CHECK=1; CHECK<=RETRIES || (RETRIES==0 && CHECK==1); CHECK++ )); do
        CNCLI_OUTPUT=$($CNCLI status \
            --byron-genesis "/opt/cardano/cnode/files/byron-genesis.json" \
            --shelley-genesis "/opt/cardano/cnode/files/shelley-genesis.json" \
            --db "/opt/cardano/cnode/guild-db/cncli/cncli.db")

        CNCLI_STATUS=$(echo "$CNCLI_OUTPUT" | jq -r '.status')
        ERROR_MESSAGE=$(echo "$CNCLI_OUTPUT" | jq -r '.errorMessage')

        if [ "$CNCLI_STATUS" == "ok" ]; then
            echo "We're healthy - cncli status is ok and synced."
            return 0  # Return 0 if the status is ok
        elif [ "$CNCLI_STATUS" == "error" ]; then
            if [ "$ERROR_MESSAGE" == "db not fully synced!" ]; then
                echo "cncli's sqlite database is not fully synced. Attempt $CHECK. Retrying in 3 minutes."
                sleep 180  # Wait 3 minutes then retry to allow time for the database to sync
            elif [ "$ERROR_MESSAGE" == "database not found!" ]; then
                echo "cncli's sqlite database not found. Attempt $CHECK. Retrying in 3 minutes."
                sleep 180  # Wait 3 minutes then retry to allow time for the database to be created
            else
                echo "Error - cncli status: $ERROR_MESSAGE. Attempt $CHECK. Retrying in 3 seconds."
                sleep 3  # Wait 3 seconds then retry for other errors
            fi
        else # If status is not "ok" or "error"
            echo "cncli status: $CNCLI_STATUS. Attempt $CHECK. Retrying in 3 seconds."
            sleep 3  # Wait 3 seconds then retry
        fi
    done

    echo "cncli status check failed after $RETRIES attempts."
    return 1  # Return 1 if retries are exhausted
}

```

Perhaps would be improved further by also checking if sync is incrementing, so the healthcheck doesn't fail during initial sync. 

How would you feel about adding me as a commit co-author if you decide to use this?

TrevorBenson:
@adamsthws I'm happy to make you a co-author even for something simple, fir example if you know how to submit a suggestion go ahead an apply one for `for (( CHECK=0; CHECK<=RETRIES; CHECK++ )); do` to modify the PR and I'll merge it as a commit.

TrevorBenson:
In regards to the larger block for cncli checks, first it clear lots of thought went into it. 

This portion:
```
            if [ "$ERROR_MESSAGE" == "db not fully synced!" ]; then
                echo "cncli's sqlite database is not fully synced. Attempt $CHECK. Retrying in 3 minutes."
                sleep 180  # Wait 3 minutes then retry to allow time for the database to sync
            elif [ "$ERROR_MESSAGE" == "database not found!" ]; then
                echo "cncli's sqlite database not found. Attempt $CHECK. Retrying in 3 minutes."
                sleep 180  # Wait 3 minutes then retry to allow time for the database to be created
```
Sleeps of 180 exceed the current timeout period of 100. Options:
* Additional documentation. My gut feeling it would lead to additional support requests for operators who don't read the docs, it also makes the monolithic container slightly more complex.
* Increasing the the containers timeout. Potential to reduce observability for node, or other processes.

With container settings of 3 retries and 5 minute interval w/ 100 second timeout it is 15 minutes from the last healthy response, or 10 minutes from the first unhealthy response, before the container exhausts retries and is marked unhealthy. I think this covers the two 180 second sleeps, even if the operator reduces the interval and timeouts when not running the node.

Separately, conversations outside of this PR and thread have pointed to some of the logic used in KOIOS for db-sync, also that it could also be used for checking the sqlite DB for cncli.

```
#!/bin/bash

export PGPASSWORD=${POSTGRES_PASSWORD}
[[ $(( $(date +%s) - $(date --date="$(psql -U ${POSTGRES_USER} -d ${POSTGRES_DB} -qt -c 'select time from block order by id desc limit 1;')" +%s) )) -lt 3600 ]] || exit 1
```

I haven't examined what the common drift might be for a db-sync instance from the last black produced and for cncli I suspect we could make it shorter than 1 hour.

----

These are just my thoughts. If you think that I overlooked some aspect please don't hesitate to continue the discussion.

adamsthws:
> I haven't examined what the common drift might be for a db-sync instance from the last block produced and for cncli I suspect we could make it shorter than 1 hour.

Thankyou for the feedback.

It seems the limitation of using 'cncli status' in the healthcheck is that it can't be used to determine if sync is progressing... Rather, it only indicates that it is either synced or not-synced. The long timeout allows time for sync to complete but I see that has introduced observability shortcomings. Instead, checking for cncli sync progress via sqlite db would seem to make more sense.

files/docker/node/addons/healthcheck.sh
```diff
@@ -1,43 +1,132 @@
...
-    SECOND=$($CCLI query tip --testnet-magic ${NWMAGIC} | jq .block)
-    if [[ "$FIRST" -ge "$SECOND" ]]; then
-        echo "there is a problem"
-        exit 1
+# Function to check if a process is running and its CPU usage
+check_process() {
+    local process_name="$1"
+    local cpu_threshold="$2"
+
+    for (( CHECK=1; CHECK<=RETRIES; CHECK++ )); do
```
adamsthws, line 39:
```suggestion
    for (( CHECK=0; CHECK<=RETRIES; CHECK++ )); do
```
Suggestion: Modify the loop condition to handle RETRIES=0. 
- With this...
  -  'RETRIES=3' results in the loop running a total of 4 times
  -  'RETRIES=0' results in the loop running a total of 1 times

adamsthws:
Here's a suggestion to check the health of cncli ptsendtip...

```
# Function to check if the tip is successfully being sent to Pooltool
check_cncli_send_tip() {

    for (( CHECK=0; CHECK<=RETRIES; CHECK++ )); do
        # Get the process ID of cncli
        PID=$(pgrep -o cncli)

        # Capture the next output from cncli that is related to Pooltool
        LOG=$(cat /proc/$PID/fd/1 | grep -i --line-buffered "pooltool" | head -n 1)

        # Define the success message to check for
        success_message='.*"success":true.*'
        # Check if the success message exists in the captured log
        if echo "$LOG" | grep -q $success_message; then
            echo "We're healthy - Tip is being sent to Pooltool."
            return 0  # Return 0 if the success message is found
        else # If the success message is not found
            retry_wait=10
            echo "Warning - Attempt $((CHECK + 1)) not successful - Retrying in $retry_wait seconds."
            sleep $retry_wait  # Wait n seconds then retry
        fi
    done
    
    # If retries are exhausted
    echo "Error - Max retries reached - Tip is not being sent to Pooltool."
    return 1  # Exit with error
}
```

> > I haven't examined what the common drift might be for a db-sync instance from the last block produced and for cncli I suspect we could make it shorter than 1 hour.
> 
> Thankyou for the feedback.
> 
> It seems the limitations of using 'cncli status' in the healthcheck are twofold...
> 
> 1 - 'cncli status' can't be used to determine if sync is progressing... Rather, it only indicates that it is either synced or not-synced. The long timeout allows time for sync to complete but I see that has introduced observability shortcomings.
> 
> Instead, checking for cncli sync progress via sqlite db would seem to make more sense.
> 
> 2 - Not all cncli containers need sqlite. (E.g. send-tip). In the cncli send-tip container (with no sqlite db), 'cncli status' always returns error, as it will never reach sync. (The tip is sucessfully being sent to Pooltool without the need of sqlite).
> 
> ```
> {
>   "status": "error",
>   "errorMessage": "database not found!"
> }
> ```
> 
> I think it's reasonable to say 'cncli status' is not a suitable way to check cncli container health.

TrevorBenson:
> I think it's reasonable to say 'cncli status' is not a suitable way to check cncli container health.

Yep, this is where I was originally taking "the easy way out" on the healthcheck of checking if the binary was running or not, and/or sleep.

TrevorBenson:
> Here's a suggestion to check the health of cncli ptsendtip, I'd love your feedback...

#General Thoughts
First, this part is more about the existing healthcheck than your code contributions. History was that node ping test was removed and the query tip check was implemented. Unfortunately to compare takes at the 20 seconds for the next slot to see it progress. In some cases much longer:
image

I added KOIOS later to try and reduce the 60 second sleep in the hopes to improve observability, if were matching KOIOS we don't even look for progression, were in sync. But in case we weren't, then retry and check for being in sync, w/ 3 second retries it should be very few retries usually to match koios. 

As a single unhealthy response requires exhausting the containers `--health-retries` before marking the container unhealthy my preference would be to aim for using "fail early" methodology on any other checks.

Now, this is just my viewpoint on providing the fastest feedback loop and observability. Its **always** open for discussion. My viewpoint isn't "the law" when it comes to how it will work, especially as we get more contributors that actually use the container and its specific tools. Feel free to push back on my views at any time.

----

1. If retry_wait is a variable (which I like) we should probably make it a global/constant so it can be adjusted by the user for any check function that does do retries.
   * While variable name collisions are probably unlikely I'm thinking of prefixing healthcheck variables just in case. Maybe `HEALTHCHECK_`, so like:
     * HEALTHCHECK_CPU_THRESHOLD
     * HEALTHCHECK_RETRIES
     * HEALTHCHECK_RETRY_WAIT
   * This provides the operator full control. For example using `HEALTHCHECK_RETRIES=0` for a single test, and tune everything else via the containers `--health-interval` and `--health-retries`, or vice versa.

2. For `process_id=$(pgrep -o cncli)` do you find that `cncli sendtip` is not running? 
   * Did you set `--health-start-period 0` and observe a race where the healthcheck happened first and failed? I'd prefer removing the retry in the function and just return 1 for unhealthy with a message to fail fast for observability and let the operator choose retries at the container level as often as possible.

3. Given the above example of 80 seconds or more between blocks I suspect that the `timeout 30` may result in not catching the entry at all.

4. Have you tested and confirmed its working as intended?
   * I tried cat on the `cncli sendtip` file descriptor 1 while watching log entries. While I saw logs from the container showing a send operation, I never saw from the file descriptor. Are you seeing something different in your tests?

I can add some logic that determines which subcommand cncli.sh was provided. Something like: `cncli_subcmd=$(ps -p 1 -o cmd= | awk '{print $NF}')` should allow us to separate `sendtip` from things where we would use sqlite.

I've got a commit I'll push tomorrow which includes db-sync and the cncli checks based on the db. Then figure out how to move forward on sendtip.

Thanks for the work on this!

TrevorBenson:
@adamsthws OK commit pushed, sorry for the delay. Spent a little time considering if healthcheck.sh could remain in `/home/guild/.scripts/` and properly source env. There is a placeholder for your check to go in once the function is added. Feel free to add a suggestion for it, or open a PR against the source branch of this PR and I'll merge that, whatever is your preference.

Thanks.

adamsthws:
>2. For `process_id=$(pgrep -o cncli)` do you find that `cncli sendtip` is not running?
>        
>* Did you set `--health-start-period 0` and observe a race where the healthcheck happened first and failed? I'd prefer removing the retry in the function and just return 1 for unhealthy with a message to fail fast for observability and let the operator choose retries at the container level as often as possible.
> 

In my limited testing have never found `cncli sendtip` to not be running, or a race when `--health-start-period 0`... I shall remove the retry.

-----

>3. Given the above example of 80 seconds or more between blocks I suspect that the `timeout 30` may in some attempts result in not catching the entry at all. So this might force us to include some retries, otherwise the `--health-retries` could miss 3 passes and mark it unhealthy even when its working.

If nothing is ever written to the file descriptor...
- With a long timeout, it will fail slowly.
- With a short timeout plus retries, it will also fail slowly 

Do you reccommend...
- Lower timeout, with retries within the function that can echo information to the user on each retry.
- Higher timeout to allow time for the next tip, and remove retries within the function?
- Something else?

>4. Have you tested and confirmed its working as intended?
>        
>I tried cat on the `cncli sendtip` file descriptor 1 while watching log entries. While I saw logs from the container showing a send operation, I never saw from the file descriptor. Are you seeing something different in your tests?

Yes, I have tested... Might the discrepancy between your results and mine be that:
- You're looking at the PID of the entrypoint script: ```pid=$(pgrep -f "${ENTRYPOINT_PROCESS}")```
- Whereas I'm looking at PID of cncli: ```pid=$(pgrep -o cncli)```

> I've got a commit I'll push tomorrow which includes db-sync and the cncli checks based on the db. Then figure out how to move forward on sendtip.

Great! I'll try to review quickly but may be a day or two. Thanks!

files/docker/node/addons/healthcheck.sh
```diff
@@ -1,43 +1,193 @@
...
+            echo "We're healthy - node: $FIRST -> node: $SECOND"
+            return 0
         fi
+    else
+        CURL=$(which curl)
+        JQ=$(which jq)
+        URL="${KOIOS_API}/tip"
+        SECOND=$($CURL -s "${URL}" | $JQ '.[0].block_no')
+
+        for (( CHECK=1; CHECK<=HEALTHCHECK_RETRIES; CHECK++ )); do
```
adamsthws, line 119:
```suggestion
        for (( CHECK=0; CHECK<=HEALTHCHECK_RETRIES; CHECK++ )); do
```

adamsthws (commented):
I haven't had chance to do any testing but here is a quick initial look over. Thanks!

TrevorBenson:
> If nothing is ever written to the file descriptor...
> 
>     * With a long timeout, it will fail slowly.
> 
>     * With a short timeout plus retries, it will also fail slowly
> 
> 
> Do you reccommend...
> 
>     * Lower timeout, with retries within the function that can echo information to the user on each retry.
> 
>     * Higher timeout to allow time for the next tip, and remove retries within the function?
> 
>     * Something else?

I suppose we could align with the original query tip (non KOIOS) and go for 60 seconds. If it fails on the first pass, then there are 3 retries before the container is marked unhealthy anyway. Thoughts?


---

Also I realize in my prior explanation I clearly had the timing off. I should have said 20 total and 15 from the first failure, since it would be defaulting to 3 retries with the 5 minute intervals, not 3 total attempts. At least when operators do not tune the interval, retries, etc. at container creation.

TrevorBenson:
> Yes, I have tested... Might the discrepancy between your results and mine be that:
> 
>     * You're looking at the PID of the entrypoint script: `pid=$(pgrep -f "${ENTRYPOINT_PROCESS}")`

Nope. This was a manual test inside my cncli sync container. So I manually checked the PID and used tmux, not the script at all. Because I wasn't using a valid API key I checked both fd 1 and 2 (and 0).

image

> 
>     * Whereas I'm looking at PID of cncli: `pid=$(pgrep -o cncli)`

I'll chalk it up to some oddity with the manual test and using an XXXX value for the API ID. As long as you are seeing the output and its working in your test that is good enough for me. 


I need to create a new account at PT to replace my old one based on email address anyway. I'll try to retest once I have a valid setup and start using ptsendtip as a container again.

files/docker/node/addons/healthcheck.sh
```diff
@@ -1,43 +1,193 @@
...
+# shellcheck disable=SC2317
+######################################
+# User Variables - Change as desired #
+# Common variables set in env file   #
+######################################
 
-source /opt/cardano/cnode/scripts/env
+ENTRYPOINT_PROCESS="${ENTRYPOINT_PROCESS:-cnode.sh}"            # Get the script from ENTRYPOINT_PROCESS or default to "cnode.sh" if not set
+HEALTHCHECK_CPU_THRESHOLD="${HEALTHCHECK_CPU_THRESHOLD:-80}"    # The CPU threshold to warn about if the sidecar process exceeds this for more than 60 seconds, defaults to 80%.
+HEALTHCHECK_RETRIES="${HEALTHCHECK_RETRIES:-20}"                # The number of retries if tip is not incrementing, or cpu usage is over the threshold
```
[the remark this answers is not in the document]
TrevorBenson: Yep, just waiting for you feedback/thoughts. Thanks!

adamsthws:
> Nope. This was a manual test inside my cncli sync container. So I manually checked the PID and used tmux, not the script at all. Because I wasn't using a valid API key I checked both fd 1 and 2 (and 0).
> 
> I'll chalk it up to some oddity with the manual test and using an XXXX value for the API ID. As long as you are seeing the output and its working in your test that is good enough for me.
> 
> I need to create a new account at PT to replace my old one based on email address anyway. I'll try to retest once I have a valid setup and start using ptsendtip as a container again.

Here's the result I get when I exec into the container to test manually (with an invalid PT API Key):
Screenshot 2025-01-14 173152

adamsthws:
> I suppose we could align with the original query tip (non KOIOS) and go for 60 seconds. If it fails on the first pass, then there are 3 retries before the container is marked unhealthy anyway. Thoughts?

Sounds like a reasonable approach!

files/docker/node/addons/healthcheck.sh
```diff
@@ -1,43 +1,194 @@
...
+            return 1
+        fi
     else
-        echo "we're healthy - node: $FIRST -> node: $SECOND"
+        # No-op. placeholder for check_cncli_ptsendtip
+        :
+        # if ! check_cncli_ptsendtip; then
+        #     return 1
+        # else
+        #     return 0
```
adamsthws, line 50:
```suggestion
        if check_cncli_send_tip; then
            return 0
        else
            return 1
        fi
```

files/docker/node/addons/healthcheck.sh
```diff
@@ -1,43 +1,194 @@
...
+    if check_tip "${NODE_SLOT}" "${CNCLI_SLOT}" "${CNCLI_DB_ALLOWED_DRIFT}"  ; then
+        echo "We're healthy - DB is in sync"
+        return 0
+    else
+        echo "Error: DB is not in sync"
+        return 1
+    fi
+}
+
+
```
adamsthws, line 70:
```suggestion

# Function to check if the tip is successfully being sent to Pooltool
check_cncli_send_tip() {
    # Timeout in seconds for capturing the log entry
    log_entry_timeout=60

    # Get the process ID of cncli
    process_id=$(pgrep -of cncli) || {
        echo "Error: cncli process not found."
        return 1  # Return 1 if the process is not found
    }

    # Loop through the retries
    for (( CHECK=0; CHECK<=HEALTHCHECK_RETRIES; CHECK++ )); do

        # Define the error suffix message for retries
        if [ "$HEALTHCHECK_RETRIES" -ne 0 ]; then
        error_message_suffix="Attempt $((CHECK + 1)). Retrying in $HEALTHCHECK_RETRY_WAIT seconds."
        else error_message_suffix="Retries disabled (HEALTHCHECK_RETRIES=0)"
        fi

        # Capture the next output from cncli that is related to Pooltool
        pt_log_entry=$(timeout $log_entry_timeout cat /proc/$process_id/fd/1 | grep -i --line-buffered "pooltool" | head -n 1)
        if [ -z "$pt_log_entry" ]; then
            echo "Unable to capture cncli output within $log_entry_timeout seconds. $error_message_suffix"
            sleep $HEALTHCHECK_RETRY_WAIT  # Wait n seconds then retry
            continue # Retry if the output capture fails
        fi

        # Define the success message to check for
        success_status='.*"success":true.*'
        failure_status='.*"success":false.*'

        # Check if the success message exists in the captured log
        if echo "$pt_log_entry" | grep -q $success_status; then
            echo "Healthy: Tip is being sent to Pooltool."
            return 0  # Return 0 if the success message is found
        elif echo "$pt_log_entry" | grep -q $failure_status; then
            failure_message=$(echo "$pt_log_entry" | grep -oP '"message":"\K[^"]+')
            echo "Failed to send tip. $failure_message"
            return 1  # Return 1 if the success message is not found
        fi
    done

    echo "Error: Max retries reached."
    return 1  # Return 1 if retries are exhausted
}


```

adamsthws: Added function to check if the tip is successfully being sent to Pooltool

adamsthws (commented):
Added the function to check if the tip is successfully being sent to Pooltool

files/docker/node/addons/healthcheck.sh
```diff
@@ -1,43 +1,241 @@
...
+    echo "Error: Max retries reached."
+    return 1  # Return 1 if retries are exhausted
+}
+
+
+check_db_sync() {
+    # Check if the DB is in sync
+    [[ -z "${PGPASSFILE}" ]] && PGPASSFILE="${CNODE_HOME}/priv/.pgpass"
+    if [[ ! -f "${PGPASSFILE}" ]]; then
+        echo "ERROR: The PGPASSFILE (${PGPASSFILE}) not found, please ensure you've followed the instructions on guild-operators website!" && exit 1
```
rdlrt, line 122: For docker/podman, there arent current instructions - we need to possibly attach a sample reference compose file , wherein PGPASSFILE can be added (or instructions added to be part of priv folder)

TrevorBenson: Is the current Guild Ops DBSync documentation which mentions this considered insufficient in this case? It documents it at the top of the page, and lower down shows examples of using the priv folder as the path.

image

Also, this healtcheck won't impose any new requirements for using `dbsync.sh` in a container that don't already exist using it on bare metal.

* Line 120 comes from line 42 of `set_defaults()` function:
https://github.com/cardano-community/guild-operators/blob/349dffdb7e273685c3045439e0e8d2ec6ebbd865/scripts/cnode-helper-scripts/dbsync.sh#L38-L47

* Lines 121-122 come from lines 52-53 of `check_defaults()` function:
https://github.com/cardano-community/guild-operators/blob/349dffdb7e273685c3045439e0e8d2ec6ebbd865/scripts/cnode-helper-scripts/dbsync.sh#L49-L60

A compose file example is definitely an option. However, I find it can also be a crutch by allowing users to skip reading documentation. If you consider it a "must have" then I think it should be part of another PR and have its own separate testing and review.

rdlrt: > Is the current Guild Ops DBSync documentation which mentions this considered insufficient in this case? It documents it at the top of the page, and lower down shows examples of using the priv folder as the path.

I look at (atleast a big % of) docker users as spoon-fed users - *not intended in any demeaning way, but the primary requirement of using docker [besides virtual env isolation] would be simplicity* , their expectations will be not having to go through document pages beyond docker reference ones.

If someone running a manual setup is moving to docker, I dont expect them to have as many queries. It does kinda begs the question if docker users would also share the pre-reqs on home page - in my experience of reading github issues on IO/intersectmbo issues, a very large section does not.

rdlrt: Talking about dbsync in particular, I would think this would likely be a seperate docker image altogether (which is why I mentioned earlier there is some work to be done):
- Would lighten size (by removing binaries) from current cardano-node image [and vice versa]
- Add more steps in build, like having schema directory specific to dbsync version, ledger-state directory
- Above mentioned PGPASSFILE will be part of it's package
- A dedicated page for operations for dbsync docker image.
In theory , same can be done for smaller components, but there is little value as their setups are mostly the same.
Once above is available, a [docker|podman]-compose will essentially just give a reference on sharing volumes between the pods|containers

files/docker/node/addons/healthcheck.sh
```diff
@@ -68,6 +67,54 @@ check_cncli_db() {
...
+        else error_message_suffix="Retries disabled (HEALTHCHECK_RETRIES=0)"
+        fi
+
+        # Capture the next output from cncli that is related to Pooltool
+        pt_log_entry=$(timeout $log_entry_timeout cat /proc/$process_id/fd/1 | grep -i --line-buffered "pooltool" | head -n 1)
+        if [ -z "$pt_log_entry" ]; then
+            echo "Unable to capture cncli output within $log_entry_timeout seconds. $error_message_suffix"
+            sleep $HEALTHCHECK_RETRY_WAIT  # Wait n seconds then retry
+            continue # Retry if the output capture fails
+        fi
```
TrevorBenson, line 96: This is the part I was mentioning earlier about "fail early". The reason being is we have compounded retries occurring which reduces observability. I want to apologize (profusely) in advance, this comment is going to be pretty verbose to break down what I mean by compounding retries and reduced observability.

---

First, I can definitely see where my earlier statements about using the query tip 60 seconds and relying on the 3 "retries" was confusing. What I was referring to is this portion of the current healthcheck node query tip:

https://github.com/cardano-community/guild-operators/blob/349dffdb7e273685c3045439e0e8d2ec6ebbd865/files/docker/node/addons/healthcheck.sh#L13-L22

It sleeps 60 between the first block and second, and has zero retries. It only relies on the `--healthcheck-retries` of the container configuration. This results in a minimum 15 minutes and maximum 24 minutes before the container actually shows an unhealthy status.  It does not compound retries which makes the status change from healthy -> unhealthy take even longer.

I'll clarify what I mean by compounding retries:

1. The containers own HEALTHCHECK defined in the dockerfile sets the container automatically to use `--health-start-period 5m --healthcheck-interval 5m` without being defined it also gets `--healthcheck-retries 3` from the default number of retries. Here is the container inspect output with time in nanoseconds.

   ```
   podman container inspect preview-ccio-pool | jq '.[].Config.Healthcheck'
   {
     "Test": [
       "CMD-SHELL",
       "/home/guild/.scripts/healthcheck.sh"
     ],
     "StartPeriod": 300000000000,
     "Interval": 300000000000,
     "Timeout": 100000000000
   }
   ```
2. The check for ptsendtip does a loop with `HEALTHCHECK_RETRIES` (default 20), with `log_entry_timeout` (set to 60), if it fails we wait `HEALTHCHECK_RETRY_WAIT` (default 3), and start again.

   * The check for ptsendtip has the potential to run for 21 minutes runtime (60 seconds per loop + 3 second wait, times 20).
   * The container healthcheck timeout is set to 100 seconds. Exceeding timeout causes an unhealthy response
   * This status changing from healthy to unhealthy requires exhausting the containers retry value, multiplied by the Interval value.

That is a minimum of 15 and maximum of almost 27 minutes (including runtime).

I see multiple options as a solution:

1. Drop the retry logic from `check_cncli_send_tip()`. Let it run once for 60 seconds and exit 1 if it doesn't see a block. This makes it similar to the query tip in the `timeout` and all other checks in regards to no retries (except the KOIOS version of check_node).

2. Keep the retry logic in `check_cncli_send_tip()` and set the HEALTCHECK_RETRIES to default to 0. 

3. Add retry logic to every check function in the healthcheck script and set HEALTCHECK_RETRIES to default to 0.

Each option results in a 15-21 minute window before knowing the container is unhealthy when the container configuration for healthcheck intervals & retries and all script variables are left at current/default values.

I'm torn between 3 and 1 as my favorite option. Improving observability even further requires either the operator to adjust the startup, interval, timeout and retries when creating the container OR for us to come to a consensus about adjusting the Dockerfile HEALTHCHECK statement when building the container image.

files/docker/node/addons/healthcheck.sh
TrevorBenson: > Ye - I agree. That - if added - does not need to be part of this PR indeed

Sounds good, after this, and some other documentation issues/PR are resolved I'll open one and start a compose example.

> Talking about dbsync in particular, I would think this would likely be a seperate docker image altogether [...]

Nice. So far I've been running the upstream db-sync image, as well as ogmios, and guild ops for node, cncli, and mithril. I'm hereby offering to assist with a new container if any help is needed.

> In theory , same can be done for smaller components, but there is little value as their setups are mostly the same. For instance, for something like ogmios - we dont need a custom image, official image is perfectly usable.

Yep. The socket, configs (files) and just about every other subdir of CNODE_HOME lend themselves to volume usage. Not scripts though. The excessive work trying to make scripts a shared volume wasn't worth the time I initially spent to get it working. I've found a container environment file with all vars + a completely stock scripts directlry and env file inside the container to be the simplest way to run separated containers.

files/docker/node/addons/healthcheck.sh
adamsthws: Great, thankyou for the thorough explination.

>Drop the retry logic from check_cncli_send_tip()

My gut feeling is that this would be the prefered approach, minimising the code to add/maintain.


> Let it run once for 60 seconds

To test how often it times out, I'm in the process of running the function 100 times with varying timeout values... 
 - timeout=60   - Result (TBD): x% timeout rate (x-fail/x-success). 
 - timeout=90   - Result (TBD): x% timeout rate (x-fail/x-success). 
 - timeout=120 - Result (TBD): x% timeout rate (x-fail/x-success). 

After removing retries from the function, if it times out, it then must wait another 5mins before retring again ```--healthcheck-interval 5m```... I propose to change the timeout within the function to something larger, maybe 90s-120s (depending on the results of my testing). 

E.g timeout=90 (rather than timeout=60) increases the observable fail time by a 30s while decreasing the observable sucess time by relatively larger 4.5min (by avoiding unnecassary timouts and retries). 

I've set it at 90s for now and will let you know the results of the testing when it completes.

TrevorBenson: > What would be the maximum time you would expect between tips?

Anecdotally I have seen **Tip (diff)** reach 60 seconds quite often. Occasionally a bit higher at 100-120 seconds. The KOIOS addition was because the original query tip I had twice over a few months hit 60 seconds on repeated retries and marked it unhealthy causing some investigations. After adding KOIOS, or even disabling KOIOS and using the original query tip, I don't remember it happening in the last year seeing unhealthy on the node.

----

In rare, extreme events, it has exceeded 5 minutes. Node 1.35, or 1.34, had one event long ago that was exceeding 5 minutes. Not sure if this linked issue/comment showing 8 minute delay between blocks is the one, or something else as I didnt read the whole thread. The time frame seems like it could be the one I'm thinking of, and I don't recall at the moment another lasting longer than 5.

https://github.com/IntersectMBO/cardano-node/issues/4421#issuecomment-1240978118

Again, these are rare "extreme" events, and the healthcheck in theory wouldn't even point this out with the current configuration anyway when it lasts less than the `--healthcheck-retries` multiplied by the interval, we should in theory only see this in grafana/prometheus and in theory the container runtimes would still be showing a healthy container even with timeout=60.

files/docker/node/addons/healthcheck.sh
```diff
@@ -1,43 +1,241 @@
...
+        for (( CHECK=0; CHECK<=HEALTHCHECK_RETRIES; CHECK++ )); do
+            if [[ "$FIRST" -eq "$SECOND" ]]; then
+                echo "We're healthy - node: $FIRST == koios: $SECOND"
+                return 0
+            elif [[ "$FIRST" -lt "$SECOND" ]]; then
+                sleep "$HEALTHCHECK_RETRY_WAIT"
+                FIRST=$($CCLI query tip --testnet-magic "${NWMAGIC}" | jq .block)
+            elif [[ "$FIRST" -gt "$SECOND" ]]; then
+                sleep "$HEALTHCHECK_RETRY_WAIT"
+                SECOND=$($CURL "${KOIOS_URL}" | $JQ '.[0].block_no')
```
TrevorBenson, line 176:
```suggestion
                SECOND=$($CURL "${URL}" | $JQ '.[0].block_no')
```

TrevorBenson:
Separately, I've been considering for awhile retries (now HEALTHCHECK_RETRIES) being dropped from 20, to 0 or 3. I originally set it to 20 retries and (now HEALTHCHECK_RETRY_WAIT) to 3 second sleeps so that the KOIOS method would at max run as long as the 60 seconds of a query tip without KOIOS. This was never really required, but its been there ever since.

I tested last night manually in a loop of 10 with a 20 second sleep 10/10 returned in under 1 second, never needing a retry. 10/10 also incremented at least 1 block.

I didn't bother to do multiple passes and calculate the ratio. However given the retries at the container level, I am pretty confident we could drop this from 20 and improve the observability (slightly) by failing earlier.

I'm interested in your feedback. What do you think about changing the default values of HEALTHCHECK_RETRIES so even KOIOS will fail earlier?

files/docker/node/addons/healthcheck.sh
adamsthws: > To test how often it times out, I'm in the process of running the function 100 times with varying timeout values...

Here are the results:
timeout=60   - Result: 9% timeout frequency (9-fail/91-success).
timeout=90   - Result: 4% timeout frequency (4-fail/96-success).
timeout=120 - Result: 2% timeout frequency (2-fail/98-success).

> I had twice over a few months hit 60 seconds on repeated retries and marked it unhealthy causing some investigations

In this limited testing sample, it has already failed sucessively (twice in a row) - (with lower timout of 60s)... So I feel using a higher timout would mean less chance of failure three times in a row, causing unnceassary investigations.

...How would you feel about making timeout=120?

adamsthws:
> I tested last night manually in a loop of 10 with a 20 second sleep 10/10 returned in under 1 second, never needing a retry. 10/10 also incremented at least 1 block.
> 
> Given the retries at the container level, I am pretty confident we could drop this from 20 and improve the observability (slightly) by failing earlier.
> 
> I'm interested in your feedback. What do you think about changing the default values of HEALTHCHECK_RETRIES so even KOIOS will fail earlier?

If it rarely (or never) fails and needs a retry, I would be in favor of removing the Koios retry loop from ```check_node()``` entirely. Simplifying the code, making it easier to maintain and understand. The only healthcheck function then that would have a retry would be ```check_process()```

In which case, there might also be an argument to remove variables: $HEALTHCHECK_RETRY_WAIT, $HEALTHCHECK_RETRIES in favor of:
- Hardcoded Sensible Retry Defaults (Within ```check_process()```)
  - Preventing Misconfiguration
  - Simplicity and Ease of Use
  - Reducing User Cognitive Overload
  - Reduced Support Overhead

TrevorBenson:
> If it rarely (or never) fails and needs a retry, I would be in favor of removing the Koios retry loop from `check_node()` entirely. Simplifying the code, making it easier to maintain and understand. The only healthcheck function then that would have a retry would be `check_process()`

Well, not "never". There are some occasions where the node is ahead, like when a block is produced locally and healcheck is almost at the exact moment. Same in reverse, KOIOS being ahead, but rarely for more than 3 seconds. Since your open to even removing the retries altogether I'll create a POC and give it a full 100 runs using a current interval of 5 minutes. 

I'll track the total times:
* That the node is ahead
* That koios is ahead 
* That they are in sync
* The number of consecutive out of sync responses occur. 

I'll also try 4, 3 and 2 minute intervals. We could potentially improve the observability, reducing the time something is wrong before the status changes to unhealthy. As long as multiple systems are not sharing a single outbound IP for KOIOS queries I don't believe this should push the limits of the free tier.

adamsthws:
> There are some occasions where the node is ahead, like when a block is produced locally and healcheck is almost at the exact moment. Same in reverse, KOIOS being ahead, but rarely for more than 3 seconds.

I wonder if including an allowable drift for the koios check, (similar to ```check_db_sync()```) would make sense in place of retries? ... But perhaps we're just swapping one type of complexity for another with that line of thought?

TrevorBenson:
> I wonder if including an allowable drift for the koios check, (similar to `check_db_sync()`) would make sense in place of retries? ... But perhaps we're just swapping one type of complexity for another with that line of thought.

I like the thought, but you might be correct that we just swap complexity.

A single line for "healthy" we get something like this for a drift of 1: 
```
if [[ "$FIRST" -eq "$SECOND" || "$((FIRST - SECOND))" -le 1 && "$((FIRST - SECOND))" -ge -1 ]]; then
```

It's not terribly complex, but not the simplest either. It could be split up w/ `elif` though. The drift could also a variable, though I think it brings us back to the discussion of cognitive overload from earlier.

TrevorBenson:
@adamsthws So the results were pretty interesting. 

FWIW, I set the healthcheck_poc.sh so that it leverages the KOIOS_API_TOKEN to be sure I didn't get any 429's as this testbed shares the WAN IP with multiple systems querying koios. 

---

# Tests

* Intervals of 300, 240, 180, 120
  * 300 and 240 used 100 runs each
  * 180 and 120 used 200 runs each
    * So that 180 and 120 had more samples with more overlap to the 300 interval test.
* Tests performed in parallel
  * Over the same time periods
  * On the same chain (Preview)

## Results

### 300 second intervals
* Run time 2025-01-17-06:10:51 AM UTC - 2025-01-17-02:26:57 PM UTC
* Results
  * 5/100 node less than koios: 5%
  * 95/100 node synched to koios: 95%
  * 0 node greater than koios: 0%
* Consecutive unsynchronized checks
  * 0

### 240 second intervals
* Run time 2025-01-17-06:10:53 UTC - 2025-01-17-12:47:58 PM UTC
* Results
  * 1/100 node less than koios: 1%
  * 99/100 node synched to koios: 99%
  * 0/100 node greater than koios: 0%
* Consecutive unsynchronized checks
  * 0

### 180 second intervals
* Run time 2025-01-17-06:10:55 AM UTC - 2025-01-17-04:10:08 PM UTC
* Results
  * 5/200 node less than koios: 2.5%
  * 195/200% node synched to koios: 97.5%
  * 0 node greater than koios: 0%
* Consecutive unsynchronized checks
  * 0

### 120 second intervals
* Run time 2025-01-17-06:10:58 AM UTC - 2025-01-17-12:51:10 PM UTC
* Results
  * 5/200 node less than koios: 2.5%
  * 195/200% node synched to koios: 97.5%
  * 0 node greater than koios: 0%
* Consecutive unsynchronized checks
  * 0

## Summary
* No tests experienced a single consecutive unsynchronized event. I think this confirms 4 consecutive unsynchonized events are fairly unlikely in the nodes KOIOS test.
* The unsynchronized results among each interval were not from the same issue/time period.
  * The node/koios tips in the DB shows the chain had progressed, even on the shortest distance between two results for a different test (2 minutes)
    ```
    15|2025-01-17-06:20:52 AM|300|lt|1|2880268|2880269
    18|2025-01-17-06:22:58 AM|180|lt|1|2880273|2880274
    ```
* The interval 300 being double the ratio of the other tests I believe to be just luck of the draw, I doubt it would come out the same if run again.

If reducing this interval does not negatively impact any other healthchecks being adding, and the healthchecks start only after the ledger has finished loading (in case of unclean shutdowns, ledger replays during major version changes, etc.), I see no drawbacks to improving the observability of the container health with shorter intervals.

I can post the script to a gist if you want to replicate the tests yourself or test different networks.

files/docker/node/addons/healthcheck.sh
TrevorBenson: Do you happen to have details about consecutive failure/timeouts? For example based on the 300 second interval if it's never observed that we have 2 consecutive failures over 100 attempts, then it still results in 0 status changes to unhealty.

To bump the timeout even higher I'd probably want to see 2, possibly 3, consecutive timeoutes reached based on X interval being tested. If it reaches 50-66% of the required consecutive failures to cause the node to be marked unhealthy then I would definitely consider bumping it up.

However, if the results were similar to KOIOS where there was never more than 1 failure and 0 consecutive failures I would prefer leaving the internal timeout for the check at 60. Raising it would also require an increase of the Dockerfile HEALTHCHECK settings timeout (aka `--healthcheck-timeout`) value currently at 100.

If you don't have the details on how many consecutive failures were observed [this gist](https://gist.github.com/TrevorBenson/090d9d937d44185f28cd4bd040e5c528) contains the POC I did to gather the details for KOIOS. I won't get time to setup a ptsendtip until sometime this weekend, which would push my ability to run this test out to Sunday or sometime next week.

adamsthws: > Do you happen to have details about consecutive failure/timeouts? For example based on the 300 second interval if it's never observed that we have 1 consecutive failures over 100 attempts, then it still results in 0 status changes to unhealty.

These tests had no interval... I didn't feel adding an interval would effect the results in any meaningful way.

- timeout=60 - Result: 9% timeout frequency (9-fail/91-success). 
   - failed consecutively 1 time. (interval=0)
 
- timeout=90 - Result: 4% timeout frequency (4-fail/96-success).
   - failed consecutively 0 times. (interval=0)
 
- timeout=120 - Result: 2% timeout frequency (2-fail/98-success).
   - failed consecutively 0 times. (interval=0)
  

> To bump the timeout even higher I'd probably want to see 2, possibly 3, consecutive timeoutes reached based on X interval being tested. If it reaches 50-66% of the required consecutive failures to cause the node to be marked unhealthy then I would definitely consider bumping it up.

I'm in the process of re-running it with timeout=60, interval=300s... I'll keep you posted.

> Raising it to 120 would also require an increase of the Dockerfile HEALTHCHECK settings timeout (aka `--healthcheck-timeout`) value currently at 100.

Oh yes, thanks for pointing that out. Let's see how the re-test goes.

> If you don't have the details on how many consecutive failures were observed [this gist](https://gist.github.com/TrevorBenson/090d9d937d44185f28cd4bd040e5c528) contains the POC I did to gather the details for KOIOS. 

This looks great. Nice use of sqlite!
Here's my test, it's not as elegant but it gets the job done:
 ```https://gist.github.com/adamsthws/dce0263bea2302047660b9c8ea458cbd```

adamsthws:
> I can post the script to a gist if you want to replicate the tests yourself or test different networks.

100% - running your test on my side now. I'll let you know the results.

adamsthws:
> Running your test on my side now. I'll let you know the results.

Results after running both of our two tests at my end...

Test: check_node()
(healthcheck_pos.ch - [Gist](https://gist.github.com/TrevorBenson/090d9d937d44185f28cd4bd040e5c528))
(Interval=120) (Tries=300)
 - Where "Status" != "synchronized":   43/300 (14.33%)
 - Where "Status" == "lt":                     4/300 (1.33%)
 - Where "Status" == "gt":                    39/300 (13%)
 - Failed consecutivly:                           1/300 (0.33%)
    - (It failed twice in a row one time)


Test: check_cncli_send_tip()
(test_check_cncli_send_tip.sh - [Gist](https://gist.github.com/adamsthws/dce0263bea2302047660b9c8ea458cbd))
(Interval=120) (Tries=300) (Timeout=60)
 - Failed checks:                                                     30/300 (10%)
 - Failed consecutivly (twice in a row):                   4/300 (1.33%)
 - Failed consecutivly (three times in a row):         1/300 (0.33%)
     - (It failed three times in a row one time)

TrevorBenson:
> So with `timeout=60s` and `--healthcheck-interval 5m`, `check_cncli_send_tip()` would (probabilistically) become unhealthy every 3.47 days

It's clear there is a need to account for your findings.

I suspect the "middle ground" here may be adjusting current HEALTHCHECK values, as well as providing documentation for optimizing based on the use case (ENTRYPOINT_PROCESS). 

---

## Thoughts

### Probabilities

A quick glance at your formula for probabilities I realized I mispoke in earlier statements 🤦🏼 

I described one of our monitoring solution's concept of "retries" before a service status changes to offline/unhealthy, instead of dockers concept. Instead of 1 initial failure + X retries to change a status, for Docker it's just X retries, or "FailingStreak: X" == `--health-retries X`. Failingstreak is available in `docker container inspect` under State.Health.

I'm a bit tired ATM, but unless my mental math is off, doesn't that bring us to unhealthy status once per 25 hours / 1.04 days (at 300 interval) or once per 10 hours if we wanted to reduce to an interval of 120?

### Node tests
Your node results are interesting.

My tests were all in the US from various systems, but still pretty low latency to a very high number of overall stake pools, and total delegation. In regions with even higher overall latency to the majority of nodes/delegation (possibly AU?), I suspect we'd see even more failures.


#### Additional Data

It might be interesting to see the following output providing details about the consecutive event observed:
```
SELECT * FROM checks WHERE status != 'synchronized';
```

This is not required, just "interesting". If you you decide to grab/provide this for review, or any other queries you might find interesting among the data, a gist link would be fine.

### Send Tip Tests

Given your current results from timeout=60 I doubt if timeout=100 (or even timeout=120) would be sufficient to combat an event with such high probability to occur. every few days (more than once an epoch, ~105 times annually) 

#### Options

1. We could retest w/ 100 and 120 to get more data points. 
   * Possibly do 60 too, and all in parallel starting the script more than once. 
   * Parallel is just to confirm failures at 60 don't drop while we evaluate the 100 & 120 results. If 60 is lower consecutive failures our findings from 100/120 might be skewed as well.
2. Consider some form of check enhancement. 
   * Possibly reduce false positives by borrowing from check_node logic? Using the tip when needed.

```
CCLI=$(which cardano-cli)
FIRST=$($CCLI query tip --testnet-magic ${NWMAGIC} | jq .block)
pt_log_entry=$(timeout $log_entry_timeout cat /proc/$process_id/fd/1 | grep --line-buffered "Pooltool" | head -n 1)
if [ -z "$pt_log_entry" ]; then
    SECOND=$($CCLI query tip --testnet-magic ${NWMAGIC} | jq .block)
    if [[ "$FIRST" -eq "$SECOND" ]]; then
        echo "Unable to capture cncli output within $log_entry_timeout seconds, but node has not moved tip."
        return 0
    else
        echo "Unable to capture cncli output within $log_entry_timeout seconds."
        return 1  # Return 1 if the output capture fails
fi

```

adamsthws:
> I described one of our monitoring solution's concept of "retries" before a service status changes to offline/unhealthy, instead of dockers concept. Instead of 1 initial failure + X retries to change a status, for Docker it's just X retries, or "FailingStreak: X" == `--health-retries X`. Failingstreak is available in `docker container inspect` under State.Health.

Thanks for pointing that out... I was thinking about it in terms of docker's behavour - where 3 consecutive failures = unhealthy.

> I'm a bit tired ATM, but unless my mental math is off, doesn't that bring us to unhealthy status once per 25 hours / 1.04 days (at 300 interval) or once per 10 hours if we wanted to reduce to an interval of 120?

I re-calculated to double check and I believe my previous working was correct.
By my calculations, with a 10% failure rate (and "FailingStreak: 3" == `--health-retries 3`):
 - 120 interval would become unhealthy every 1.38 days
 - 300 interval would become unhealthy every 3.47 days 
 
 > My tests were all in the US from various systems, but still pretty low latency to a very high number of overall stake pools, and total delegation. In regions with even higher overall latency to the majority of nodes/delegation (possibly AU?), I suspect we'd see even more failures.
 
 That's a good point. FWIW I'm testing from the UK.
 
 > Additional Data
> It might be interesting to see the following output providing details about the consecutive event observed:
> ```SELECT * FROM checks WHERE status != 'synchronized';```
> This is not required, just "interesting". If you you decide to grab/provide this for review, or any other queries you might find interesting among the data, a gist link would be fine.

Unfortunately the container has been re-created since running your test and the results db has not persisted. Happy to re-run the test if you think it would be valuable?

> Given your current results from timeout=60 I doubt if timeout=100 (or even timeout=120) would be sufficient to combat an event with such high probability to occur. every few days (more than once an epoch, ~105 times annually)

I've tested with timout=90 and now awaiting timout=120 test to complete... Once we have these results it might give us a better idea of the values to test for in the next round (while testing all simultaneously). Here's the results of Timeout=90...

- Test: check_cncli_send_tip()
  (test_check_cncli_send_tip.sh - [Gist](https://gist.github.com/adamsthws/dce0263bea2302047660b9c8ea458cbd))
  (Interval=120) (Tries=300) (Timeout=90)
   Failed checks: 11/300 (3.67%)
   Probablilty of 3 consecutive failed checks: 0.00495%
   Probabilistic frequency of 3 consecutive failed checks: 70.24 days

TrevorBenson:
> Unfortunately the container has been re-created since running your test and the results db has not persisted. Happy to re-run the test if you think it would be valuable?

No need. The parallel run with 60/90/120 would be the most useful, as we can see if timeout=60 results in the same level of failures as your first pass.

If you can persist the DB on the parallel run it would be useful to investigate each consecutive failure. The comparison of each overlapping check for the different timeout values should be invaluable to determine if the adjustment really addresses the probabilities calculated previously, or if you just had a lucky run, which seeing the parallel timeout=60 will determine.

TrevorBenson:
> Thanks for pointing that out... I was thinking about it in terms of docker's behavour - where 3 consecutive failures = unhealthy.

Great. I must have multiplied by 100 to think about the % on the 2 failure scenario, then just continued with that number when I added the third failure.

Sleep is good, I should try it more often 😅.

adamsthws:
> I've tested with timout=90 and now awaiting timout=120 test to complete... Once we have these results it might give us a better idea of the values to test for in the next round (while testing all simultaneously). Here's the results of Timeout=90...
> 
>     * Test: check_cncli_send_tip() - [Gist](https://gist.github.com/adamsthws/dce0263bea2302047660b9c8ea458cbd)
>       (Interval=120) (Tries=300) (Timeout=90)
>       Failed checks: 11/300 (3.67%)
>       Probablilty of 3 consecutive failed checks: 0.00495%
>       Expected frequency of unhealthy container status: 70.24 days


Results of timout=120:
 
- Test: check_cncli_send_tip() - [Gist](https://gist.github.com/adamsthws/dce0263bea2302047660b9c8ea458cbd)
  (Interval=120) (Tries=300) (Timeout=120)
  Failed checks: 3/300 (1%)
  Probablilty of 3 consecutive failed checks: 0.0001%
  Expected frequency of unhealthy container status: 3472 days (9.5 years)
  
Quick observation:
As the length of timout increases linearly, the failure rate decreases exponentially .
As the failure rate decreases linearly, the probability of three consecutive failures decreases exponentially .

I will now go on to test simultaneously to verify results.
It's starting to look like ```timout=120s``` might have an acceptable failure rate.

TrevorBenson:
> Results of timout=120:
> 
>     * Test: check_cncli_send_tip() - [Gist](https://gist.github.com/adamsthws/dce0263bea2302047660b9c8ea458cbd)
>       (Interval=120) (Tries=300) (Timeout=120)
>       Failed checks: 3/300 (1%)
>       Probablilty of 3 consecutive failed checks: 0.0001%
>       Expected frequency of unhealthy container status: 3472 days (9.5 years)
> 
> 
> Quick observation: As the length of timout increases linearly, the failure rate decreases exponentially . As the failure rate decreases linearly, the probability of three consecutive failures decreases exponentially .
> 
> I will now go on to test simultaneously to verify results, using the following timout values: 60, 90, 100, 110, 120.


## Send Tip
Fabulous. With confirmation from the parallel data showing 60 was close to or the same as your first while we can compare in SQL for every other timeout during the same period should make the choice quite simple.

## Node

The only concerning thing remaining was your results from check_node POC. The 13% result being much higher than I observed (worst at 5% w/ 0 consecutive). The probability increase is dramatic, 5% being 0.000125/0.012%, 10% (send tip) being 0.001/0.1% and 13% being 0.002197/0.22%. Essentially 800% more likely than in any test I performed, and ~120% more likely than the send tip healthcheck failure ratio observed w/ timeout=60.

Due to this I'm thinking about an adjustment to the logic for check_node using KOIOS.
* Allow up to 1 block drift in either direction or fully in sync to return 0.
* Use 1-2 retries (instead of current 20) with a 3-4 second sleep.

I suspect either will significantly reduce the POC failure rates. Do you have a preference or other options to suggest? 

After your response I'll create an adjustment to the node healthcheck POC and run them in parallel for easy comparison.

----

FWIW, I think moving this out of draft is quite close. I do need to come up with something for mithril signer. I don't want that to delay this work getting reviewed, so if I don't find time to implement/test something before the next POC test finishes, I'll just add that to a list of other PR's I need to start work on.

Thanks for all the time an energy you've spent collaborating on this!

adamsthws:
> The 13% result being much higher than I observed

I wonder if this large discrepancey was due to my not using ```KOIOS_API_TOKEN``` and hitting api rate limiting. Though, I don't have a great deal of nodes on this WAN IP. I'd be happy to re-run it with an api key to verify however, even if using your 5% failure rate, that would be high enough that it should be accounted for with one of your two proposed adjsutments (drift/retries). Plus, the healthcheck should probably be able to account for operators running without a Koios api key.


> Due to this I'm thinking about an adjustment to the logic for check_node using KOIOS.
> 
> * Allow up to 1 block drift in either direction or fully in sync to return 0.
> 
> * Use 1-2 retries (instead of current 20) with a 3-4 second sleep.

Great!
The time it takes to test if it has drifted by >1 blocks would be greater than the time it takes test with (short) retries.
The drift feels like a nicer solution for some reason but retries would fail faster. 
I imagine either would work well... Which is your preference?


> FWIW, I think moving this out of draft is quite close. I do need to come up with something for mithril signer. I don't want that to delay this work getting reviewed, so if I don't find time to implement/test something before the next POC test finishes, I'll just add that to a list of other PR's I need to start work on.
> 
> Thanks for all the time an energy you've spent collaborating on this!

Thankyou for saying so...It's been a pleasure working through it with you :)
And thanks for being so open to the suggestions!

TrevorBenson:
> I imagine either would work well... Which is your preference?

Currently I think allowing a minor block drift is my preference. We reduce the queries to KOIOS, which shoudn't really be an issue unless multiple machines share an IP. However, it also aligns with how we (and others already) do the DB checks where they expect some level of drift, and even allow it to be pretty large.

I'll create a POC test and run in parallel for a 1 block and a 2 block drift, as well as 1 retry and 2 retry. Short of a major difference I think a 1 (or 2) block drift would be the direction I'd lean. I'll let you know later today when the updated POC scripts are ready, in case you want to test on your end as well.

TrevorBenson:
Apologies on the delay, came down with a cold and was decided to rest for the remainder of Friday before a long trip this week. 

While putting together the new POC scripts I did some extremely short interval testing (1 second between checks) to validate the script worked as intended. I ran into a very small amout of koios and node being in sync, but 1 second later I was on the same block, but the koios endpoint was behind.

```
id   timestamp               interval  status        count  node      koios     drift  allowed_drift  allowed_retries  retries_used
---  ----------------------  --------  ------------  -----  --------  --------  -----  -------------  ---------------  ------------
96   2025-01-26-08:05:53 PM  1         synchronized  6      11403349  11403349  0      0              1                0
97   2025-01-26-08:05:56 PM  1         gt            1      11403350  11403346  4      0              1                1
98   2025-01-26-08:05:57 PM  1         synchronized  1      11403350  11403350  0      0              1                0
99   2025-01-26-08:06:00 PM  1         gt            1      11403350  11403348  2      0              1                1
100  2025-01-26-08:06:02 PM  1         synchronized  1      11403350  11403350  0      0              1                0
```


```
id   timestamp               interval  status        count  node      koios     drift  allowed_drift  allowed_retries  retries_used
---  ----------------------  --------  ------------  -----  --------  --------  -----  -------------  ---------------  ------------
272  2025-01-26-08:13:12 PM  1         synchronized  72     11403370  11403370  0      2              0                0
273  2025-01-26-08:13:13 PM  1         gt            1      11403370  11403367  3      2              0                0
274  2025-01-26-08:13:15 PM  1         synchronized  1      11403370  11403370  0      2              0                0
```

The healthcheck_poc2.sh is available in [this gist](https://gist.github.com/TrevorBenson/4f9c747083866f94595e316d416fbfb1)

adamsthws:
Hey, I wanted to give you a quick update as I've not made comment for a while - I've been quietly testing, I just didn't want you thinking I'd lost interest.

When testing the function ```check_cncli_send_tip()``` with 300 retrires the results are slightly varying so I've increased retries to 1000 for a larger sample size (which is taking a long time to complete).
[Here the test method](https://gist.github.com/adamsthws/dce0263bea2302047660b9c8ea458cbd)

I also have results from running your: [healthcheck_poc2.sh](https://gist.github.com/TrevorBenson/4f9c747083866f94595e316d416fbfb1). I've not had chance to look at the results yet but will get those to you soon.

Thanks for your patience and I hope you're feeling better now!
Cheers, Adam

TrevorBenson:
> Hey, I wanted to give you a quick update as I've not made comment for a while - I've been quietly testing, I just didn't want you thinking I'd lost interest.
> 

Great, I also got a bit busy with travel for work so haven't been as active as I initially planned on the poc2 phase.

> When testing the function `check_cncli_send_tip()` with 300 retrires the results are slightly varying so I've increased retries to 1000 for a larger sample size (which is taking a long time to complete). [Here the test method](https://gist.github.com/adamsthws/dce0263bea2302047660b9c8ea458cbd)

👍🏼 

> I also have results from running your: [healthcheck_poc2.sh](https://gist.github.com/TrevorBenson/4f9c747083866f94595e316d416fbfb1). I've not had chance to look at the results yet but will get those to you soon.

Same, I ran what I described, but hadn't gotten back around to creating the ratios and examining probabilities. I hope to get a few minutes this evening to jump into it and update this thread.

> Thanks for your patience and I hope you're feeling better now! Cheers, Adam

Just started to after the flights. At least my two weeks in the EU won't be suffering 😆 . Thanks!

TrevorBenson:
@adamsthws Apologies for the delays. I've updated the gist with results from my past tests. Given the potential for drift between the node and koios, at least based on the results I had, an allowed drift appears to be a more resilient option than including additional retries.

I'm interested in the findings from your tests and if you observed similar results or had any differences that should be considered.

Thanks

adamsthws:
> Apologies for the delays. 

No need to apologise, I understand you've been travelling... I've been travelling this week too on a snowboarding trip to France :)  Hope you're having fun in Europe! 

> I've updated the gist with results from my past tests. 

Great, I've done the same.

> I'm interested in the findings from your tests and if you observed similar results or had any differences that should be considered.

It looks like our results are quite similar. I'd be happy to re-run the test with a larger amount of checks if you think a larger sample size would provide any benefit.

Thanks

adamsthws:
I've added my results for my sendtip tests to the gist [here](https://gist.github.com/adamsthws/dce0263bea2302047660b9c8ea458cbd)

> We could retest w/ 100 and 120 to get more data points. Possibly do 60 too, and all in parallel.
>
> Parallel is just to confirm failures at 60 don't drop while we evaluate the 100 & 120 results. If 60 is lower consecutive failures our findings from 100/120 might be skewed as well.


In the initial test with timout=60 test I measured a 9% failure rate. In the latest results (while testing differeing timout values simultaneously) there was a 8.4% failure rate with timout=60, so the failure rate seems to be fairly consistent.

TrevorBenson:
> > Apologies for the delays. 
> 
> No need to apologise, I understand you've been travelling... I've been travelling this week too on a snowboarding trip to France :)  Hope you're having fun in Europe! 
> 
> > I've updated the gist with results from my past tests. 
> 
> Great, I've done the same.
> 
> > I'm interested in the findings from your tests and if you observed similar results or had any differences that should be considered.
> 
> It looks like our results are quite similar. I'd be happy to re-run the test with a larger amount of checks if you think a larger sample size would provide any benefit.
> 
> Thanks 
> 
> 

Given a decent overlap in results I'm inclined to push for an allowed drift. To account for the potential of the AU to have a larger drift I'm thinking of 6 blocks for the moment, which should equate to around 2 minutes, and also cover when one koios endpoint is behind on sync more than others.

TrevorBenson:
> I've added my results for my sendtip tests to the gist [here](https://gist.github.com/adamsthws/dce0263bea2302047660b9c8ea458cbd)
> 
> > We could retest w/ 100 and 120 to get more data points. Possibly do 60 too, and all in parallel.
> >
> > Parallel is just to confirm failures at 60 don't drop while we evaluate the 100 & 120 results. If 60 is lower consecutive failures our findings from 100/120 might be skewed as well.
> 
> 
> In the initial test with timout=60 test I measured a 9% failure rate. In the latest results (while testing differeing timout values simultaneously) there was a 8.4% failure rate with timout=60, so the failure rate seems to be fairly consistent.

One thing in the results is confusing me:

>Quick note: 'unhealthy status' means 3 consecutive failures
[ ... ]
```
docker exec cardano-mainnet-bp_cncli-pt-send-tip4 tail -n 6 /home/guild/.scripts/result.log
-----
Tests completed: 1000 | Timeout: 140 | Interval: 180
Healthy: 999 | Unhealthy: 1 | Consecutive Failures: 0
Measured failure rate: .10000000%
Probability of 3 consecutive failures: 0%
Expected frequency of unhealthy status: Never.
```


If a single unhealthy status means 3 consecutive failures occurred in the test, and his test shows **Unhealthy: 1**, how is the probability 0%?

---

Separately.

1. From your results, what are you inclined to do for the pt send tip healthcheck?
2. For other health checks where I was considering Drift (currently 6), do you have any preference or suggestions to use less or more drift?

I think we have enough details to move forward with a final version and move this out of draft. However you want to include your work is fine with me (PR on upstream source, or anything else you prefer to include your commits). 

I'll confer with you before moving the PR out of draft so we're sure we didn't miss/overlook anything along the way 

Thanks again, and I hope you're enjoying the snowboarding! I should be going again in a couple of weeks 🏂

TrevorBenson:
> My feeling is probably leaning more toward option 2 what do you think?

I like option 2 for the reduction in both documentation and cognitive load, especially with the main node still becoming the unhealthy component if the tip no longer progresses.

As @rdlrt mentioned containers users may be those who spend less time reading documentation on how to customize container behavior. I think option 2 creates a winning combination for monitoring health which also likely reduces users asking container health questions in the Koios Support channel.

---

This week I'll sync the parts from my POC content into the sidecar-healthchecks branch. Then we can merge your commits (however you like).

I'll use this finalized version to replace the healthcheck in each of my containers (I still don't run a send tip though), and let it run for a couple of days. Then we can transition this out of draft and request reviews/approvals.

adamsthws:
> I think option 2 creates a winning combination

New [PR](https://github.com/chaincrucial/guild-operators/pull/2) - improvements to test_check_cncli_send_tip.sh
(I've added you as a co-author on this)

New test [gist](https://gist.github.com/adamsthws/45b3ca80fa3ce387451c6db882d51c33) - for test_check_cncli_send_tip.sh
(It's probably wise to have a second person test and validate my results)

files/docker/node/addons/healthcheck.sh
```diff
@@ -1,43 +1,242 @@
...
-CCLI=$(which cardano-cli)
+######################################
+# Do NOT modify code below           #
+######################################
 
-if [[ "$NETWORK" == "guild-mainnet" ]]; then NETWORK=mainnet; fi
+[[ ${0} != '-bash' ]] && PARENT="$(dirname $0)" || PARENT="$(pwd)"
+# Check if env file is missing in current folder (no update checks as will mostly run as daemon), source env if present
+[[ ! -f "${PARENT}"/env ]] && echo -e "\nCommon env file missing in \"${PARENT}\", please ensure latest guild-deploy.sh was run and this script is being run from ${CNODE_HOME}/scripts folder! \n" && exit 1
+. "${PARENT}"/env offline
```
adamsthws, line 25: The healthcheck.sh script currently lives here: ```~/.scripts```, but this expects it to be run from here: ```$CNODE_HOME/scripts```

TrevorBenson: The `$CNODE_HOME/scripts/env` file is designed such that to source it for variables requires the script to be in the same path. Otherwise this results in missing variables (KOIOS_API_HEADERS, etc. etc.) this is why there was a change to the dockerfile_bin lines 97-99 changing its path and 108, changing the HEALTHCHECK CMD (i.e. `--health-cmd`)

https://github.com/cardano-community/guild-operators/pull/1842/files#diff-589eef580db414083df62cbd394cc6d89e8ad464d0a1e6772d051c4837f63f81R97-R99

https://github.com/cardano-community/guild-operators/pull/1842/files#diff-589eef580db414083df62cbd394cc6d89e8ad464d0a1e6772d051c4837f63f81R108

adamsthws: > there was a change to the dockerfile_bin lines 97-99 changing its path when added to the container

Whoops, I missed that... Applogies!

TrevorBenson:
Big thanks to @adamsthws for the final push to add additional healthchecks and get us back to testing. Our plan is to generate the testing container from alpha and then run it for each testnet we participate in giving healthchecks time to cook.

@rdlrt @Scitz0 I believe this PR is finally ready for review.

Scitz0:
Docker isn't really my thing so I will leave review to @rdlrt or @hodlonaut who is better suited to check changes.

rdlrt:
Think we're all relatively comfortable with this PR, happy to get it merged to alpha for wider testing and fix any further issues iteratively. Also, this has downstream impacts for any existing compose solutions (eg: Koios lite-node will need changes), so would be good to analyze impact from real-world testing

rdlrt (approved):

PR #1851, opened 2025-02-03
rhel/fedora fixes

Scitz0:
## Description
guild-deploy fixes for recent RHEL / Fedora versions


scripts/cnode-helper-scripts/guild-deploy.sh
```diff
@@ -229,7 +229,7 @@ os_dependencies() {
       pkg_list="${pkg_list} libusb ncurses-compat-libs pkgconfig"
     elif [[ "${VERSION_ID}" =~ "8" ]] || [[ "${VERSION_ID}" =~ "9" ]]; then
       #RHEL/CentOS/RockyLinux 8/9
-      pkg_opts="${pkg_opts} --allowerasing"
+      if ${pkgmgrcmd} -h  | grep -q "\--allowerasing"; then pkg_opts="${pkg_opts} --allowerasing"; fi
```
TrevorBenson, line 232: This works for dnf v4 as the root usage menu displays this as a **General DNF option**. However, in dnf v5 `--allowerasing` no longer visible here, it is only visible in the usage for `dnf install`'s **Options**. 

This can be used to achieve your goal as both v4 and v5 list this in the usage menu for the install and update subcommands:

```suggestion
      if ${pkgmgrcmd} install -h  | grep -q "\--allowerasing"; then pkg_opts="${pkg_opts} --allowerasing"; fi
```

scripts/cnode-helper-scripts/guild-deploy.sh
```diff
@@ -238,7 +238,7 @@ os_dependencies() {
       fi
     elif [[ "${DISTRO}" =~ Fedora ]]; then
       #Fedora
-      pkg_opts="${pkg_opts} --allowerasing"
+      if ${pkgmgrcmd} -h  | grep -q "\--allowerasing"; then pkg_opts="${pkg_opts} --allowerasing"; fi
```
TrevorBenson, line 241: Same as above comment for dnf v4 vs. v5:

```suggestion
      if ${pkgmgrcmd} install -h  | grep -q "\--allowerasing"; then pkg_opts="${pkg_opts} --allowerasing"; fi
```

TrevorBenson (changes requested):

rdlrt (approved):

PR #1859, opened 2025-04-18
Cardano node 10.3.1

Scitz0:
> :warning: **WIP**: Do NOT deploy!

## Description
Changes to adapt to the new Cardano node 10.3.x and Cardano CLI 10.7.x. This update affects many components and scripts due to Oroboros Genesis preparation, new tracing system, and more. 

- Version bumps
  - [x] cardano-node 10.3.1
  - [x] cardano-cli 10.8.0.0
  - [x] Haskell GHC 9.6.7
  - [x] Haskell Cabal 3.12.1.0
  - [x] node-deps.json
  - [ ] ogmios ?.?.?
  - [ ] DBSync ?.?.?.?
- config.json (all networks except sancho)
  - Added:
    - [x] `"ConsensusMode": "PraosMode"`. `GenesisMode` supported but for now opted to use current consensus mode. See https://book.world.dev.cardano.org/adv-preview.html#ouroboros-genesis-mode for more details.
    - [x] `"MinBigLedgerPeersForTrustedState": 5`
    - [x] `"SyncTargetNumberOfActiveBigLedgerPeers": 30`
    - [x] `"SyncTargetNumberOfActivePeers": 0`
    - [x] `"SyncTargetNumberOfEstablishedBigLedgerPeers": 50`
    - [x] `"SyncTargetNumberOfKnownBigLedgerPeers": 100`
    - [x] `"UseTraceDispatcher": false`
  - Changed:
    - [x] Bumped `MinNodeVersion` to `10.2.1`, <mark>use 10.3.1 intead?</mark>
    - [x] `"TargetNumberOfEstablishedPeers": 40` to match defaults (except guild)
    - [x] `"TargetNumberOfRootPeers": 100` to match defaults (except guild)
  - Removed:
    - [x] `EnableLogging`
    - [x] `EnableLogMetrics`
    - [x] `EKGViewBK` in `setupBackends`
- topology.json [no changes due to `PraosMode`]
- env
  - [x] min node/cli version check bump
  - <mark>TBD</mark>
- cardano-cli
  - [x] Moved `genesis hash` command to `hash genesis-file`
  - [x] Replace all `conway` references with `latest`.
  - [x] Add canonical CBOR output toggle for transaction building and signing commands.
    Keeping cardano-hw-cli transform if HW is involved as it only covers one part of CIP-21.
  - [x] New default (JSON) for `query utxo`, add `--output-text` to revert to old style.
  - [x] Remove NETWORK_ERA from env as its not used anywhere.
  - [x] Add governance action deposits to the output of `query stake-address-info`. This also renames the field `stakeDeposit` to `stakeRegistrationDeposit` in the JSON output.
    Parsed but not used.
  - [x] `query spo-stake-distribution` to include the DRep delegation
    Proposal vote summary calculation now take into account SPO default delegation (abstain | no confidence).
  - [x] Added `query stake-pool-default-vote` to pool > show output
  - `query proposals` to replace `query gov-state | jq .proposals`
    No change needed right now, can be something for later.
  - `governance drep id` have --output-bech32 and --output-hex instead of --output-format STRING
    We dont use --output-format and default is the same (bech32), no changes needed.
  - `transaction txid` add --output-[json,text] flag to control format of the output
    Same default (text) as before, no changes needed.
  - transaction submit, new output
    We dont print or parse output, only check the return code for error, no change needed.
  - Add the ratify-state query
    No use case currently, ignore.
- Questions
  - Remove Sanchonet?
    <mark>No longer part of Cardano Book and, from my understanding, doesn't have any use case anymore.</mark>
  - Support for mnemonic sentence generation and extended key derivation from mnemonic sentences added to cardano-cli.
    <mark>Still a bit too limited compared to cardano-address, no custom purpose (1854), no custom index for gov keys.
    Could potentially replace it by Martins cardano-signer that does allow for arbitrary path?</mark>
  - Add stake address registration and delegation certificate and stake pool delegation certificate to compatible commands
    <mark>Is this to create combination certificates introduced in conway? 
    Needed or stick to single purpose tx like we have?</mark>


rdlrt:
Status: This PR is primarily waiting for corresponding component releases, we might bump this to 10.4.x depending on new versions compatibility that become available upstream

PR #1861, opened 2025-05-04
Support new fullDuplexConns metric for duplex connections

TommieGoldhammer:
## Description
In Cardano Node v10.4.1 the EKG/Prometheus metric for active duplex connections was renamed from

cardano_node_metrics_connectionManager_prunableConns

to

cardano_node_metrics_connectionManager_fullDuplexConns

This patch updates gLiveView’s metric parsing to:
1. First try to read fullDuplexConns (new key).
2. Then, for backward-compatibility, fall back to the old prunableConns key if fullDuplexConns is missing.
3. Ensure the displayed “Duplex” count remains accurate on both new and older Node versions.

With this, gLiveView will correctly report duplex connection counts on Cardano-node v10.4.1+ while remaining compatible with earlier releases.


rdlrt:
Changing target branch as this shouldn't go to alpha until proper ecosystem tooling is available for "perceived ready for mainnet" releases (which is often just for node itself). While 10.4.1 isnt mentioned as ready - it's something that'd build off node -10.3.1 branch instead of alpha

Scitz0:
Thanks for taking the time to create this one. I have yet to look at changes for 10.4.x. 

One note on the PR is that we don't normally add code for backward compatibility. When needed, we tag existing code and then bump the minimum node version requirement. Otherwise, it would be unmanageable. I also saw that your change was for EKG metrics. We want to make the same changes for Prometheus if it's changed there as well.

PR #1866, opened 2025-05-19
Feature: mithril snapshot download command to include ledger snapshot db

asnakep:
Hello CNTools/Koios Team,

The latest mithril-client v0.12.1 introduces the --include-ancillary option to include the ledger snapshot when downloading a snapshot, enabling faster node bootstrap.

In line with these recent changes, I would like to propose the following update to the mithril-client.sh script:

The current "download" command under mithril-client.sh cardano-db download would be replaced by two options: "immutable" and "full".

immutable – Download Cardano DB from Mithril snapshot (Immutable db only requires ledger replay)

full – Download Cardano DB from Mithril snapshot (Includes ledger snapshot for fast bootstrap)

The goal is to provide users with both choices:

Download only the immutable DB and let the node perform ledger replay

Download both the immutable and ledger snapshots for a faster bootstrap

Thank you,
Cheers,
Manuel

### DIFF
--- a/scripts/cnode-helper-scripts/mithril-client.sh
+++ b/scripts/cnode-helper-scripts/mithril-client.sh
@@ -32,7 +32,8 @@ usage() {
 			  override            Override default variable in the mithril environment file
 			  update              Update mithril environment file
 			cardano-db            Interact with Cardano DB
-			  download            Download Cardano DB from Mithril snapshot
+			  immutable           Download Cardano DB from Mithril snapshot (Immutable db only requires ledger replay)
+			  full	              Download Cardano DB from Mithril snapshot (Includes ledger snapshot for fast bootstrap)
 			  snapshot            Interact with Mithril snapshots
 			    list              List available Mithril snapshots
 			      json            List availble Mithril snapshots in JSON format
@@ -96,9 +97,13 @@ function main() {
     cardano-db)
       mithril_init client || exit 1
       case $2 in
-        download)
+        immutable)
           check_db_dir
           download_snapshot
+          ;;
+        full)
+          check_db_dir
+          download_snapshot_withledger
           ;;
         snapshot)
           case $3 in
--- a/scripts/cnode-helper-scripts/mithril.library
+++ b/scripts/cnode-helper-scripts/mithril.library
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 # shellcheck disable=SC2034,SC2086,SC2230,SC2206,SC2140,SC2059,SC2154
-# shellcheck source=/dev/null
+#shellcheck source=/dev/null
 
 ######################################
 # Do NOT modify code below           #
@@ -278,6 +278,7 @@ update_mithril_environment_for_client() {
 		RELEASE=${RELEASE}
 		AGGREGATOR_ENDPOINT=https://aggregator.${RELEASE}-${NETWORK_NAME,,}.api.mithril.network/aggregator
 		DB_DIRECTORY=${DB_DIR}
+		ANCILLARY_VERIFICATION_KEY=$(wget -q -O - https://raw.githubusercontent.com/input-output-hk/mithril/main/mithril-infra/configuration/${RELEASE}-${NETWORK_NAME,,}/ancillary.vkey)
 		GENESIS_VERIFICATION_KEY=$(wget -q -O - https://raw.githubusercontent.com/input-output-hk/mithril/main/mithril-infra/configuration/${RELEASE}-${NETWORK_NAME,,}/genesis.vkey)
 		SNAPSHOT_DIGEST=latest
 		EOF"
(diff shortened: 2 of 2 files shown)


TrevorBenson:
@asnakep Instead of replacing the **download** subcommand what do you think about making **immutable** and **full** subs of **download**? When undefined (e.g. `mithril-client.sh cardano-db download` is run without any subcommand to download) the default behavior could be set the continue doing what was the default before `--include-ancillary` command line option.

This would provide the new functionality you suggest while also avoiding the introduction of a breaking change. In the rare occurance of any SPO's which might have their own workflows using the existing scripts and the original behavior is preserved and they should not need to adjust anything. The user can explicitly set the alternate option if they want to adjust the behavior.

Additionally it removes the requirement to update the mithril_snapshot_download function in cnode.sh:

https://github.com/cardano-community/guild-operators/blob/541307fc120368b5ceb7ca7b140b07a43b3f92af/scripts/cnode-helper-scripts/cnode.sh#L79-L86

asnakep:
Hi @TrevorBenson 
thanks for the observation, I initially did it with subcommands then changed idea at last minute because new commands seemed a cleaner solution to me, during next hours i'll review it change new commands as sub commands of download, thanks for the advice on cnode.sh that I should have updated accordingly but missed it. For that I would use download full if you agree. See you later

asnakep:
Hello Trevor and team, 

I just pushed the version with one subcommand

cardano-db           Interact with Cardano DB
  -- download        Download Cardano DB from Mithril snapshot (Immutable db only requires node ledger replay)
    ---- full              Includes ledger snapshot for fast node bootstrap
 
Cheers,
Manuel

asnakep:
Hello, 

I added cnode.sh from alpha, I previously pushed mine one which is different,
sorry for the inconvenience

thank you,
Manuel

asnakep:
I just renamed function download_snapshot_withledger that was a ugly name
to download_snapshot_ancillary

scripts/cnode-helper-scripts/mithril-client.sh
```diff
@@ -97,8 +98,16 @@ function main() {
...
+          case $3 in
+            full)
+              check_db_dir
+              download_snapshot_ancillary
+              ;;
+            *)
+              check_db_dir
+              download_snapshot
+              ;;
+          esac
```
TrevorBenson, line 110: I believe the ledger state and last immutable file were initially in one snapshot, then separated when ancillary feature was introduced. I think it's part of the incremental DBs implementation.

@rdlrt @Scitz0: Are there security concerns or other reasons we should make MIthril's default download behavior skip the ledger state and require a replay? If not my preference is to maintain the SPO experience of the fastest startup times when using Mithril.

asnakep: https://github.com/input-output-hk/mithril/security/advisories/GHSA-qv97-5qr8-2266

At now ledger snapshots are signed and served by IOG through the only mainnet aggregator the mithril network has, the risk of tampered ledger snapshots distribution could rise in the future from one or more possible malicious aggregators in a multi-aggregators scenario.

TrevorBenson: Going with consistent behavior removes the need to change cnode.sh to use **cardano-db download full** to ensure ledger snapshots are included for fast startup.

Then either **skip-ancillary** or **immutable** would be the **download** subcommand to bypass the ledger snapshot. FWIW, my preference is **skip-ancillary** as the subcommand name is close to the `--include-ancillary` argument/flag. For users who know the upstream mithril-client binary it should be easy to understand the expected behavior with only a short review of the usage menu.

As @rdlrt mentions, documentation should also include a description of the new subcommand.

asnakep:
Hi Trevor and team,

I just pushed the changes as per last requirements, I go for documentation update

Cheers,
Manuel

asnakep:
Hello, 

I've just updated docs/Mithril/mithril-overview.md

4. Download the Mithril snapshot without the ledger state (skip-ancillary): This option downloads only the immutable db files. The ledger state will be computed from the genesis block when the Cardano node starts, which results in a longer bootstrap time.

./mithril-client.sh cardano-db download skip-ancillary

image

asnakep:
I wasn't fine with previous menu

image

I just updated it as

image

About docs probably the related section in docs/Mithril/mithril-overview.md could be enriched with some explanations about ancillary option meaning, the client version 0.12.1 which has this change, a link to https://mithril.network/doc/dev-blog/2025/05/06/client-breaking-change.

Scitz0:
> Hello Team,
> 
> just a question, do you want me to extend mithril-overview.md doc with information about client version 0.12.1 and its changes?
> 
> Thank you cheers, Manuel

I don't think we should have a changelog on the overview page, it should reflect current state of things for how our scripts wrap mithril. 

However, I would like to see an update also to the mithril-client markdown file as it has some stuff around the db download as well. It might be enough to just update the Usage section as the subcommand itself is mentioned on overview page.

Scitz0:
> Hello @Scitz0
> 
> some days ago we added point 4. in docs/Mithril/mithril-overview.md
> 
> Download the Mithril snapshot without the ledger state (skip-ancillary): This option downloads only the immutable db files. The ledger state will be computed from the genesis block when the Cardano node starts, which results in a longer bootstrap time.
> 
> ./mithril-client.sh cardano-db download skip-ancillary
> 
> image
> 
> cheers Manuel

Yes but in Scripts/mithril-client.md documentation we print the Usage output from -h. In there the new skip-ancillary subcommand isn't added.

Scitz0 (approved):
I have nothing further to add but will leave final go-ahead to @TrevorBenson

Issue #1879, opened 2025-08-03
Add  `cardano-tracer` support with backward compatibility.


ilap:
### Description

A new tracing system using the standalone cardano-tracer binary has been introduced. While the node still supports legacy tracing (via built-in EKG and Prometheus endpoints), the new system will become mandatory starting with `Cardano Node 10.6.0`.

The new tracer outputs telemetry in a different JSON/YAML structure, breaking compatibility with existing parsing logic in tools like gLiveView.sh.

Note that the new tracing system can be partially used without the `cardano-tracer` binary, but in this mode it only provides a minimal Prometheus endpoint directly from the node.

This issue proposes adding support for the new `cardano-tracer`-based system in a backward-compatible way, so users running different node versions can continue using the same tooling without modification.

### Goals

- Add support for metrics output by the new tracing system format.
- Ensure existing tools (e.g., `gLiveView.sh`) continue to work with legacy EKG/Prometheus endpoints.
- Avoid breaking changes — configuration should remain seamless across node versions.

### Acceptance Criteria
 - [x] Environment configuration (`env`) supports detecting and adapting the new tracing system.
 - [x] Tools like `gLiveView.sh` correctly parse metrics from either legacy or new tracer sources.
 - [ ] Maintains compatibility with Cardano Node versions prior to 10.5.1.
 - [x] Prepares for future versions (10.6.0+) where only the new tracing system will be supported.

### Additional Notes
- The change primarily involves updates in the `env` file (used by multiple guild tools), not in the tooling scripts themselves.
- Detection of tracer format may be manual (via env variable) or automatic if feasible.
- Prometheus remains available in both tracing systems, but with different structures and endpoint behaviors.
- EKG will only be available through `cardano-tracer` as forwarder in the new tracing system.

PR #1880, opened 2025-08-03
feat(env): add support for cardano-tracer with legacy fallback (#1879)

ilap:
### Summary

This PR introduces partial support for the new tracing system available in `cardano-node` 10.5.1+, which uses a standalone `cardano-tracer` binary. The implementation ensures backward compatibility with legacy telemetry (built-in EKG and Prometheus endpoints), allowing `gLiveView.sh` and related tools to work seamlessly across different node versions.

### Key Changes

- Enhanced `env` file with optional config flags to support `cardano-tracer`
- Compatible parsing logic for both legacy and tracer-based telemetry
- Maintains seamless usage of tools across versions (pre-10.5.1 and 10.5.1+)
- No changes required to `gLiveView.sh` logic directly — all handled via `env`

### Notes

- The new system will be **mandatory** from node version 10.6.0 onward.
- Prometheus is supported in both systems but with different structures and endpoints.
- EKG is no longer directly exposed by the node in the new system; only via `cardano-tracer` forwarding.
- This is a preparatory enhancement and does not yet finalize full tracer support.

### Testing

- Tested against both legacy and new telemetry configurations
- `gLiveView.sh` confirmed to load metrics and operate without error in both modes

Refs #1879

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,7 +17,7 @@ jobs:
     - name: SH Lint Check
       uses: ludeeus/action-shellcheck@master
       env:
-        SHELLCHECK_OPTS: -e SC2089 -e SC2090 -e SC2034
+        SHELLCHECK_OPTS: -e SC2089 -e SC2090 -e SC2034 -e SC2327 -e SC2328
       with:
         scandir: "./scripts"
         severity: "warning"
--- a/scripts/cnode-helper-scripts/env
+++ b/scripts/cnode-helper-scripts/env
@@ -600,7 +600,7 @@ getNodeMetrics() {
   CNODE_PID=$(pgrep -fn "$(basename ${CNODEBIN}).*.--port ${CNODE_PORT}") # Define again - as node could be restarted since last attempt of sourcing env
   [[ -n ${CNODE_PID} ]] && uptimes=$(( $(date -u +%s) - $(date -d "$(ps -p ${CNODE_PID} -o lstart=)" +%s) )) || uptimes=0
   if [[ ${USE_EKG} = 'Y' ]]; then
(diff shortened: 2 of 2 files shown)


ilap:
Some additional logic should be added when using different modes, see details below:
1. Using legacy tracer
  - UseTraceDispatcher = false or non-existing
  - ChainDB must be set
2. Using new tracer /w no forwarder
  -  gLiveView must only use the node's embedded `PrometheusSimple`.
  - UseTraceDispatcher = true and
  - TracerOptions.Forwarder must contain `PrometheusSimple`
  - USE_EKG="N" must be set "N", as there is no exposed EKG in non-forwarder mode. 
4. Using new tracer with forwarder
  - UseTraceDispatcher = true and
  - the cardano-tracer's config file must exist used by the TRACE_CONFIG's env variable
  - Check EKG or PROM
  - Prometheus can be exposed either in cardano-tracer or cardano-node.
  - CNODE_PID must be unset or add similar logic for cardano-tracer e.g. ad TRACER_PID logic.

rdlrt:
Thank you for the PR.

While it mostly looks good, we were arguing if we need to involve tracer at all, what advantage it'd bring v/s using SimplePrometheus which stays within node and does not require any additional setup change beyond config, and accordingly cleanup EKG part of the code.

It is good to have the option ofcourse - will rebase my branch on to yours and submit

rdlrt:
Ye - am onboard with support for tracer, especially for logging.
But given that legacy will be retired in 10.6.x, those using older node/configs will anyways have to specify something like `-b node-10.1.4` on gLV.

Accordingly, from my understanding - there shouldn't be a use-case for having EKG post setting disptacher to true (which would eventually be an enforced option), as we have been for a while replicating two options for metrics.

Thus, I believe there's opportunity to cleanup EKG references going forward and not have to worry about legacy mode (which will be available via tag for those who need to use it)

ilap:
Fixed the linting issues sorry.

ilap:
> Yip - happy to go with that as initial implementation and revisit EKG in future

Do you need anything to be done from me to get it merged?

rdlrt (approved):

PR #1886, opened 2025-09-04
Alternative missed slot calculation based on uptime (overcomes missing 10.5.1 metrics when using minSeverity Notice)

orpheus-antpool:
## Description
Provides an optional alternative calculation of missed slots based on **uptimes** instead of **about_to_lead** Prometheus counter (which is missing in 10.5.1 when **minSeverity>=Notice**). The idea is that **uptimes** is very close to **about_to_lead** after the block producer has been running for a while.

## Which issue it fixes?
10.5.1 requires **minSeverity<=Info** to trace several metrics, including **about_to_lead** counter. This is too verbose for some SPO setups. Without the alternative calculation and **minSeverity>=Notice**, missed slots percentile is incorrect (**about_to_lead** is missing and remains zero, so if *missed_slots** = 0 then it results in division by zero, otherwise it stays at 100%).

### DIFF
--- a/scripts/cnode-helper-scripts/gLiveView.sh
+++ b/scripts/cnode-helper-scripts/gLiveView.sh
@@ -1,3 +1,4 @@
+
 #!/usr/bin/env bash
 #shellcheck disable=SC2009,SC2034,SC2059,SC2206,SC2086,SC2015,SC2154
 #shellcheck source=/dev/null
@@ -20,6 +21,7 @@
 #CNCLI_CONNECT_ONLY=false                 # By default cncli measure full connect handshake duration. If set to false, only connect is measured similar to other tools
 #HIDE_DUPLICATE_IPS=N                     # If set to 'Y', duplicate and local IP's will be filtered out in peer analysis, else all connected peers are shown (default: N)
 #VERBOSE=N                                # Start in verbose mode showing additional metrics (default: N)
+#USE_UPTIME=N                             # Use uptime instead of about_to_lead Prometheus metric in the calculation of missed slots
 #GLV_LOG="${LOG_DIR}/gLiveView.log"       # Log gLiveView errors, set empty to disable. LOG_DIR set in env file.
 
 #####################################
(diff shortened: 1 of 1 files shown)


scripts/cnode-helper-scripts/gLiveView.sh
```diff
@@ -1311,7 +1312,11 @@ while true; do
       if [[ ${VERBOSE} = "Y" ]]; then
         printf "${VL} Missed slot leader checks"
         mvTwoSecond
-        LC_NUMERIC=C printf -v missed_slots_pct "%.4f" "$(bc -l <<<"(${missed_slots}/(${about_to_lead}+${missed_slots}))*100")"
+	if [[ ${USE_UPTIME} = "Y" ]]; then
+	    LC_NUMERIC=C printf -v missed_slots_pct "%.4f" "$(bc -l <<<"(${missed_slots}/${uptimes})*100")"
```
rdlrt, line 1316: Thank you for your PR.
1. Do we have any reference for the consideration that uptimes = about_to_lead + missed_slots?
2. Rather than using input variable, should we instead fallback based on availability of the metric about_to_lead?
3. Taking a step back - instead of fallback, it would be nice to actually find out which logger needs to be enhanced to bring back the counter (eg: @Scitz0 recently put ChainDb metric to `Info` - if there are others relevant to specific metric, happy to get that enabled in config itself)

Since we're still in alpha, not final merge - and we dont have a reference config from IO-managed environment book configs, we have to tune the default tracers anyways, would be nice to bring tuning requirements in config in-line with what currently have, instead of changing root logger to Info

orpheus-antpool: To your points:

uptimes = about_to_lead + missed_slots + init_time. After a while, init_time << about_to_lead + missed_slots, so it's a good approximation
yes, that's a good idea, fixed
I think there is merit in providing a fallback because gLiveView should not depend on a specific node config to work.

Scitz0: I agree that while a fallback might be good, let's try to find the correct trace tuning to get the metric returned by prometheus simple backend.

rdlrt: Ah gotcha, ye - we've made that breaking change purely due to their comms so far suggesting old configs will be removed at node 10.6.x (tho I find it hard to believe while new tracing infra isnt adopted by their own CI yet), but the new config format's only reference example we have currently is [this one](https://github.com/IntersectMBO/cardano-node/blob/master/configuration/cardano/mainnet-config-new-tracing.json).

PR #1899, opened 2026-02-10
Node patch bump 10.5.3 -> 10.5.4 & CNTools governance fix

Scitz0:
## Description
Bump node version to 10.5.4 to make it easy for users to deploy the latest patch version that includes a network security fix.


gufmar (approved):
LGTM

PR #1907, opened 2026-06-08
update cntools.sh / 1024 bytes / CIP-6

CardanoSecurity:
cntools: raise main metadata file cap to 1024 bytes when extended field present (CIP-6)

CIP-6 raises the 512-byte main metadata file cap to 1024 bytes when the file references extended metadata. Pools with longer descriptions that also include the extended field currently can't register/update through CNTools because the locally-built JSON crosses 512 bytes once the extended URL is appended.

- Locally-built file: 1024-byte cap when meta_extended_option is set, 512 otherwise.
- Downloaded file: peek for legacy `extended` or CIP-6 ext fields via jq; if any are present, allow 1024 bytes, else keep strict 512.
- Bump extended URL length check from 64 to 128 chars to match CIP-6 (extDataUrl is 128 char max) and the main URL check that was already bumped to 128 in this same flow.

### DIFF
--- a/scripts/cnode-helper-scripts/cntools.sh
+++ b/scripts/cnode-helper-scripts/cntools.sh
@@ -2094,7 +2094,9 @@ function main {
               metadata_done=false
               meta_tmp="${TMP_DIR}/url_poolmeta.json"
               if curl -sL -f -m ${CURL_TIMEOUT} -o "${meta_tmp}" ${meta_json_url} && jq -er . "${meta_tmp}" &>/dev/null; then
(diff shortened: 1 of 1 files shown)


Scitz0:
Thanks for the PR and it looks good to me. 

Just a quick note before I approve the PR. 
Please bump the patch version at the start of cntools.library and add a small note to cntools-changelog.md file. Saves me from doing it later :)

gufmar (approved):
LGTM

